PR #22, opened 2017-11-02
21: Fixed naming issue in endpoint. Added test to assert everything w…

dorpvom:
…orks.

New test did fail before fix was made. Functionality was only tested in unit tests prior. There the function is mocked and was named correctly.


src/test/integration/web_interface/rest/test_rest_file_object.py
```diff
@@ -22,6 +22,15 @@ def test_rest_download_valid(self):
         assert b'hid' in rv.data
         assert b'size' in rv.data
 
+    def test_rest_request_multiple_file_objects(self):
+        test_file_object = create_test_file_object()
```
weidenba: remove unused code

PR #39, opened 2017-12-19
Avm sqfs unpacker

weidenba:


src/plugins/unpacking/squashFS/code/squashFS.py
```diff
@@ -2,21 +2,42 @@
...
 def unpack_function(file_path, tmp_dir):
     '''
     file_path specifies the input file.
     tmp_dir should be used to store the extracted files.
     '''
-    output = execute_shell_command('fakeroot {} -d {}/faf_extracted {}'.format(squash_unpacker, tmp_dir, file_path))
-    return {'output': output}
+    unpack_result = dict()
+    for unpacker in squash_unpacker:
+        output = execute_shell_command('fakeroot {} -d {}/faf_extracted {}'.format(unpacker, tmp_dir, file_path))
```
tbarabosch: faf is legacy, right? Should be fact.

src/plugins/unpacking/squashFS/code/squashFS.py
```diff
@@ -2,21 +2,42 @@
...
+            unpack_result['{} - error'.format(_get_unpacker_name(unpacker))] = output
+    return unpack_result
+
+
+def _get_unpacker_name(unpacker_path):
+    return unpacker_path.split('/')[-1]
+
+
+def _unpack_success(tmp_dir):
+    return len(get_files_in_dir(tmp_dir)) > 0
```
tbarabosch: Can it happen that there is one accidental file? Add another more fuzzy test to see if there are folders like /bin or /var...

Issue #50, opened 2018-01-08
Firmware Tagging Feature


weidenba:
It should be possible to apply tags to firmware Images.

dorpvom:
Solved in #53

PR #53, opened 2018-01-10
50 firmware tag feature

weidenba:


weidenba:
solves #50

src/storage/db_interface_frontend.py
```diff
@@ -23,11 +23,12 @@ def get_meta_list(self, firmware_list=None):
...
             if firmware:
-                if firmware['processed_analysis']['unpacker']['file_system_flag']:
-                    unpacker = self.retrieve_analysis(deepcopy(firmware['processed_analysis']))['unpacker']['plugin_used']
+                if 'tags' in firmware:
+                    tags = firmware['tags']
                 else:
-                    unpacker = firmware['processed_analysis']['unpacker']['plugin_used']
-                list_of_firmware_data.append((firmware['_id'], self.get_hid(firmware['_id']), unpacker))
+                    tags = dict()
+                tags[firmware['processed_analysis']['unpacker']['plugin_used']] = TagColor.LIGHT_BLUE
```
dorpvom: There is a bug in case the unpacking result is sanitized.

src/web_interface/templates/database/database_browse.html
```diff
@@ -67,10 +67,8 @@
             {% endif %}
                 <td class="active clickable" onclick="location.href='/analysis/{{ firmware[0] }}'">
                     {{ firmware[1] }}
-                    <span class="label label-pill {% if firmware[2] == 'generic_carver' %} label-warning
```
dorpvom: see above regarding bug

src/storage/db_interface_frontend.py
weidenba: should be fixed

src/web_interface/templates/database/database_browse.html
weidenba: see above

PR #65, opened 2018-01-25
fixed exploit mitigation graph

0x00rick:
made exploit mitigations independent of each other (fixed #64 )


dorpvom:
Declined for now

PR #70, opened 2018-02-16
xerox unpacker added

weidenba:


src/plugins/unpacking/xerox/test/test_dlm.py
```diff
@@ -0,0 +1,59 @@
...
+    def setUp(self):
+        super().setUp()
+
+        self.test_file = FileObject(file_path=os.path.join(TEST_DATA_DIR, 'DLM-First_1MB.DLM'))
+        self.test_firmware = self.test_file.file_path
+        self.firmware_container = XeroxDLM(self.test_firmware)
+
+    def tearDown(self):
+        self.ds_tmp_dir.cleanup()
+        self.tmp_dir.cleanup()
```
jstucke: it seems these 2 lines are also called in the super class

PR #80, opened 2018-03-12
Add file header plugin

dorpvom:
New compare plugin providing a "binwalk -Ww"-ish view on the first 512 byte block of all compared files

### DIFF
--- a/src/plugins/compare/file_header/code/file_header.py
+++ b/src/plugins/compare/file_header/code/file_header.py
@@ -6,6 +6,9 @@ from compare.PluginBase import CompareBasePlugin
 from storage.binary_service import BinaryService
 from web_interface.components.additional_functions.hex_dump import convert_binary_to_ascii_with_dots
 
+COLUMN_WIDTH = 32
+BYTES_TO_SHOW = 512
+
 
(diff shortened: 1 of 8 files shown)


src/plugins/compare/file_header/code/file_header.py
```diff
@@ -0,0 +1,113 @@
+import binascii
+
+from flask import Markup
+
+from compare.PluginBase import CompareBasePlugin
+from storage.binary_service import BinaryService
```
weidenba: unused import

src/plugins/compare/file_header/test/test_file_header.py
```diff
@@ -0,0 +1,29 @@
+from unittest.mock import patch
+
+from flask import Markup
+
+from storage.binary_service import BinaryService
```
weidenba: unused import

src/plugins/compare/file_header/test/test_file_header.py
```diff
@@ -0,0 +1,29 @@
+from unittest.mock import patch
+
```
weidenba: unused import

PR #82, opened 2018-03-13
added script

0roman:


src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
...
+    global run
+    logging.info('shutting down FACT Analysis Component')
+    run = False
+
+
+signal.signal(signal.SIGINT, shutdown)
+signal.signal(signal.SIGTERM, shutdown)
+
+
+def main():
```
dorpvom: The command_line_options have to be given as parameter (and propagated to program_setup).

src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
...
+    args, config = program_setup(PROGRAM_NAME, PROGRAM_DESCRIPTION)
+    analysis_service = AnalysisScheduler(config=config)
+    analysis_service.shutdown()
+    logging.info('Restart Analysis component')
+    analysis_service = AnalysisScheduler(config=config)
```
dorpvom: Why Start the Scheduler twice?
Also: The CompareScheduler has to also be started, as it also does view sync.

src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
...
+    run = False
+
+
+signal.signal(signal.SIGINT, shutdown)
+signal.signal(signal.SIGTERM, shutdown)
+
+
+def main():
+    args, config = program_setup(PROGRAM_NAME, PROGRAM_DESCRIPTION)
+    analysis_service = AnalysisScheduler(config=config)
```
dorpvom: Scheduler won't start without MongoMgr being started before.

src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
+import logging
```
dorpvom: Script missing shebang and execution permissions

src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
...
+    analysis_service.shutdown()
+    logging.info('Restart Analysis component')
+    analysis_service = AnalysisScheduler(config=config)
+
+    run = True
+    while run:
+        try:
+            pass
+        except KeyboardInterrupt:
+            break
```
dorpvom: I don't want to kill this manually. Instead just wait some seconds and then start the teardown

src/test/acceptance/run_scripts/test_run_scripts.py
```diff
@@ -29,7 +30,7 @@ def test_start_script_help_and_version(script):
     gc.collect()
 
 
-@pytest.mark.parametrize('script', [init_database, migrate_database, update_statistic, update_variety_data])
+@pytest.mark.parametrize('script', [init_database, migrate_database, update_statistic, update_variety_data, restart_analysis_and_reload_templates])
```
dorpvom: test fails. See above for reasons.

src/restart_analysis_and_reload_templates.py
```diff
@@ -0,0 +1,43 @@
...
+
+from helperFunctions.program_setup import program_setup
+from scheduler.Analysis import AnalysisScheduler
+
+
+PROGRAM_NAME = 'Restart FACT Analysis Component'
+PROGRAM_DESCRIPTION = 'Restart FACT Analysis Component and reload Analysis Templates'
+
+
+def shutdown(signum, frame):
```
dorpvom: not needed if script self destroys (also applies for lines 20 + 21)

src/restart_analysis_and_reload_templates.py
dorpvom: Still missing ...

dorpvom:
Before merging into master, master has to be merged into this branch and the script has to be moved to a suitable location ( != src )

PR #87, opened 2018-03-23
Analysis tag scheduler

dorpvom:


src/plugins/analysis/crypto_material/view/crypto_material.html
```diff
@@ -3,7 +3,7 @@
 {% block analysis_result_details %}
 
 	{% for key in firmware.processed_analysis[selected_analysis] %}
-		{% if key not in ['summary', 'plugin_version', 'analysis_date'] %}
```
weidenba: Change should be applied to src/web_interface/templates/analysis_plugins/generic.html as well

PR #90, opened 2018-04-03
Add parameter and lots of tests

dorpvom:
When adding `?summary=true` to a firmware request, the summaries are collected from all child objects.


weidenba:
please add new parameter to wiki

PR #93, opened 2018-04-04
Pre install

weidenba:
Docker needs a reboot and cannot be installed during the regular installation :(


dorpvom:
As of now, only internal plugins use docker. So this will only be merged, once the first docker plugin makes it to **core**

PR #95, opened 2018-04-10
show admin options if logged in as admin only

weidenba:


src/helperFunctions/web_interface.py
```diff
@@ -77,3 +78,10 @@ def __exit__(self, *args):
 def get_template_as_string(view_name):
     path = os.path.join(get_template_dir(), view_name)
     return get_binary_from_file(path).decode('utf-8')
+
+
+def is_superuser(user):
+    if isinstance(user._get_current_object(), AnonymousUser):  # auth disabled
```
jstucke: it would be more intuitive if this functionality was put in its own function (e.g. "auth_is_enabled(user)"), since it has little to do with whether or not the user is a "superuser"
it would then need to be tested first (e.g. "if not auth_is_enabled(user) or is_superuser(user)")

src/web_interface/components/analysis_routes.py
```diff
@@ -78,7 +78,8 @@ def _show_analysis_results(self, uid, selected_analysis=None, root_uid=None):
                                           firmware_including_this_fo=firmware_including_this_fo,
                                           analysis_plugin_dict=analysis_plugins,
                                           other_versions=other_versions,
-                                          uids_for_comparison=uids_for_comparison)
+                                          uids_for_comparison=uids_for_comparison,
+                                          show_admin_panal=is_superuser(current_user))
```
jstucke: typo? (panel)

src/web_interface/templates/show_analysis.html
```diff
@@ -407,7 +407,7 @@ <h2>Analysis for <span class="text-primary">{{ firmware.get_hid(root_uid=root_ui
 	    </div>
     </div>
     {# admin buttons #}
-    {% if firmware.vendor %}
+    {% if firmware.vendor and show_admin_panal %}
```
jstucke: see above

src/helperFunctions/web_interface.py
jstucke: user_has_admin_clearance()

PR #106, opened 2018-04-24
needed dep

0x00rick:
added needed dependency in jinja_filter.py for custom view

### DIFF
--- a/src/web_interface/components/jinja_filter.py
+++ b/src/web_interface/components/jinja_filter.py
@@ -2,6 +2,7 @@
 
 import json
 import random
+import hashlib
 
 from common_helper_filter.time import time_format
 from flask import render_template
@@ -123,3 +124,5 @@ class FilterClass:
         self._app.jinja_env.filters['fix_cwe'] = fix_cwe
         self._app.jinja_env.filters['vulnerability_class'] = vulnerability_class
         self._app.jinja_env.filters['auth_enabled'] = self.check_auth
+	self._app.jinja_env.filters['hash'] = lambda s: hashlib.md5(s.encode()).hexdigest()
+


src/web_interface/components/jinja_filter.py
```diff
@@ -124,6 +125,7 @@ def _setup_filters(self):
         self._app.jinja_env.filters['fix_cwe'] = fix_cwe
         self._app.jinja_env.filters['vulnerability_class'] = vulnerability_class
         self._app.jinja_env.filters['auth_enabled'] = self.check_auth
+	self._app.jinja_env.filters['hash'] = lambda s: hashlib.md5(s.encode()).hexdigest()
```
weidenba: Please use "helperFunctions.hash.get_md5"
Otherwise you must write a test for that  :wink:

weidenba: By the way: I think it would be better to call the filter 'md5_hash' or something like that.

src/web_interface/components/jinja_filter.py
```diff
@@ -125,7 +125,7 @@ def _setup_filters(self):
         self._app.jinja_env.filters['fix_cwe'] = fix_cwe
         self._app.jinja_env.filters['vulnerability_class'] = vulnerability_class
         self._app.jinja_env.filters['auth_enabled'] = self.check_auth
-	self._app.jinja_env.filters['hash'] = lambda s: hashlib.md5(s.encode()).hexdigest()
+        self._app.jinja_env.filters['md5_hash'] = lambda s: get_md5(s.encode())
```
weidenba: The helper function should be capable of handling any kind of input. Is the encoding really necessary at this point?

src/web_interface/components/jinja_filter.py
```diff
@@ -125,7 +125,7 @@ def _setup_filters(self):
         self._app.jinja_env.filters['fix_cwe'] = fix_cwe
         self._app.jinja_env.filters['vulnerability_class'] = vulnerability_class
         self._app.jinja_env.filters['auth_enabled'] = self.check_auth
-        self._app.jinja_env.filters['md5_hash'] = lambda s: get_md5(s.encode())
+        self._app.jinja_env.filters['md5_hash'] = lambda s: get_md5(s)
```
weidenba: Do we need the lambda function any more?
`self._app.jinja_env.filters['md5_hash'] = get_md5` should do the same?

PR #123, opened 2018-06-07
Early database update

dorpvom:
Added `pre_analysis` function to scheduler. This can be specified like `post_analysis` using the constructor of the scheduler class.

The `pre_analysis` function defaults to the add_object method previously being the default of `post_analysis`, thus the objects are now stored in the database and accessible to the frontend directly after the unpacker finished.

This introduces the problem, that the `file_type` plugin is not necessarily completed before some views are shown for an object. The subsequent bugs in the analysis views have been mitigated.


dorpvom:
Note that the post_analysis method is now called after each successfully run plugin such that each analysis might now be written to the database multiple times. Since mongo does not present a performance bottleneck here, this is not a problem.

Issue #102, opened 2018-04-24
Binary search on single firmware


weidenba:
It should to scan a single firmware with yara.

PR #127, opened 2018-06-25
102 binary search on single

jstucke:
addresses #102

### DIFF
--- a/src/helperFunctions/yara_binary_search.py
+++ b/src/helperFunctions/yara_binary_search.py
@@ -37,7 +37,7 @@ class YaraBinarySearchScanner:
     def _execute_yara_search_for_single_firmware(self, rule_file_path, firmware_uid):
         with ConnectTo(YaraBinarySearchScannerDbInterface, self.config) as connection:
             file_paths = connection.get_file_paths_of_files_included_in_fo(firmware_uid)
-        result = [self._execute_yara_search(rule_file_path, path) for path in file_paths]
+        result = (self._execute_yara_search(rule_file_path, path) for path in file_paths)
         return b'\n'.join(result)
 
     @staticmethod
@@ -70,8 +70,13 @@ class YaraBinarySearchScanner:
         '''
         with NamedTemporaryFile() as temp_rule_file:
             yara_rules, firmware_uid = task
-            temp_rule_file.write(yara_rules)
+            try:
+                compiled_rules = yara.compile(source=yara_rules.decode())
+            except yara.SyntaxError as e:
+                return YaraRuleError('There seems to be an error in the rule file:\n{}'.format(e))
+            compiled_rules.save(file=temp_rule_file)
             temp_rule_file.flush()
+
             try:
                 if firmware_uid is None:
                     raw_result = self._execute_yara_search(temp_rule_file.name)
--- a/src/helperFunctions/yara_binary_search.py
+++ b/src/helperFunctions/yara_binary_search.py
@@ -3,6 +3,11 @@ from subprocess import check_output, CalledProcessError, STDOUT
 from tempfile import NamedTemporaryFile
 
 import yara
+from typing import Tuple, List, Optional
+
+from helperFunctions.web_interface import ConnectTo
+from storage.db_interface_common import MongoInterfaceCommon
+from storage.fs_organizer import FS_Organizer
 
 
 class YaraRuleError(Exception):
@@ -16,7 +21,7 @@ class YaraBinarySearchScanner:
         self.config = config
         self.db_path = self.config['data_storage']['firmware_file_storage_directory']
 
-    def _execute_yara_search(self, rule_file_path):
+    def _execute_yara_search(self, rule_file_path, target_path=None):
         '''
         scans the (whole) db directory with the provided rule file and returns the (raw) results
         yara-python cannot be used, because it (currently) supports single-file scanning only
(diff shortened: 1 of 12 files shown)


src/helperFunctions/yara_binary_search.py
```diff
@@ -51,16 +62,21 @@ def _eliminate_duplicates(result_dict):
...
-    def get_binary_search_result(self, yara_rules):
+    def get_binary_search_result(self, task: Tuple[bytes, Optional[str]]):
         '''
-        :param yara_rules: byte string with the contents of the yara rule file
+        :param task: tuple containing the yara_rules (byte string with the contents of the yara rule file) and optionally a firmware uid if only the contents
+                     of a single firmware are to be scanned
         :return: dict of matching rules with lists of (unique) matched UIDs as values
         '''
         with NamedTemporaryFile() as temp_rule_file:
+            yara_rules, firmware_uid = task
```
dorpvom: So, as i understand it we execute yara with plain text rules, i.e. non-compiled rules. Correct?
If so, this does not introduce major overhead on a single execution, since the rules are compiled only once. If though, we run the single firmware search, we would yara possibly on thousands of files, making this thousands of compilations.

If my assumption is correct I would suggest compiling the rules into a rule file first (see yarac).

src/helperFunctions/yara_binary_search.py
```diff
@@ -51,16 +62,21 @@ def _eliminate_duplicates(result_dict):
...
             temp_rule_file.write(yara_rules)
             temp_rule_file.flush()
             try:
-                raw_result = self._execute_yara_search(temp_rule_file.name)
+                if firmware_uid is None:
+                    raw_result = self._execute_yara_search(temp_rule_file.name)
+                else:
+                    raw_result = self._execute_yara_search_for_single_firmware(temp_rule_file.name, firmware_uid)
             except YaraRuleError as e:
                 return e
```
dorpvom: Not your code, but why **return** exception here. Would prefer returning a string if that's the whole purpose.

src/helperFunctions/yara_binary_search.py
```diff
@@ -81,3 +97,15 @@ def get_yara_error(rules_file):
...
         return e
+
+
+class YaraBinarySearchScannerDbInterface(MongoInterfaceCommon):
+
+    READ_ONLY = True
+
+    def get_file_paths_of_files_included_in_fo(self, fo_uid: str) -> List[str]:
+        fs_organizer = FS_Organizer(self.config)
+        return [
```
dorpvom: May be nitpicking, but since this can be a huge list (thousands of paths) maybe use an iterator instead (round instead square brackets)

src/storage/db_interface_common.py
```diff
@@ -238,6 +239,12 @@ def get_set_of_all_included_files(self, fo):
         else:
             return set()
 
+    def get_uids_of_all_included_files(self, uid: str) -> Set[str]:
```
dorpvom: Same size remark from `yara_binary_search.py` would apply here but I don't think there is such a thing as an iterator set ...

src/web_interface/templates/database/database_binary_search.html
```diff
@@ -22,9 +22,19 @@ <h2>Binary Search</h2>
                     for (i = 0; i < tablinks.length; i++) {
                         tablinks[i].className = tablinks[i].className.replace(" active", "");
                     }
-                    document.getElementById(tabname).style.display = "block";
+                    document.getElementById(tabname).style.display = "inline";
```
dorpvom: As always I won't code review the javascript mess. The functional test was successful though.

src/test/unit/helperFunctions/test_yara_binary_search.py
```diff
@@ -3,30 +3,90 @@
...
+from helperFunctions.config import get_config_for_testing
+from helperFunctions.yara_binary_search import YaraRuleError
+
+
+TEST_FILE_1 = 'binary_search_test'
+TEST_FILE_2 = 'binary_search_test_2'
+TEST_FILE_3 = 'binary_search_test_3'
+
+
+class MockCommonDbInterface:
```
dorpvom: Is it possible to use one of the existing mock interfaces. I would rather enjoy it if we would reduce the number of hard coded mocks instead of increasing it ; )

src/test/unit/helperFunctions/test_yara_binary_search.py
```diff
@@ -3,30 +3,90 @@
...
+        result = self.yara_binary_scanner.get_binary_search_result((self.yara_rule, None))
+        self.assertEqual(result, {'test_rule': [TEST_FILE_1]})
+
+    def test_get_binary_search_result_error(self):
+        result = self.yara_binary_scanner.get_binary_search_result((b'invalid yara rule {}', None))
+        assert isinstance(result, YaraRuleError)
+
+    def test_get_binary_search_result_for_single_firmware(self):
+        yara_rule = b'rule test_rule_2 {strings: $a = "TEST_STRING!" condition: $a}'
+        result = self.yara_binary_scanner.get_binary_search_result((yara_rule, 'single_firmware'))
```
dorpvom: You could test the propagation of the uid by changing the mocked get_uids_of_all_included_files method.

src/storage/db_interface_common.py
jstucke: this should be no issue here, since this is only a set of UIDs. Even a set of 1000 UIDs should take less than 50KB of memory.

PR #128, opened 2018-06-27
fixed issue with bad directory cleanup

dorpvom:
There were some problems with downloading tar.gz archives.
This stemmed from files being extracted as root-owned or missing write permissions.

The problem is fixed via an already existing function in the `unpackBase`.
Code was refactored though.

### DIFF
--- a/src/unpacker/unpackBase.py
+++ b/src/unpacker/unpackBase.py
@@ -88,10 +88,10 @@ class UnpackBase(object):
 
     def change_owner_back_to_me(self, directory=None, permissions='u+r'):
         with Popen('sudo chown -R {}:{} {}'.format(getuid(), getgid(), directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
-            pl.communicate()[0].decode()
+            pl.communicate()
         self.grant_read_permission(directory, permissions)
 
     @staticmethod
     def grant_read_permission(directory, permissions):
         with Popen('chmod --recursive {} {}'.format(permissions, directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
-            pl.communicate()[0].decode()
+            pl.communicate()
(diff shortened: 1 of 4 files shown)


src/unpacker/unpackBase.py
```diff
@@ -86,13 +86,12 @@ def _extract_files_from_file_using_specific_unpacker(self, file_path, tmp_dir, s
         meta_data['analysis_date'] = time()
         return get_files_in_dir(tmp_dir), meta_data
 
-    def change_owner_back_to_me(self, directory=None):
-        uid = getuid()
-        gid = getgid()
-        with Popen('sudo chown -R {}:{} {}'.format(uid, gid, directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
+    def change_owner_back_to_me(self, directory=None, permissions='u+r'):
+        with Popen('sudo chown -R {}:{} {}'.format(getuid(), getgid(), directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
             pl.communicate()[0].decode()
```
jstucke: does the `[0].decode()` have any purpose?

src/unpacker/unpackBase.py
```diff
@@ -86,13 +86,12 @@ def _extract_files_from_file_using_specific_unpacker(self, file_path, tmp_dir, s
...
-        self.grant_read_permission(directory)
+        self.grant_read_permission(directory, permissions)
 
-    def grant_read_permission(self, directory=None):
-        with Popen('chmod --recursive u+r {}'.format(directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
+    @staticmethod
+    def grant_read_permission(directory, permissions):
+        with Popen('chmod --recursive {} {}'.format(permissions, directory), shell=True, stdout=PIPE, stderr=PIPE) as pl:
             pl.communicate()[0].decode()
```
jstucke: see above

PR #152, opened 2018-08-06
146 rest binary search

jstucke:


src/test/acceptance/rest/test_rest_binary_search.py
```diff
@@ -0,0 +1,41 @@
...
+        self._stop_backend()
+        super().tearDown()
+
+    def test_binary_search(self):
+        self._upload_firmware()
+        time.sleep(2)  # wait for binary to be saved
+
+        data = {'rule_file': 'rule rulename {strings: $a = "MyTestRule" condition: $a }'}
+        rv = self.test_client.post('/rest/binary_search', data=json.dumps(data), follow_redirects=True)
+        result = json.loads(rv.data)
```
dorpvom: jenkins fails on this line since `data` has type `bytes` while `json.loads` expects a `str`. Suspiciously this does not fail on my local machine. Nevertheless, using `.decode()` here makes sense.

dorpvom: Also in line 33

src/test/unit/web_interface/rest/test_rest_binary_search.py
```diff
@@ -0,0 +1,51 @@
...
+    assert 'Started binary search' in result['message']
+
+
+def test_start_binary_search_with_uid(test_app):
+    data = json.dumps({'rule_file': YARA_TEST_RULE, 'uid': 'uid_in_db'})
+    result = decode_response(test_app.post('/rest/binary_search', data=data))
+    assert 'Started binary search' in result['message']
+
+
+def test_get_result_without_search_id(test_app):
```
dorpvom: This should return a **404** if the `=None` is removed as suggested above (or below?).

src/web_interface/rest/rest_binary_search.py
```diff
@@ -0,0 +1,76 @@
...
+        The result of the search request can only be fetched once. After this the search needs to be started again.
+        The results have the form:
+        {'binary_search_results': {'<rule_name_1>': ['<matching_uid_1>', ...], '<rule_name_2>': [...], ...}
+        '''
+
+        if search_id is None:
+            return error_message('The search_id was not found. It is needed to fetch the search result', self.URL)
+
+        with ConnectTo(InterComFrontEndBinding, self.config) as intercom:
+            result, rule = intercom.get_binary_search_result(search_id)
```
dorpvom: `rule` is unused

PR #153, opened 2018-08-09
Rabbitmq remote scheduler

weidenba:


dorpvom:
Will not be merged in current form.

PR #162, opened 2018-10-08
Restructure readme

dorpvom:
Removed installation details from readme and added a more extensive markdown for installation.
Fixed some deprecated information and added a nicer representation for requirements.


INSTALL.md
```diff
@@ -0,0 +1,86 @@
+# FACT Installation
+
+FACT consists of three components: frontend, database and backend. All components can be installed on one machine as well as on different machines.
+There is an automated installation program supporting Ubuntu 16.04 and 18.04 host systems.
+
+## TL;DR
+
+0. `(cd path/to/fact/sources && ./pre_install.sh)`
```
weidenba: path should be "src/bootstrap/pre_install.sh"

INSTALL.md
```diff
@@ -0,0 +1,86 @@
...
+$ ./install.py --help
+```
+
+## Multi system setup
+
+The two worker components (frontend, backend) communicate exclusively through the database. The database in turn does not needed any knowledge of its place in the network, other than on which **ip:port** combination the database server has to be hosted.
+The main.cfg on the frontend system has to be altered so that the values of `data_storage.mongo_server` and `data_storage.mongo_port` match the **ip:port** for the database.
+The same has to be done for the backend. In addition, since the raw firmware and file binaries are stored in the backend, the `data_storage.firmware_file_storage_directory` has to be created (by default `/media/data/fact_fw_data`).
+On the database system, the `mongod.conf` has to be given the correct `net.bindIp` and `net.port`. In addition the path in `storage.dbPath` of the `mongod.conf` has to be created. 
+
```
weidenba: The install options should be explained as well ./install.py [-B -D -F]
Furhtermore we should mention that the databse should be started first.

dorpvom: I disagree with the second request. Starting the database first - or at all - has nothing to do with installation but execution. This should be documented in a getting started page.

PR #171, opened 2018-10-30
Locking unpacked files

dorpvom:
Problem of race condition has appeared since binary deletion was activated recently.
Whenever a firmware is deleted, issuing a cascade of file deletions, a new analysis can overlap with the files. Then a file can be extracted during analysis, deleted shortly after due to the firmware deletion, and analyzed later on at which point the file has been deleted.
This happens basically every time a re-do anlysis is issued, but can happen in other scenarios as well.

A locking of extracted files should prevent that from happening. That is:
1. Unpacker finds file
2. file is locked
3. file is stored on fs
4. file is analyzed
5. analysis is stored in database
6. lock is released

As of now, if a deletion request is issued in between steps 3 and 5, the file is remove as no database link exists to it. Once step 5 is processed, a new link in the database exists and the lock can be safely removed. As long as the lock is in place, any deletion attempt will be discarded.

### DIFF
--- a/src/intercom/back_end_binding.py
+++ b/src/intercom/back_end_binding.py
@@ -182,11 +182,14 @@ class InterComBackEndDeleteFile(InterComListener):
         if self._entry_was_removed_from_db(task['_id']):
             logging.info('remove file: {}'.format(task['_id']))
             self.fs_organizer.delete_file(task['_id'])
-        else:
-            logging.warning('file not removed, because database entry exists: {}'.format(task['_id']))
         return None
 
     def _entry_was_removed_from_db(self, uid):
         with ConnectTo(MongoInterfaceCommon, self.config) as db:
-            entry_is_in_database = db.existence_quick_check(uid)
-        return not entry_is_in_database
+            if db.existence_quick_check(uid):
+                logging.debug('file not removed, because database entry exists: {}'.format(uid))
+                return False
+            elif db.check_unpacking_lock(uid):
+                logging.debug('file not removed, because it is processed by unpacker: {}'.format(uid))
+                return False
+        return True
--- a/src/test/common_helper.py
+++ b/src/test/common_helper.py
@@ -75,6 +75,7 @@ class DatabaseMock:
 
     def __init__(self, config=None):
         self.tasks = []
+        self.locks = []
 
     def shutdown(self):
         pass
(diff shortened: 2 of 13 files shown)


src/intercom/back_end_binding.py
```diff
@@ -182,11 +182,14 @@ def post_processing(self, task, task_id):
...
             self.fs_organizer.delete_file(task['_id'])
-        else:
-            logging.warning('file not removed, because database entry exists: {}'.format(task['_id']))
         return None
 
     def _entry_was_removed_from_db(self, uid):
         with ConnectTo(MongoInterfaceCommon, self.config) as db:
-            entry_is_in_database = db.existence_quick_check(uid)
-        return not entry_is_in_database
+            if db.existence_quick_check(uid):
```
jstucke: it would be great if you could add a test for the new functionality

src/test/common_helper.py
```diff
@@ -320,6 +320,15 @@ def get_file_name(self, uid):
         if uid == 'deadbeef00000000000000000000000000000000000000000000000000000000_123':
             return 'test_name'
 
+    def set_unpacking_lock(self, uid):
+        pass
+
+    def check_unpacking_lock(self, uid):
```
jstucke: check_unpacking_lock and release_unpacking_lock are currently unused by the tests

src/test/common_helper.py
```diff
@@ -320,6 +321,18 @@ def get_file_name(self, uid):
...
             return 'test_name'
 
+    def set_unpacking_lock(self, uid):
+        self.locks.append(uid)
+
+    def check_unpacking_lock(self, uid):
+        return uid in self.locks
+
+    def release_unpacking_lock(self, uid):
+        self.locks.remove(uid)
```
jstucke: function is unused

src/test/unit/unpacker/test_unpacker.py
```diff
@@ -135,6 +135,12 @@ def test_detect_unpack_loss_no_data_lost(self):
         self.assertIn('no data lost', container.processed_analysis['unpacker']['summary'])
         self.assertNotIn('data loss', container.processed_analysis['unpacker'])
 
+    def test_file_is_locked(self):
+        assert not self.unpacker.db_interface.check_unpacking_lock(self.test_fo.uid)
+        file_pathes = ['{}/get_files_test/testfile1'.format(get_test_data_dir())]
```
jstucke: typo

PR #172, opened 2018-11-02
dahua unpacker added

weidenba:
Extraction plugin for dahua firmware images.

Dahua uses zip containers but change the magic bytes from 'PK' to 'DH'.


src/plugins/unpacking/dahua/code/dahua.py
```diff
@@ -0,0 +1,23 @@
...
+from common_helper_files import get_binary_from_file, write_binary_to_file
+from pathlib import Path
+
+name = 'dahua'
+mime_patterns = ['firmware/dahua']
+version = '0.1'
+
+
+def unpack_function(file_path, tmp_dir):
+    output_file_path = Path(tmp_dir, 'dahua_firmware.zip')
```
jstucke: imho it would look better in the file tree (and be more intuitive) if the original file name was kept and a ".zip" was appended

Issue #175, opened 2018-11-06
Python installer aborts on depedencies already installed by distutils


weidenba:
We should think about installation issues that should not stop installation completely.
```
2018-11-06 08:35:03][install][INFO]: Installing pymongo pyyaml
Traceback (most recent call last):
  File "src/install.py", line 154, in <module>
    common(distribution)
  File "/home/faf/FACT_core/src/install/common.py", line 66, in main
    pip3_install_packages('pymongo', 'pyyaml')
  File "/home/faf/FACT_core/src/helperFunctions/install.py", line 92, in pip3_install_packages
    return _pip_install_packages(3, args)
  File "/home/faf/FACT_core/src/helperFunctions/install.py", line 83, in _pip_install_packages
    return run_shell_command_raise_on_return_code('sudo -EH pip{} install --upgrade {}'.format(version, ' '.join(args)), 'Error in installation of python package(s) {}'.format(' '.join(args)), True)
  File "/home/faf/FACT_core/src/helperFunctions/install.py", line 51, in run_shell_command_raise_on_return_code
    raise InstallationError(error)
helperFunctions.install.InstallationError: Error in installation of python package(s) pymongo pyyaml
Collecting pymongo
  Downloading https://files.pythonhosted.org/packages/b1/45/5440555b901a8416196fbf2499c4678ef74de8080c007104107a8cfdda20/pymongo-3.7.2-cp36-cp36m-manylinux1_x86_64.whl (408kB)
Collecting pyyaml
Installing collected packages: pymongo, pyyaml
  Found existing installation: pymongo 3.7.1
    Uninstalling pymongo-3.7.1:
      Successfully uninstalled pymongo-3.7.1
  Found existing installation: PyYAML 3.12
Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.

```

weidenba:
same for psutils

weidenba:
solved in #176

PR #176, opened 2018-11-07
handling distutils case seperately

dorpvom:
Packets are now installed separately so that only the installation of specific packets fail.
Not that of all other packets that are installed with the packet in question.


dorpvom:
Should solve #175

Issue #164, opened 2018-10-16
REST-API keyword change of firmware_version


IoT-junkrat:
Hey FACT-Team,

I just want to mention that you changed one of the keywords for the "/rest/firmware" API endpoint.
It has changed from "firmware_version" to "version".
If I wanted to search for firmware images before, I had to execute that command:
`curl 'http://localhost:5000/rest/firmware -X GET -G --data-urlencode 'query={"device_class": "Printer", "device_name": "XY", "firmware_version": "2.0", "vendor": "Dell"}'`

Now, I have to execute this command:
`curl http://localhost:5000/rest/firmware -X GET -G --data-urlencode 'query={"device_name": "XY", "device_class": "Printer", "version": "2.0", "vendor": "Dell"}'`

Took me a while to recognize that, since it does not show up in the Wiki.
Maybe someone else stumbles over it too, because FACT is not replying with any UIDs anymore using keyword "firmware_version".
Cheers,
IoT-junkrat

weidenba:
Hi IoT-junkrat,

Thanks for the pointer. Indeed there is a name inconsistency between the PUT and the GET function. We are going to discuss this internally and fix it in the near future. 

Peter

PR #183, opened 2018-11-21
Align version naming

dorpvom:
Fixing mismatched naming of firmware version, where in database and object `version` was used and upload used `firmware_version`.

Side effect: Removed unused function `convert_fw_obj_to_analysis_task` from task conversion helper.

Name was already changed in wiki.


dorpvom:
Should close #164

PR #186, opened 2018-12-12
Fs metadata plugin

jstucke:
moved FS metadata plugin to core


src/plugins/analysis/file_system_metadata/review
```diff
@@ -0,0 +1,49 @@
+quotation mix:
```
tbarabosch: This file shouldn't be here?!

tbarabosch:
Looks fine now.

PR #212, opened 2019-02-11
raw extractor added

weidenba:


src/mime/custom_mime_raw
```diff
@@ -0,0 +1,4 @@
+# RAW data from flash
+
+8	string	\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff	RAW
+!:mime data/raw
```
jstucke: seems new line is missing here which leeds to warning 
```
custommime, 4361: Warning: MIME type ` data/raw# ====================== UEFI ======================' has bad char '#'
```
when compiling the file

src/plugins/unpacking/raw/test/test_raw.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+from test.unit.unpacker.test_unpacker import TestUnpackerBase
+
+from common_helper_files.fail_safe_file_operations import get_binary_from_file
+
+TEST_DATA_DIR = Path(Path(__file__).parent, 'data')
+
+
+class TestHpStreamUnpacker(TestUnpackerBase):
```
jstucke: is the class name correct?

PR #219, opened 2019-02-20
Generic fs mbr support

dorpvom:
Extracts images containing master boot record and multiple partitions


src/plugins/unpacking/generic_fs/code/generic_fs.py
```diff
@@ -25,18 +29,63 @@ def unpack_function(file_path, tmp_dir):
...
+    sleep(1)  # Necessary since initialization of special devices seem to take some time
+    if not return_code == 0:
+        return 'Failed to mount master boot record image:\n{}'.format(output)
+
+    loop_devices = _extract_loop_devices(output)
+
+    mount_dir = TemporaryDirectory()
+    for index, loop_device in enumerate(loop_devices):
+        output += _mount_loop_device(loop_device, mount_dir.name, tmp_dir, index)
+    mount_dir.cleanup()
```
jstucke: maybe something like
```python
    with TemporaryDirectory() as mount_dir:
        for index, loop_device in enumerate(loop_devices):
            output += _mount_loop_device(loop_device, mount_dir, tmp_dir, index)
```
would be a more elegant solution

jstucke: and maybe it would be better to extract the search for loop devices to it's own function

src/plugins/unpacking/generic_fs/code/generic_fs.py
```diff
@@ -25,18 +29,63 @@ def unpack_function(file_path, tmp_dir):
...
     output += execute_shell_command('sudo cp -av {}/* {}/'.format(mount_dir.name, tmp_dir))
     output += execute_shell_command('sudo umount -v {}'.format(mount_dir.name))
     mount_dir.cleanup()
+    return output
 
-    return {'output': output}
+
+def _mount_from_boot_record(file_path, tmp_dir):
+    output, return_code = execute_shell_command_get_return_code('sudo kpartx -a -v {}'.format(file_path))
+    sleep(1)  # Necessary since initialization of special devices seem to take some time
```
jstucke: typo: *seems

src/plugins/unpacking/generic_fs/code/generic_fs.py
```diff
@@ -1,17 +1,21 @@
 '''
 This plugin unpacks SquashFS filesystem images
```
jstucke: I'm pretty sure this isn't right

PR #227, opened 2019-03-19
added warning if repository is not checked out via git

dorpvom:


dorpvom:
Initial tries to add repository meta data (.git folder, history etc.) to allow submodule update after tarball installation failed.
Help/Ideas would be appreciated.

PR #228, opened 2019-03-27
Refactoring and test coverage

jstucke:


src/storage/db_interface_compare.py
```diff
@@ -23,56 +33,61 @@ def add_compare_result(self, compare_result):
...
 
     def delete_old_compare_result(self, compare_id):
         try:
-            self.compare_results.remove({'_id': unify_string_list(compare_id)})
+            self.compare_results.remove({'_id': normalize_compare_id(compare_id)})
             logging.debug('old compare result deleted: {}'.format(compare_id))
-        except Exception as e:
-            logging.warning('Could not delete old compare result: {} {}'.format(sys.exc_info()[0].__name__, e))
+        except Exception as exception:
+            logging.warning('Could not delete old compare result: {} {}'.format(sys.exc_info()[0].__name__, exception))
```
dorpvom: replace `sys.exc_info()[0].__name__` with `type(exception)`

src/web_interface/components/io_routes.py
```diff
@@ -52,10 +56,12 @@ def _download_base64_decoded_section(self, uid, section, expression_id):
             binary = binascii.a2b_base64(binary[span_in_section[0]:span_in_section[1]])
         except binascii.Error as error:
             return render_template('error.html', message=str(error))
-        else:
-            resp = make_response(binary)
-            resp.headers['Content-Disposition'] = 'attachment; filename={}'.format(file_obj.file_name + '_0x%X' % (span_in_binary[0] + span_in_section[0]) + '-0x%X_decoded' % (span_in_binary[1] - span_in_section[2]))
-            return resp
+        response = make_response(binary)
+        file_name = (file_obj.file_name
```
dorpvom: original nasty code seems to come from me, still..
Would it not conform to our string format style if the string + string + string part is rewritten to something like `'{}{}{}'.format(string, string, string)`?

jstucke: I just can't wait for the day we adapt f-strings...

PR #261, opened 2019-05-28
Refactoring and test coverage

jstucke:
isort should be working correctly now with the option `known_first_party=analysis,test`

### DIFF
--- a/src/plugins/analysis/string_evaluation/code/string_eval.py
+++ b/src/plugins/analysis/string_evaluation/code/string_eval.py
@@ -1,13 +1,13 @@
-import os
 import sys
-
-from common_helper_files import get_dir_of_file
+from pathlib import Path
 
 from analysis.PluginBase import AnalysisBasePlugin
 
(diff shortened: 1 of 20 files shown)


src/plugins/analysis/string_evaluation/code/string_eval.py
```diff
@@ -1,13 +1,13 @@
...
-from common_helper_files import get_dir_of_file
+from pathlib import Path
 
 from analysis.PluginBase import AnalysisBasePlugin
 
-THIS_FILE_DIR = get_dir_of_file(__file__)
-sys.path.append(os.path.join(THIS_FILE_DIR, '..', 'internal'))
-from string_eval import eval_strings
+try:
+    from ...string_evaluation.internal.string_eval import eval_strings
```
weidenba: why `...` instead of `..internal`

jstucke: fixed

PR #272, opened 2019-07-03
'faf' -> 'fact' and pylint warnings fixed

weidenba:
There are still some to broad exception warnings, but these need further investigation.

### DIFF
--- a/src/objects/file.py
+++ b/src/objects/file.py
@@ -2,15 +2,14 @@ import logging
 import os
 
 from common_helper_files import get_binary_from_file
-
-from helperFunctions.dataConversion import make_bytes, make_unicode_string, get_value_of_first_key
+from helperFunctions.dataConversion import get_value_of_first_key, make_bytes, make_unicode_string
 from helperFunctions.hash import get_sha256
 from helperFunctions.uid import create_uid
 
 
-class FileObject(object):
+class FileObject():  # pylint: disable=too-many-instance-attributes
     '''
-    This is the base file objects. All files in FAF should be implemented as this object type.
+    This is the base file objects. All files in FACT should be implemented as this object type.
     '''
 
     def __init__(self, binary=None, file_name=None, file_path=None, scheduled_analysis=None):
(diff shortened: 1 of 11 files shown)


src/objects/file.py
```diff
@@ -121,7 +120,7 @@ def get_virtual_paths_for_one_uid(self, root_uid=None):
             return ["insufficient information: firmware analysis not complete"]
 
     def get_virtual_file_paths(self):
-        if len(self.virtual_file_path.keys()) > 0:
+        if self.virtual_file_path.keys():
```
jstucke: wouldn't `if self.virtual_file_path:` suffice?

src/objects/file.py
```diff
@@ -140,8 +139,7 @@ def get_top_of_virtual_path(virtual_path):
     def get_root_uid(self):
         if self.root_uid is not None:
             return self.root_uid
-        else:
-            return list(self.get_virtual_file_paths().keys())[0]
+        return list(self.get_virtual_file_paths().keys())[0]
```
jstucke: the `.keys()` could be redundant here, too

src/test/integration/storage/test_db_interface_admin.py
```diff
@@ -67,7 +66,7 @@ def test_remove_object_field(self):
     def test_remove_virtual_path_entries_no_other_roots(self):
         self.db_backend_interface.add_file_object(self.child_fo)
         self.assertIn(self.uid, self.db_backend_interface.file_objects.find_one(self.child_uid, {'virtual_file_path': 1})['virtual_file_path'])
-        removed_vps, deleted_files = self.admin_interface._remove_virtual_path_entries(self.uid, self.child_fo.get_uid())
+        removed_vps, deleted_files = self.admin_interface._remove_virtual_path_entries(self.uid, self.child_fo.get_uid())  # pylint: disable=protected-access
```
jstucke: `# pylint: disable=protected-access` can be moved to the top of the file so it doesn't need to be repeated for each line
IMO  this makes sense for test files (maybe we can get a separate test configuration for pylint at some point)

PR #273, opened 2019-07-08
CVE lookup plugin

mellowCS:


src/plugins/analysis/cve_lookup/internal/setup_repository.py
```diff
@@ -0,0 +1,252 @@
+import argparse
```
dorpvom: Theres multiple issues here. I fixed some minor ones already. At least
- the default arguments don't work. This can't be run without parameters
- the relative imports don't work. Furthermore when I fix the issues to make setup work, the plugin doesn't work anymore
- The `glob()` for the `extraction_path` only works when the argument ends on a `/` this is not intuitive and likely leads to unforced errors. Also the IndexError is not handled when the glob is empty.

weidenba:
moved to external plugin until refactoring is complete
https://github.com/fkie-cad/FACT_analysis-plugin_CVE-lookup

Issue #32, opened 2017-11-30
refactor tabs in navbar


dorpvom:
1. Step: Database should be dropdown instead of page.
2. Step: Reduce amount of tabs by merging some of them into new dropdown menues.

dorpvom:
Solved by merge of Pull Request #49

PR #280, opened 2019-07-30
Fix broken cwe_checker plugin

tbarabosch:
Since cwe_checker changed its output format, we have to parse json instead of sexp. Good news: this makes the FACT plugin much smaller! :+1: 

Please do not merge before PR #32 of cwe_checker has been merged (https://github.com/fkie-cad/cwe_checker/pull/32).

### DIFF
--- a/src/plugins/analysis/cwe_checker/code/cwe_checker.py
+++ b/src/plugins/analysis/cwe_checker/code/cwe_checker.py
@@ -22,7 +22,6 @@ from analysis.PluginBase import AnalysisBasePlugin
 
 BAP_TIMEOUT = 10  # in minutes
 DOCKER_IMAGE = 'fkiecad/cwe_checker:latest'
-DOCKER_IMAGE = 'cwe-checker:latest'
 
 
 class AnalysisPlugin(AnalysisBasePlugin):
--- a/src/plugins/analysis/cwe_checker/code/cwe_checker.py
+++ b/src/plugins/analysis/cwe_checker/code/cwe_checker.py
@@ -15,6 +15,7 @@ lifts to the following architectures:
 import logging
 import json
 from collections import defaultdict
+from subprocess import Popen, PIPE, DEVNULL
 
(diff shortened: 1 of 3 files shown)


src/plugins/analysis/cwe_checker/code/cwe_checker.py
```diff
@@ -106,38 +65,36 @@ def _parse_module_versions(bap_output):
...
     @staticmethod
     def _parse_bap_output(output):
         tmp = defaultdict(list)
-        cwe_parser = CweWarningParser()
-
-        for line in output.splitlines():
-            if 'WARN' in line:
-                cwe_warning = cwe_parser.parse(line)
-                tmp[cwe_warning.name].append(cwe_warning)
+        j = json.loads(output)
```
dorpvom: If you have pre-commit set up, this would not have been allowed. Please use more semantic name to comply with coding guidelines

src/plugins/analysis/cwe_checker/code/cwe_checker.py
```diff
@@ -106,38 +65,36 @@ def _parse_module_versions(bap_output):
...
+        if 'warnings' in j:
+            for warning in j['warnings']:
+                tmp[warning['name']] = tmp[warning['name']] + [warning]
```
dorpvom: I'd prefer the [warning, ] syntax to explicitly show warning to be one element. If warning would be a string, the string would be split into chars using the syntax without trailing comma

src/plugins/analysis/cwe_checker/install.sh
```diff
@@ -20,10 +20,6 @@ else
     rm -fR internal
     git clone https://github.com/fkie-cad/cwe_checker.git internal
 
-    echo "Cleaning up"
-    rm -rf internal/src/_build
-    rm -f internal/src/cwe_checker.plugin
-
```
dorpvom: Please also remove sexpdata installation

PR #292, opened 2019-09-27
fixed issue with missing unpacker data and added regression test

dorpvom:
unpacker "analysis result" misses when unpacker depth limit is reached. In this case NOP is popped in there as a placeholder

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -4,14 +4,14 @@ import sys
 from copy import deepcopy
 
 from helperFunctions.compare_sets import remove_duplicates_from_list
-from helperFunctions.dataConversion import get_value_of_first_key
 from helperFunctions.database_structure import visualize_complete_tree
-from helperFunctions.file_tree import get_partial_virtual_path, FileTreeNode
+from helperFunctions.dataConversion import get_value_of_first_key
+from helperFunctions.file_tree import FileTreeNode, get_partial_virtual_path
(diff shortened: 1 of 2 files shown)


src/storage/db_interface_frontend.py
```diff
@@ -24,22 +24,19 @@ def get_meta_list(self, firmware_list=None):
...
-                    submission_date = 0
+                tags = firmware['tags'] if 'tags' in firmware else dict()
+                tags[self._get_unpacker_name(firmware)] = TagColor.LIGHT_BLUE
+                submission_date = firmware['submission_date'] if 'submission_date' in firmware else 0
                 list_of_firmware_data.append((firmware['_id'], self.get_hid(firmware['_id']), tags, submission_date))
         return list_of_firmware_data
 
+    def _get_unpacker_name(self, firmware):
+        if 'unpacker' not in firmware['processed_analysis']:
+            return 'NOP'
```
jstucke: why "NOP"? 
Is this a case that could happen regularly or should there be a log message?

dorpvom: As stated above, this will happen regularly for firmware images with multiple layers.
I saw `NOP` as fitting, since this is the name of our unpack plugin that _runs_ whenever no extraction is wanted (e.g. blacklisted file types such as **.doc** or **.pdf**).
So as unpack depth is a special case of _should-not-be-extracted_ this looks sort of coherent.

PR #301, opened 2019-10-25
replaced .get_uid() with property

jstucke:
Using getters and setters is not especially "pythonic". I therefore propose this solution using pyhton properties.

### DIFF
--- a/src/objects/file.py
+++ b/src/objects/file.py
@@ -58,6 +58,10 @@ class FileObject():  # pylint: disable=too-many-instance-attributes
         if self.file_name is None:
             self.set_name(os.path.basename(file_path))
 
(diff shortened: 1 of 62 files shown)


src/objects/file.py
```diff
@@ -58,14 +59,20 @@ def set_file_path(self, file_path):
             self.set_name(os.path.basename(file_path))
 
     def get_uid(self):
-        if self.uid is None and self.binary is not None:
-            self.uid = create_uid(self.binary)
+        logging.warning('Deprecation warning: "get_uid()" was replaced by "uid" and will be removed in a future update')
```
weidenba: Testing this function should be easy ;)

Issue #173, opened 2018-11-05
Analysis not complete warning bug


jstucke:
If two elements with the same UID are in both the firmware DB and the file object DB, then the "analysis is not complete yet" warning will be displayed on the analysis page of the object containing the file object.

Reason: the warning is displayed when the number of file objects included in the object is different than the number of objects with the UIDs in the DB. Since there is an entry in both the firmware DB and the file object DB, it will be counted twice and the numbers won't match.

Possible long-term fix: merging firmware objects and file objects into a single database

jstucke:
fixed in #308

PR #308, opened 2019-11-14
analysis not complete bugfix

jstucke:
fix for #173

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -124,7 +124,7 @@ class FrontEndDbInterface(MongoInterfaceCommon):
             return True
         query = self._build_search_query_for_uid_list(uid_list)
         number_of_results = self.get_firmware_number(query) + self.get_file_object_number(query)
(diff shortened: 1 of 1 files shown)


src/storage/db_interface_frontend.py
```diff
@@ -124,7 +124,7 @@ def all_uids_found_in_database(self, uid_list):
             return True
         query = self._build_search_query_for_uid_list(uid_list)
         number_of_results = self.get_firmware_number(query) + self.get_file_object_number(query)
-        return len(uid_list) == number_of_results
+        return number_of_results >= len(uid_list)
```
dorpvom: Can you explain why this is necessary?

jstucke: The bug occurs if the same object is in the file object and firmware databases.
The number of uids is compared to the number of objects in the databases and (since one uid is in both databases) they don't match up.

jstucke: Originally, the plan was to fix this by merging both databases (if that ever happens remains to be seen)

Issue #118, opened 2018-06-04
re organize tabs on fact navbar


dorpvom:
In admin mode there are now 8 elements + logo + quicksearch button on the navbar.
Some of the elements should be reordered to allow for better overview.

e.g. profile and admin can be moved to one dropdown.

PR #313, opened 2019-11-20
restructured nav bar

jstucke:
addresses #118


dorpvom:
I approve of the code. We should find a better name than "Misc" though

PR #315, opened 2019-11-20
improved test coverage + refactoring

jstucke:
add tests for untested functions to improve test coverage
moved function `get_exclusive_files` to `CompareDbInterface` for it to be testable


src/test/integration/storage/test_db_interface_compare.py
```diff
@@ -1,28 +1,26 @@
...
+class TestCompare:
 
     @classmethod
-    def setUpClass(cls):
+    def setup_class(cls):
         cls._config = get_config_for_testing()
         cls.mongo_server = MongoMgr(config=cls._config)
 
-    def setUp(self):
+    def setup(self):
```
weidenba: Is this still functional? `setUp` is a special function name defined in unittest
https://docs.python.org/3/library/unittest.html

src/test/integration/storage/test_db_interface_compare.py
```diff
@@ -1,28 +1,26 @@
...
+from storage.MongoMgr import MongoMgr
 from test.common_helper import create_test_firmware
 
 
-class TestCompare(unittest.TestCase):
+class TestCompare:
 
     @classmethod
-    def setUpClass(cls):
+    def setup_class(cls):
```
weidenba: s.b.

src/test/integration/storage/test_db_interface_compare.py
```diff
@@ -32,8 +30,9 @@ def setUp(self):
         self.fw_two = create_test_firmware()
         self.fw_two.set_binary(b'another firmware')
         self.compare_dict = self._create_compare_dict()
+        self.compare_id = '{};{}'.format(self.fw_one.uid, self.fw_two.uid)
 
-    def tearDown(self):
+    def teardown(self):
```
weidenba: s.a.

src/test/integration/storage/test_db_interface_compare.py
```diff
@@ -42,36 +41,40 @@ def tearDown(self):
         gc.collect()
 
     @classmethod
-    def tearDownClass(cls):
+    def teardown_class(cls):
```
weidenba: s.a.

src/storage/db_interface_compare.py
```diff
@@ -96,3 +95,13 @@ def get_ssdeep_hash(self, uid):
     def get_entropy(self, uid):
         file_object_entry = self.file_objects.find_one({'_id': uid}, {'processed_analysis.unpacker.entropy': 1})
         return file_object_entry['processed_analysis']['unpacker']['entropy'] if 'unpacker' in file_object_entry['processed_analysis'] else 0.0
+
+    def get_exclusive_files(self, compare_id: str, root_uid: str) -> List[str]:
```
weidenba: I am not so happy with this decision since this a functionality that is only used in the compare plug-in. We should discuss this next week.

src/test/integration/storage/test_db_interface_compare.py
jstucke: `setup`, `teardown`, `setup_class` and `teardown_class` are the equivalent functions in `pytest`

src/storage/db_interface_compare.py
jstucke: The problem is: In its previous location it's almost impossible to test

PR #319, opened 2019-11-27
Refactor check exceptions

jstucke:
moved similar `check_exception` functions to `helper_functions/process.py` and added a test

### DIFF
--- a/src/scheduler/Unpacking.py
+++ b/src/scheduler/Unpacking.py
@@ -85,7 +85,7 @@ class UnpackingScheduler:
 
     def start_work_load_monitor(self):
         logging.debug('Start work load monitor...')
-        return start_single_worker(0, 'unpack-load', self._work_load_monitor)
+        return start_single_worker(None, 'unpack-load', self._work_load_monitor)
 
     def _work_load_monitor(self):
         while self.stop_condition.value == 0:
--- a/src/scheduler/Unpacking.py
+++ b/src/scheduler/Unpacking.py
@@ -5,12 +5,12 @@ from queue import Empty
 from time import sleep
 
 from helperFunctions.parsing import bcolors
-from helperFunctions.process import ExceptionSafeProcess, terminate_process_and_childs
+from helperFunctions.process import ExceptionSafeProcess, check_worker_exceptions, start_single_worker
 from storage.db_interface_common import MongoInterfaceCommon
 from unpacker.unpack import Unpacker
 
 
-class UnpackingScheduler(object):
+class UnpackingScheduler:
     '''
     This scheduler performs unpacking on firmware objects
     '''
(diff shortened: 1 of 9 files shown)


src/scheduler/Unpacking.py
```diff
@@ -115,24 +114,4 @@ def _get_combined_analysis_workload(self):
...
-                    self._start_single_worker(process_index)
-        return return_value
-
-    def _start_single_worker(self, process_index):
-        process = ExceptionSafeProcess(target=self.unpack_worker,
-                                       name='Unpacking-Worker-{}'.format(process_index),
-                                       args=(process_index,))
-        process.start()
-        self.workers.append(process)
+        return check_worker_exceptions(self.workers, 'Unpacking', self.config, self.unpack_worker)
```
dorpvom: Unpacker does not "die" even if config parameter that tell it to is set

dorpvom:
Also, there is another import error here:

Since you now use the `bcolors` class from parsing here, you have to import `parsing` from everywhere that `process` is imported, e.g. `init_database.py` which is called in the `db.py`.
`parsing` though has `pyopenssl` as dependency and that is installed in the backend. :boom: 

```
13:56:57 Traceback (most recent call last):
13:56:57   File "./src/install.py", line 169, in <module>
13:56:57     install()
13:56:57   File "./src/install.py", line 155, in install
13:56:57     db(distribution)
13:56:57   File "/home/jenkins/workspace/xenial_dev/src/install/db.py", line 50, in main
13:56:57     raise InstallationError('Unable to initialize database\n{}'.format(init_output))
13:56:57 helperFunctions.install.InstallationError: Unable to initialize database
13:56:57 Traceback (most recent call last):
13:56:57   File "init_database.py", line 23, in <module>
13:56:57     from storage.MongoMgr import MongoMgr
13:56:57   File "/home/jenkins/workspace/xenial_dev/src/storage/MongoMgr.py", line 8, in <module>
13:56:57     from helperFunctions.config import get_config_dir
13:56:57   File "/home/jenkins/workspace/xenial_dev/src/helperFunctions/config.py", line 6, in <module>
13:56:57     from helperFunctions.process import complete_shutdown
13:56:57   File "/home/jenkins/workspace/xenial_dev/src/helperFunctions/process.py", line 13, in <module>
13:56:57     from helperFunctions.parsing import bcolors
13:56:57   File "/home/jenkins/workspace/xenial_dev/src/helperFunctions/parsing.py", line 4, in <module>
13:56:57     import OpenSSL
13:56:57 ImportError: No module named 'OpenSSL'
```

src/scheduler/Unpacking.py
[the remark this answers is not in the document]
jstucke: Yes, but this is also the case in `master`
This should probably be fixed in another branch

jstucke: `self.workers.append()` was missing so the check ran on an empty list ... should be fixed now

PR #330, opened 2020-01-10
ReFACToring 2

dorpvom:
- Compare exception handler would crash system whenever an exception occured in compare
- hash plugin did run in timeout on some peculiar files


src/scheduler/Compare.py
```diff
@@ -86,7 +86,10 @@ def _decide_whether_to_process(uid, redo, compares_done):
...
         processes_to_check = [self.worker]
         shutdown = check_worker_exceptions(processes_to_check, 'Compare', self.config, self._compare_scheduler_main)
+        if shutdown:
+            return shutdown
+
         if new_worker_was_started(new_process=processes_to_check[0], old_process=self.worker):
             self.worker = processes_to_check.pop()
 
-        return shutdown
+        return False
```
jstucke: could be simplified to
```python
shutdown = check_worker_exceptions(processes_to_check, 'Compare', self.config, self._compare_scheduler_main)
if not shutdown and new_worker_was_started(new_process=processes_to_check[0], old_process=self.worker):
    self.worker = processes_to_check.pop()
return shutdown
```

src/plugins/analysis/hash/code/hash.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, plugin_administrator, config=None, recursive=True):
 
         # additional init stuff can go here
 
-        super().__init__(plugin_administrator, config=config, recursive=recursive, plugin_path=__file__)
+        super().__init__(plugin_administrator, config=config, recursive=recursive, plugin_path=__file__, timeout=600)
```
jstucke: maybe we should split the plugin in the long run into a "basic hash" plugin for mandatory hashes and an "advanced hash" plugin for hashes that could take a while or even fail to compute

dorpvom: I agree. But since tlsh is a dependecy for compare we then either make the advanced hashes mandatory as well or we have to account for that in the compare. This is also a lead to cross-plugin results since tlsh is calculated in the tlsh plugin (which depends, like compare, on tlsh being present for all files (although in this case the non existence is covered) and imphash can be easily calculated using the results of elf_analysis

Issue #338, opened 2020-02-07
bugs in file tree


jstucke:
- files with the same uid are named the same (they are named after the file name and not the virtual file path)
- if two files with the same uid are in the same level of the file tree, only one is displayed

jstucke:
fixed in #339

PR #339, opened 2020-02-07
fix bugs in file tree + refactoring

jstucke:
addresses #338

### DIFF
--- a/src/web_interface/components/ajax_routes.py
+++ b/src/web_interface/components/ajax_routes.py
@@ -27,8 +27,7 @@ class AjaxRoutes(ComponentBase):
 
     @roles_accepted(*PRIVILEGES['view_analysis'])
     def _ajax_get_tree_children(self, uid, root_uid=None, compare_id=None):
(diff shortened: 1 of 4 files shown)


src/web_interface/components/ajax_routes.py
```diff
@@ -27,16 +27,21 @@ def _init_component(self):
...
         return jsonify(children)
 
-    def _generate_exclusive_files_tree(self, exclusive_files, root_uid, uid):
+    def _get_exclusive_files(self, compare_id, root_uid):
+        if compare_id:
+            with ConnectTo(CompareDbInterface, self._config) as sc:
+                return sc.get_exclusive_files(compare_id, root_uid)
+        return None
+
+    def _generate_exclusive_files_tree(self, exclusive_files, root_uid, uid) -> FileTreeNode:
```
weidenba: the wording is misleading, since it is not just used for the exclusive file tree.

` def _generate_files_tree(self, whitelist, root_uid, uid) -> FileTreeNode:`

we should add a docstring explaining "the whitlist" feature.

Issue #354, opened 2020-02-21
IP and URI finder plugin seems to generate false positive URI 0.4.1


weidenba:
IP and URI finder plugin seems to generate false positive URI 0.4.1 which is equal to the system version

weidenba:
solved in #361

PR #361, opened 2020-02-28
fix view of ip-and-uri-finder

weidenba:
solves #354

### DIFF
--- a/src/plugins/analysis/ip_and_uri_finder/view/ip_and_uri_finder.html
+++ b/src/plugins/analysis/ip_and_uri_finder/view/ip_and_uri_finder.html
@@ -3,7 +3,7 @@
 {% block analysis_result_details %}
 
 	{% for key in firmware.processed_analysis[selected_analysis] %}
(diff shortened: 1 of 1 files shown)


src/plugins/analysis/ip_and_uri_finder/view/ip_and_uri_finder.html
```diff
@@ -3,7 +3,7 @@
 {% block analysis_result_details %}
 
 	{% for key in firmware.processed_analysis[selected_analysis] %}
-		{% if key not in ['summary', 'plugin_version', 'analysis_date', 'skipped'] %}
+		{% if key not in ['summary', 'plugin_version', 'system_version', 'analysis_date', 'skipped'] %}
 			<tr>
```
jstucke: we need a better solution to this problem or we will encounter it again each time a new key is added to the results dict
the real question is: how can we accomplish that without breaking the view for old analyses?

jstucke: we could save all results under the key `results` or use a prefix like `__` but this would break the view for old analyses

Issue #344, opened 2020-02-13
Never analyzed files


IoT-junkrat:
Hey guys,

I sometimes encounter the issue that a single file analysis can not be completed (due to analysis plugin crashs or whatever reason), the UI shows - as expected- "not analyzed yet" forever.
What I am missing now is a functionality to check all files if they have been analyzed, even if the current analysis queue is showing 0 files to process
.
This feature could be run initially by starting up FACT, like an iterative command every x days or on a manual base. 

For now, even if I dig through all files in the UI file tree, I am not able to restart an analysis of a single file, because of the "Analysis might still be pending" message. So the current solution is to dig through all file trees of all binaries and re-do the analysis of the whole binary.

I hope you get my point 😀 
Would be very grateful if you consider adding this feature.

Thank you & Cheers!

dorpvom:
Hi,

just for clarification, the "not analyzed yet" and "might still be pending" messages refer to files that have not been analyzed at all yet (i.e. no plugin finished), correct?
To this point I've seen this issue only for firmware images that crash FACT somehow, e.g. by running out of memory.

If not a single plugin has run on a given file, the file must have been dropped by the scheduler somehow. Again, I've only encountered this when the scheduler itself has crashed.

Meanwhile, when all queues show 0 pending tasks, tasks may still be running on the workers. I regularly encounter this with the cwe_checker plugin. To measure this, we'd need some kind of process introspection for the worker processes. This could be achieved using a decorator. But we'd have to discuss if this is sensible in the context of our current data flow.

IoT-junkrat:
> just for clarification, the "not analyzed yet" and "might still be pending" messages refer to files that have not been analyzed at all yet (i.e. no plugin finished), correct?
Yes, that is correct.

Exactly, this happens when the problems occur, which you described.

In my case, the workers are not processing the file, since I would be "waiting" for 3 months then and I am using FACT on a powerful machine. :)

dorpvom:
So, in the meantime we've decided to test a solution for the tracking of running tasks. Once we have some action there, I'll link to this issue.
That said, your scenario ("waiting for 3 months") sounds like it should be handled by our plugin timeouts. Or is the long running time referring to the extraction part?

IoT-junkrat:
That sounds nice.
I don't have any insights what kind of mechanism broke (extraction or analysis plugin). I only see files in the tree tagged as not yet analyzed and staying there forever.
[...]

PR #367, opened 2020-03-18
Experimental find failed analyses

jstucke:
addresses #344
added new page `/admin/missing_analyses` that shows:
- missing files (files that are included in another file_object/firmware but are not found in the DB)
- missing analyses (analyses that are present in the firmware object but not in included file_objects)

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -249,3 +249,33 @@ class FrontEndDbInterface(MongoInterfaceCommon):
     def rest_get_object_uids(database, offset, limit, query):
         uid_cursor = database.find(query, {'_id': 1}).skip(offset).limit(limit)
         return [result['_id'] for result in uid_cursor]
+
+    def find_missing_files(self):
+        uids_in_db = set()
+        parent_to_included = {}
+        for collection in [self.file_objects, self.firmwares]:
+            for result in collection.find({}, {'_id': 1, 'files_included': 1}):
+                uids_in_db.add(result['_id'])
+                parent_to_included[result['_id']] = set(result['files_included'])
+        for parent_uid, included_files in list(parent_to_included.items()):
+            included_files.difference_update(uids_in_db)
+            if not included_files:
+                parent_to_included.pop(parent_uid)
+        return parent_to_included
+
+    def find_missing_analyses(self):
+        missing_analyses = {}
+        query_result = self.firmwares.aggregate([
+            {'$project': {'temp': {'$objectToArray': '$processed_analysis'}}},
+            {'$unwind': '$temp'},
+            {'$group': {'_id': '$_id', 'analyses': {'$addToSet': '$temp.k'}}},
+        ], allowDiskUse=True)
+        for result in query_result:
+            firmware_uid, analysis_list = result['_id'], result['analyses']
+            query = {"$and": [
+                {"virtual_file_path.{}".format(firmware_uid): {"$exists": True}},
+                {"$or": [{"processed_analysis.{}".format(plugin): {"$exists": False}} for plugin in analysis_list]}
+            ]}
+            for entry in self.file_objects.find(query, {'_id': 1}):
+                missing_analyses.setdefault(firmware_uid, set()).add(entry['_id'])
+        return missing_analyses
--- a/src/test/integration/storage/test_db_interface_frontend.py
+++ b/src/test/integration/storage/test_db_interface_frontend.py
@@ -229,7 +229,7 @@ class TestStorageDbInterfaceFrontend(unittest.TestCase):
         assert missing_analyses == {}
 
         test_fw_1.processed_analysis['foobar'] = {'foo': 'bar'}
-        self.db_backend_interface.add_firmware(test_fw_1)
+        self.db_backend_interface.add_analysis(test_fw_1)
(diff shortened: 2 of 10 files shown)


dorpvom:
For reference my runtime on a smallish database with 18 images:

missing

dorpvom:
The endpoint is hidden as of now, since the admin dropdown is only shown if user management is enabled. So we have do discuss if that should always be shown or give another way of finding this functionality.

dorpvom:
Also, this seems to flag files as missing an analysis when the unpack depth is reached and the extraction is not run anymore. These are false positives.

src/test/integration/storage/test_db_interface_frontend.py
```diff
@@ -202,3 +202,34 @@ def test_get_specific_fields_for_multiple_entries(self):
...
+        test_fw_1.files_included.add(test_fo.uid)
+        test_fo.virtual_file_path = {test_fw_1.uid: ['|foo|bar|']}
+        self.db_backend_interface.add_firmware(test_fw_1)
+        self.db_backend_interface.add_file_object(test_fo)
+
+        missing_analyses = self.db_frontend_interface.find_missing_analyses()
+        assert missing_analyses == {}
+
+        test_fw_1.processed_analysis['foobar'] = {'foo': 'bar'}
+        self.db_backend_interface.add_firmware(test_fw_1)
```
dorpvom: there is an `add_analysis` function in the backend interface ; )

src/storage/db_interface_frontend.py
```diff
@@ -249,3 +249,33 @@ def rest_get_file_object_uids(self, offset, limit, query=None):
     def rest_get_object_uids(database, offset, limit, query):
         uid_cursor = database.find(query, {'_id': 1}).skip(offset).limit(limit)
         return [result['_id'] for result in uid_cursor]
+
+    def find_missing_files(self):
+        uids_in_db = set()
```
dorpvom: This will in a line 260 include all uids stored in the database. Correct?

src/web_interface/components/miscellaneous_routes.py
```diff
@@ -68,5 +70,39 @@ def _app_delete_firmware(self, uid):
...
             return render_template('error.html', message='Firmware not found in database: {}'.format(uid))
         with ConnectTo(AdminDbInterface, config=self._config) as sc:
-            deleted_virtual_file_path_entries, deleted_files = sc.delete_firmware(uid)
-        return render_template('delete_firmware.html', deleted_vps=deleted_virtual_file_path_entries, deleted_files=deleted_files, uid=uid)
+            deleted_virtual_path_entries, deleted_files = sc.delete_firmware(uid)
+        return render_template('delete_firmware.html', deleted_vps=deleted_virtual_path_entries, deleted_files=deleted_files, uid=uid)
+
+    @roles_accepted(*PRIVILEGES['delete'])
+    def _app_find_missing_analyses(self):
+        data = self._find_missing_files()
```
dorpvom: `missing_files = ...` ?

src/web_interface/components/miscellaneous_routes.py
```diff
@@ -68,5 +70,39 @@ def _app_delete_firmware(self, uid):
...
+            'missing_analyses_duration': self._get_duration(start),
+        }
+
+    @staticmethod
+    def _count_values(dictionary: Dict[str, Sized]) -> int:
+        return sum(len(e) for e in dictionary.values())
+
+    @staticmethod
+    def _get_duration(start: float) -> str:
+        return '{}s'.format(si_format(time() - start, precision=2))
```
dorpvom: Ah, do you really need an external library for this? That looks like an easy arithmetic operation

jstucke:
> The endpoint is hidden as of now, since the admin dropdown is only shown if user management is enabled. So we have do discuss if that should always be shown or give another way of finding this functionality.

What do you propose? Of course the user can still access the endpoint directly. Do we want to treat this as "hidden feature"?

> Also, this seems to flag files as missing an analysis when the unpack depth is reached and the extraction is not run anymore. These are false positives.

This is not really the fault of this feature, since the UIDs are part of the `files_included` of the file object. This should be fixed during unpacking.

src/web_interface/components/miscellaneous_routes.py
jstucke: Do you really think we should write our own library for formatting SI units if we could use an existing one from PyPI just as well?

src/test/integration/storage/test_db_interface_frontend.py
jstucke: changed

src/storage/db_interface_frontend.py
jstucke: that is correct, yes

src/web_interface/components/miscellaneous_routes.py
jstucke: I wanted to prevent passing the `render_template` function 9999 variables (this time)
renamed (slightly) to `template_data`

src/web_interface/components/miscellaneous_routes.py
dorpvom: Generally sure, take the existing code. But we're not talking writing a library here, are we? Isn't this just division by `10**x` for some static `x`. Maybe I'm at a miss here.

Issue #365, opened 2020-03-09
Script update_statistic.py failes when processing too many records


Harald883:
It seems that the script 'update_statistic.py' fails when processing too many database records:

`ubuntu@db-server:~/firmware-analysis/FACT_core/src$ ./update_statistic.py
[2020-03-09 11:23:39][update_statistic][INFO]: Try to start Mongo Server...
[2020-03-09 11:23:39][MongoMgr][INFO]: start local mongo database
Traceback (most recent call last):
  File "./update_statistic.py", line 50, in <module>
    sys.exit(main())
  File "./update_statistic.py", line 39, in main
    updater.update_all_stats()
  File "/home/ubuntu/firmware-analysis/FACT_core/src/statistic/update.py", line 42, in update_all_stats
    self.db.update_statistic('unpacking', self.get_unpacking_stats())
  File "/home/ubuntu/firmware-analysis/FACT_core/src/statistic/update.py", line 247, in get_unpacking_stats
    self.db.file_objects, '$processed_analysis.unpacker.entropy', unwind=True, match={'processed_analysis.unpacker.summary': 'unpacked'})).keys())
  File "/home/ubuntu/firmware-analysis/FACT_core/src/statistic/update.py", line 376, in _get_objects_and_count_of_occurrence_single_db
    tmp = get_objects_and_count_of_occurrence(database, object_path, unwind=unwind, match=merge_dict(match, self.match))
  File "/usr/local/lib/python3.6/dist-packages/common_helper_mongo/aggregate.py", line 84, in get_objects_and_count_of_occurrence
    result = list(collection.aggregate(pipeline))
  File "/usr/local/lib/python3.6/dist-packages/pymongo/collection.py", line 2380, in aggregate
    **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/pymongo/collection.py", line 2299, in _aggregate
    retryable=not cmd._performs_write)
  File "/usr/local/lib/python3.6/dist-packages/pymongo/mongo_client.py", line 1464, in _retryable_read
    return func(session, server, sock_info, slave_ok)
  File "/usr/local/lib/python3.6/dist-packages/pymongo/aggregation.py", line 148, in get_cursor
    user_fields=self._user_fields)
  File "/usr/local/lib/python3.6/dist-packages/pymongo/pool.py", line 613, in command
    user_fields=user_fields)
  File "/usr/local/lib/python3.6/dist-packages/pymongo/network.py", line 167, in command
[...]

PR #368, opened 2020-03-18
365 update statistics bugfix

jstucke:
addresses  #365
added `allowDiskUse=True` to `aggregate()` calls in order to (hopefully) fix the memory problems

### DIFF
--- a/src/statistic/update.py
+++ b/src/statistic/update.py
@@ -68,7 +68,7 @@ class StatisticUpdater:
                 pipeline_group={'_id': '$_id', 'size': {'$push': '$size'}},
                 additional_projection={'size': 1}
             )
-            query_result = [item['size'][0] for item in self.db.file_objects.aggregate(aggregation_pipeline)]
+            query_result = [item['size'][0] for item in self.db.file_objects.aggregate(aggregation_pipeline, allowDiskUse=True)]
             stats['number_of_unique_files'] = len(query_result)
             stats['total_file_size'] = sum(query_result)
             stats['average_file_size'] = avg(query_result)
@@ -96,7 +96,7 @@ class StatisticUpdater:
             additional_projection={'processed_analysis.exploit_mitigations.summary': 1})
 
         result_list_of_lists = [list(itertools.chain.from_iterable(d['exploit_mitigations']))
-                                for d in self.db.file_objects.aggregate(aggregation_pipeline)]
+                                for d in self.db.file_objects.aggregate(aggregation_pipeline, allowDiskUse=True)]
         result_flattened = list(itertools.chain.from_iterable(result_list_of_lists))
         result = self._count_occurrences(result_flattened)
         self.get_stats_nx(result, stats)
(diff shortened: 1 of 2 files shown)


src/statistic/update.py
```diff
@@ -273,53 +269,54 @@ def get_architecture_stats(self):
             pipeline_match={'processed_analysis.cpu_architecture.summary': {'$exists': True, '$not': {'$size': 0}}},
             additional_projection={'processed_analysis.cpu_architecture.summary': 1}
         )
-        result = [self._shorten_architecture_string(self._find_most_frequent_architecture(list(itertools.chain.from_iterable(item['architecture']))))
-                  for item in self.db.file_objects.aggregate(aggregation_pipeline)]
+        query_result = self.db.file_objects.aggregate(aggregation_pipeline, allowDiskUse=True)
+        result = [
+            self._shorten_architecture_string(self._find_most_frequent_architecture(list(itertools.chain.from_iterable(item['architecture']))))
+            for item in query_result
```
dorpvom: Why is the query_result not inline here anymore?

src/statistic/update.py
```diff
@@ -273,53 +269,54 @@ def get_architecture_stats(self):
...
+        ]
         return {'cpu_architecture': self._count_occurrences(result)}
 
     def _find_most_frequent_architecture(self, arch_list):
         try:
             arch_frequency = sorted(self._count_occurrences(arch_list), key=lambda x: x[1], reverse=True)
             return arch_frequency[0][0]
-        except Exception as e:
-            logging.error('Could not get arch frequency: {} {}'.format(sys.exc_info()[0].__name__, e))
+        except (AttributeError, KeyError, TypeError) as exception:
```
dorpvom: Applause :clap:

src/statistic/update.py
```diff
@@ -273,53 +269,54 @@ def get_architecture_stats(self):
...
     @staticmethod
-    def _count_occurrences(l):
-        return [(item, l.count(item)) for item in set(l)]
+    def _count_occurrences(result_list):
+        return [(item, result_list.count(item)) for item in set(result_list)]
 
     @staticmethod
-    def _shorten_architecture_string(s):
-        if s is None:
+    def _shorten_architecture_string(string):
```
dorpvom: I generally try to omit `string` as identifier since there is the built-in library is called `string`. But I guess it's fair game as long as you don't need that..

src/statistic/update.py
```diff
@@ -273,53 +269,54 @@ def get_architecture_stats(self):
...
-        except Exception as e:
-            logging.error('Could not get arch frequency: {} {}'.format(sys.exc_info()[0].__name__, e))
+        except (AttributeError, KeyError, TypeError) as exception:
+            logging.error('Could not get arch frequency: {} {}'.format(sys.exc_info()[0].__name__, exception))
             return None
 
     @staticmethod
-    def _count_occurrences(l):
-        return [(item, l.count(item)) for item in set(l)]
+    def _count_occurrences(result_list):
```
dorpvom: Can this be replaced by collections.Counter?

dorpvom:
Script runs smoothly on my machine. Jenkins should still be run though.

src/statistic/update.py
jstucke: I just thought this might be a bit more readable but I can change it back if you wish

src/statistic/update.py
jstucke: yes, fixed

PR #371, opened 2020-03-20
Added check for enum34, update copyright strings

0xricksanchez:
### Issue:

The FACT installation may break on systems having the [enum34](https://pypi.org/project/enum34/) package installed.

### Error:

```
AttributeError: module 'enum' has no attribute 'IntFlag'
```

### Root cause analysis:

Since python 3.4 there's a standard library `enum` module, so we should uninstall enum34,
which is no longer compatible with the `enum` in the standard library since `enum.IntFlag` was added in [python 3.6](https://docs.python.org/3/library/enum.html).

### Solution 

```
pip3 uninstall -y enum34
```
However, automatic uninstalling may cause unexpected side effects for the affected users still using the package. As a result, this PR modifies the `pre_install.sh` script to exit preemptively if the package is found

### DIFF
--- a/src/compile_yara_signatures.py
+++ b/src/compile_yara_signatures.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
 '''
     Firmware Analysis and Comparison Tool (FACT)
-    Copyright (C) 2015-2018  Fraunhofer FKIE
+    Copyright (C) 2015-2020  Fraunhofer FKIE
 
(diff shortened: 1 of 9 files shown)


0xricksanchez:
Also included this part to the [Troubleshooting wiki](https://github.com/fkie-cad/FACT_core/wiki/troubleshooting-runtime)

PR #372, opened 2020-03-23
Sd 896 yara search show firmware

weidenba:


src/web_interface/components/database_routes.py
```diff
@@ -189,23 +207,24 @@ def _app_show_binary_search_results(self):
...
-        for rule in uid_dict:
-            with ConnectTo(FrontEndDbInterface, self._config) as connection:
-                firmware_list = [
-                    connection.firmwares.find_one(uid) or connection.file_objects.find_one(uid)
-                    for uid in uid_dict[rule]
-                ]
-                firmware_dict[rule] = sorted(connection.get_meta_list(firmware_list))
-        return firmware_dict
+    def _do_binary_search_query(self, binary_search_results: list, yara_rules: str) -> str:
+        query = '{"_id": {"$in": ' + str(binary_search_results).replace('\'', '"') + '}}'
```
jstucke: is it OK to use `+` for string formatting again?
if not: `'{{"_id": {{"$in": {}}}}}'.format(str(binary_search_results).replace('\'', '"'))`

src/web_interface/components/database_routes.py
```diff
@@ -189,23 +207,24 @@ def _app_show_binary_search_results(self):
...
-        firmware_dict = {}
-        for rule in uid_dict:
-            with ConnectTo(FrontEndDbInterface, self._config) as connection:
-                firmware_list = [
-                    connection.firmwares.find_one(uid) or connection.file_objects.find_one(uid)
-                    for uid in uid_dict[rule]
-                ]
-                firmware_dict[rule] = sorted(connection.get_meta_list(firmware_list))
-        return firmware_dict
+    def _do_binary_search_query(self, binary_search_results: list, yara_rules: str) -> str:
```
jstucke: we are not doing/executing any searches here, right? So the name should probably be something like "save/store binary search query"?

src/web_interface/components/database_routes.py
weidenba: to be honest, in this case it is much more readable to use + string formating. Therefore, I would prefer to leave it this way.

PR #373, opened 2020-03-24
Sd 871 known password tag

weidenba:
add an analysis tag if users and passwords plug-in finds a password

### DIFF
--- a/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
+++ b/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import re
-import sys
 from contextlib import suppress
 from tempfile import NamedTemporaryFile
 
@@ -60,8 +59,8 @@ class AnalysisPlugin(AnalysisBasePlugin):
                     result[key]['password-hash'] = entry[1].decode(encoding='utf_8', errors='replace')
                     cracked_pw = self._crack_hash(entry, result, key)
                     result[key]['cracked'] = bool(cracked_pw)
-            except Exception as excep:
-                logging.error('Invalid Format: {} - {}'.format(sys.exc_info()[0].__name__, excep))
+            except (IndexError, AttributeError, TypeError):
+                logging.error('Invalid Format:', exc_info=True)
         return result
 
     def _crack_hash(self, passwd_entry, result_dict, key):
--- a/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
+++ b/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
@@ -24,6 +24,10 @@ class AnalysisPlugin(AnalysisBasePlugin):
 
     wordlist_path = os.path.join(get_src_dir(), 'bin/passwords.txt')
 
+    def __init__(self, plugin_administrator, config=None, recursive=True):
+        self.config = config
+        super().__init__(plugin_administrator, config=config, recursive=recursive, no_multithread=True, plugin_path=__file__)
+
     def process_object(self, file_object):
         if self.NAME not in file_object.processed_analysis:
             file_object.processed_analysis[self.NAME] = {}
--- a/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
+++ b/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
@@ -9,6 +9,7 @@ from common_helper_process import execute_shell_command
 
 from analysis.PluginBase import AnalysisBasePlugin
 from helperFunctions.fileSystem import get_src_dir
+from helperFunctions.tag import TagColor
 
 
 class AnalysisPlugin(AnalysisBasePlugin):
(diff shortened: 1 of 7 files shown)


dorpvom:
View has to ignore tags field otherwise tag will be shown as result in listing:

root_tag

weidenba:
> View has to ignore tags field otherwise tag will be shown as result in listing:
> 
> root_tag

The view should be correct. However, there seems to be a problem with the view sync mechanism. Updated views stored in the plug-in within the backend are not synchronized to the frontend. I am going to have a look at this today.

weidenba:
View sync issue should be fixed. @dorpvom you can rereview if you like

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -9,51 +9,47 @@
...
 class AnalysisPlugin(AnalysisBasePlugin):
     '''
-    This Plugin trys to find and crack passwords
+    This plug-in tries to find and crack passwords
     '''
     NAME = 'users_and_passwords'
     DEPENDENCIES = []
     MIME_BLACKLIST = ['audio', 'filesystem', 'image', 'video']
     DESCRIPTION = 'search for UNIX and httpd password files, parse them and try to crack the passwords'
-    VERSION = '0.4.1'
```
dorpvom: Hm. Okay.

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -9,51 +9,47 @@
...
-        If you want to propagate results to parent objects store a list of strings 'summary' entry of your result dict
-        '''
         if self.NAME not in file_object.processed_analysis:
             file_object.processed_analysis[self.NAME] = {}
         file_object.processed_analysis[self.NAME]['summary'] = []
 
         for passwd_regex in [
-            b'[a-zA-Z][a-zA-Z0-9_-]{2,15}:[^:]?:\\d+:\\d*:[^:]*:[^:]*:[^\n ]*',
-            b'[a-zA-Z][a-zA-Z0-9_-]{2,15}:\\$[^\\$]+\\$[^\\$]+\\$[a-zA-Z0-9\\./]{16,128}'
+                b'[a-zA-Z][a-zA-Z0-9_-]{2,15}:[^:]?:\\d+:\\d*:[^:]*:[^:]*:[^\n ]*',
```
dorpvom: I guess pylint wants this? PyCharm has a different opinion.

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -63,9 +59,9 @@ def _generate_analysis_entry(self, passwd_entries):
                 if entry[1][0] == ord('$'):
                     result[key]['password-hash'] = entry[1].decode(encoding='utf_8', errors='replace')
                     cracked_pw = self._crack_hash(entry, result, key)
-                    result[key]['cracked'] = True if cracked_pw else False
-            except Exception as e:
-                logging.error('Invalid Format: {} - {}'.format(sys.exc_info()[0].__name__, e))
+                    result[key]['cracked'] = bool(cracked_pw)
+            except Exception as excep:
+                logging.error('Invalid Format: {} - {}'.format(sys.exc_info()[0].__name__, excep))
```
dorpvom: This would be much slimmer and you can drop the `sys` import:
```python
except Exception:
    logging.error('Invalid Format:', exc_info=True)
```
Also: Why can't we have a less general exception?

dorpvom:
```python
except (IndexError, AttributeError, TypeError):
```
should do I guess.

src/plugins/analysis/users_and_passwords/test/test_plugin_password_file_analyzer.py
```diff
@@ -23,25 +24,26 @@ def test_process_object(self):
...
+            self.assertIn(item, results['summary'])
         self.assertIn('password-hash', results['max'])
         self.assertIn('password', results['max'])
         self.assertEqual(results['max']['password'], 'dragon')
         self.assertIn('password-hash', results['johndoe'])
         self.assertIn('password', results['johndoe'])
         self.assertEqual(results['johndoe']['password'], '123456')
+        self.assertEqual(results['tags']['johndoe_123456']['value'], 'Password: johndoe:123456')
 
         test_file = FileObject(file_path=os.path.join(TEST_DATA_DIR, 'passwd.bin'))
```
dorpvom: It would be nice to split this test in two, since here is a logical break

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
weidenba: yes pylint wants it this way ;-)

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
weidenba: > This would be much slimmer and you can drop the `sys` import:
> 
> ```python
> except Exception:
>     logging.error('Invalid Format:', exc_info=True)
> ```
We should have a discussion about our "default" error logging and we should document it somewhere

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
weidenba: there was a never published minor version in between ;-)

PR #378, opened 2020-04-01
added analysis info for unpacking depth

jstucke:
- added unpacker analysis entry in case unpacking was skipped because the analysis depth was reached
- added test


src/test/unit/unpacker/test_unpacker.py
```diff
@@ -62,12 +62,19 @@ def main_unpack_check(self, test_object, number_unpacked_files, first_unpacker):
...
+    def check_depths_of_children(self, parent, extracted_files):
         for item in extracted_files:
             self.assertEqual(item.depth, parent.depth + 1, 'depth of child not correct')
 
     def test_main_unpack_function(self):
         test_file = FileObject(file_path=os.path.join(get_test_data_dir(), 'container/test.zip'))
         self.main_unpack_check(test_file, 3, '7z')
+
+    def test_unpacking_depth_reached(self):
+        test_file = FileObject(file_path=os.path.join(get_test_data_dir(), 'container/test.zip'))
```
weidenba, line 76: Is there a reason why we use os.path instead of a Path object?

jstucke: naturally, I blindly copied that -- fixed

PR #388, opened 2020-04-09
added error handling for missing singature files

jstucke:
- added error handling and shutdown for FACT if signatures are missing

### DIFF
--- a/src/start_fact.py
+++ b/src/start_fact.py
@@ -21,6 +21,7 @@ import logging
 import os
 import signal
 import sys
+from pathlib import Path
 from shlex import split
 from subprocess import Popen, TimeoutExpired
 from time import sleep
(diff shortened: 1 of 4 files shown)


src/start_fact.py
```diff
@@ -45,27 +45,28 @@ def _evaluate_optional_args(args):
...
-    if os.path.exists(script_path):
-        logging.info('starting {}'.format(component))
-        optional_args = _evaluate_optional_args(args)
-        command = '{} -l {} -L {} -C {} {}'.format(
-            script_path, config['Logging']['logFile'], config['Logging']['logLevel'], args.config_file, optional_args
-        )
-        p = Popen(split(command))
-        return p
-    else:
+    if not os.path.exists(script_path):
```
weidenba, line 48: if we change the line, we should switch from os.path to Path, right?

jstucke: fixed

PR #389, opened 2020-04-14
Software signatures improved and added

weidenba:
Improved Signature of ThreadX and added Signature for eCos

### DIFF
--- a/src/plugins/analysis/software_components/signatures/http_server.yara
+++ b/src/plugins/analysis/software_components/signatures/http_server.yara
@@ -36,3 +36,18 @@ rule nginx
 	condition:
 		$a and no_text_file
 }
+
+rule eCosWebServer
+{
+	meta:
+		software_name = "eCos Embedded Web Server"
+		open_source = true
+		website = "https://www.ecoscentric.com"
+		description = "Web-Server"
+		format_string = true
+	strings:
+		$a = /eCos Embedded Web Server/ nocase ascii wide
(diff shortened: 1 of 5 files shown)


src/plugins/analysis/software_components/signatures/http_server.yara
```diff
@@ -36,3 +36,18 @@ rule nginx
...
+rule eCosWebServer
+{
+	meta:
+		software_name = "eCos Embedded Web Server"
+		open_source = true
+		website = "https://www.ecoscentric.com"
+		description = "Web-Server"
+		format_string = true
+	strings:
+		$a = /eCos Embedded Web Server/ nocase ascii wide
```
jstucke, line 49: `/.../` is used for regular expressions, right? So `"..."` would be better here i guess? (though it probably doesn't make a difference in the end)

src/plugins/analysis/software_components/signatures/os.yara
```diff
@@ -138,4 +138,18 @@ rule Contiki
...
+rule eCos
+{
+	meta:
+		software_name = "eCos"
+		open_source = false
+		website = "https://www.ecoscentric.com"
+		description = "Real Time Operating System"
+		format_string = true
+	strings:
+		$a = /eCos Release: %d.%d.%d/ nocase ascii wide
```
jstucke, line 152: again not really a regex

Issue #282, opened 2019-08-02
update_variety_data script should ignore "skip" and "file_system_flag" entries


weidenba:
These entries are just for internal use. There is no use of these for a user.

jstucke:
fixed in #399

PR #399, opened 2020-04-20
removed uninteresting keys from variety result

jstucke:
- removed uninteresting keys from `varietyjs` results database

### DIFF
--- a/src/update_variety_data.py
+++ b/src/update_variety_data.py
@@ -18,12 +18,12 @@
 '''
 
 import logging
-import os
 import sys
+from pathlib import Path
 from time import time
 
 from common_helper_filter import time_format
-from common_helper_process import execute_shell_command_get_return_code
+from common_helper_process import execute_shell_command, execute_shell_command_get_return_code
 
 from helperFunctions.fileSystem import get_src_dir
 from helperFunctions.program_setup import program_setup
@@ -34,7 +34,7 @@ PROGRAM_DESCRIPTION = 'Initialize or update database structure information used 
 
 
 def _create_variety_data(config):
-    full_variety_path = os.path.join(get_src_dir(), config['data_storage']['variety_path'])
+    varietyjs_script_path = Path(get_src_dir()) / config['data_storage']['variety_path']
     mongo_call = (
         'mongo --port {mongo_port} -u "{username}" -p "{password}" --authenticationDatabase "admin" '.format(
             mongo_port=config['data_storage']['mongo_port'],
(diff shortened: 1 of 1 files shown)


dorpvom:
How / Why does the mongo query work? Aren't the keys deleted permanently?

jstucke:
> How / Why does the mongo query work? Aren't the keys deleted permanently?

This addresses #282 so yes they are removed permanently so that they don't appear on the advanced search page

dorpvom:
> > How / Why does the mongo query work? Aren't the keys deleted permanently?
> 
> This addresses #282 so yes they are removed permanently so that they don't appear on the advanced search page

But only in the variety table, not the real file_objects table?

jstucke:
> But only in the variety table, not the real file_objects table?

yes

src/update_variety_data.py
```diff
@@ -35,16 +35,27 @@
 
 def _create_variety_data(config):
     full_variety_path = os.path.join(get_src_dir(), config['data_storage']['variety_path'])
```
weidenba, line 37: if you migrated `os` -> `Path`, I would have merged it immidiatly :wink:

src/update_variety_data.py
```diff
@@ -35,16 +35,27 @@
...
+        )
+    )
+    output, return_code = execute_shell_command_get_return_code(
+        '{mongo_call} {database} --eval "var collection = \'file_objects\', persistResults=true" {script_path}'.format(
+            mongo_call=mongo_call,
+            database=config['data_storage']['main_database'],
             script_path=full_variety_path),
         timeout=None
     )
+    execute_shell_command_get_return_code(
```
weidenba, line 52: Should we check this return code as well? Or does it not matter at all and we could use the `execute_shell_command` method?

src/update_variety_data.py
```diff
@@ -35,16 +35,27 @@
...
-        '--eval "var collection = \'file_objects\', persistResults=true" {script_path}'.format(
+    mongo_call = (
+        'mongo --port {mongo_port} -u "{username}" -p "{password}" --authenticationDatabase "admin" '.format(
             mongo_port=config['data_storage']['mongo_port'],
             username=config['data_storage']['db_admin_user'],
             password=config['data_storage']['db_admin_pw'],
-            main_database=config['data_storage']['main_database'],
+        )
+    )
+    output, return_code = execute_shell_command_get_return_code(
```
weidenba, line 45: should we abort if return_code is not `0`

src/update_variety_data.py
jstucke: switched to `execute_shell_command`

src/update_variety_data.py
jstucke: I don't know... that's pretty much all that happens in the script anyway

src/update_variety_data.py
jstucke: changed to Path

src/update_variety_data.py
jstucke: And the return code is used as the exit code of the whole script. Of course we could add some error messages in case of an error. I tried to add as few lines as possible though because the script isn't tested.

PR #403, opened 2020-04-22
Analysis header redesign

dorpvom:
Changes:
- Buttons are moved to top in a toolbar style design
- General information is more individually set to allow more compact design
- Collapsible lists are redesigned to directly show first item and remove list-group borders
- Some other minor changes and new and fix tests


src/web_interface/filter.py
```diff
@@ -366,3 +379,16 @@ def is_not_mandatory_analysis_entry(item: str, additional_entries: Optional[List
         item not in ['analysis_date', 'plugin_version', 'skipped', 'summary', 'system_version', 'tags']
         and (additional_entries is None or item not in additional_entries)
     )
+
+
+def random_collapse_id():
+    return ''.join((random.choice(ascii_letters) for _ in range(10)))
```
jstucke, line 385: why not use `random.choices()`?

src/web_interface/filter.py
```diff
@@ -366,3 +379,16 @@ def is_not_mandatory_analysis_entry(item: str, additional_entries: Optional[List
...
+
+
+def version_links(versions, selected_analysis=None):
```
jstucke, line 388: took me a moment to understand what the filter does based on the name. maybe `other_version_links` would be clearer?

src/web_interface/filter.py
```diff
@@ -366,3 +379,16 @@ def is_not_mandatory_analysis_entry(item: str, additional_entries: Optional[List
...
+    if selected_analysis:
+        template = '<a href="/analysis/{{}}/{}">{{}}</a>'.format(selected_analysis)
+    else:
+        template = '<a href="/analysis/{}">{}</a>'
+
+    return [template.format(version['_id'], version['version']) for version in versions]
```
jstucke, line 394: it seems to me `version` is not a particularly good name for the variable here. maybe `other_fw` or something idk

src/web_interface/filter.py
jstucke: > *New in version 3.6.*

nevermind ...

src/web_interface/templates/generic_view/nice_fo_list.html
```diff
@@ -1,7 +1,15 @@
-<div class="list-group" style="font-size: 15px;">
+{% from "generic_view/file_object_macros.html" import file_information_span, collapse_head %}
```
jstucke, line 1: `collapse_head` is imported but not used?

src/test/unit/web_interface/test_filter.py
```diff
@@ -361,3 +361,34 @@ def test_comment_out_regex_meta_chars(input_data, expected_result):
...
+
+def test_random_collapse_id():
+    collapse_id = random_collapse_id()
+    assert isinstance(collapse_id, str)
+    assert not any(collapse_id.startswith(digit) for digit in '0123456789')
+
+
+def test_list_group_collapse():
+    from web_interface.frontend_main import WebFrontEnd  # pylint: disable=import-outside-toplevel
+    from test.common_helper import get_config_for_testing  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
dorpvom: This test is up for debate.
It's not a unit test but an integration test. There are no integration tests for filter.py yet though. So I used this to signify that there is something to discuss.
So. Do you think this is correctly placed here or should be moved to a new test module in `test/integration`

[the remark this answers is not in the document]
jstucke: I could live with the former but would prefer the latter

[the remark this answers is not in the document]
jstucke: (but I could only live with the former if there are no imports in the function for no reason)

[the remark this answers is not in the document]
dorpvom: Yoda voice: "Hm, No reason you say.."

PR #407, opened 2020-04-29
firmware analysis status

jstucke:
- added option to track the analysis status of all currently analyzed firmware images (incl. when all files are fully analyzed)
- added error messages for failed analyses in the web interface

### DIFF
--- a/src/scheduler/Analysis.py
+++ b/src/scheduler/Analysis.py
@@ -3,7 +3,7 @@ from concurrent.futures import ThreadPoolExecutor
 from configparser import ConfigParser
 from copy import copy
 from distutils.version import LooseVersion
-from multiprocessing import Queue, Value
+from multiprocessing import Manager, Queue, Value
 from queue import Empty
 from time import sleep, time
 from typing import List, Optional, Set, Tuple
@@ -17,6 +17,7 @@ from helperFunctions.plugin import import_plugins
 from helperFunctions.process import ExceptionSafeProcess, check_worker_exceptions
 from helperFunctions.tag import add_tags_to_object, check_tags
 from objects.file import FileObject
+from objects.firmware import Firmware
 from storage.db_interface_backend import BackEndDbInterface
 
 MANDATORY_PLUGINS = ['file_type', 'file_hashes']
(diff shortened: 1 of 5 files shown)


src/web_interface/templates/system_health.html
```diff
@@ -73,21 +73,37 @@ <h3>{{component['name']}} machine status</h3>
                 <div class="col-sm-6 col-md-4 col-lg-3">
                     <h3>{{ component['name'] }} workload</h3>
                     <table class="table table-responsive-md table-hover table-bordered">
+                        {% set current_analyses = component['analysis']['current_analyses'] %}
+                        <tr>
+                            <th class="table-head-light" colspan=3 style="text-align: center; padding:5px; color: #495057">Currently Analyzed Firmware</th>
+                        </tr>
+                        {% if current_analyses %}
```
dorpvom, line 80: We could make the complete block optional. So drop the Empty and move the if above the "Currently Analyzed Firmware" heading.

src/scheduler/Analysis.py
```diff
@@ -352,6 +360,8 @@ def _handle_analysis_tags(self, fw, plugin):
     def check_further_process_or_complete(self, fw_object):
         if not fw_object.scheduled_analysis:
             logging.info('Analysis Completed:\n{}'.format(fw_object))
+            if isinstance(fw_object, FileObject):
```
dorpvom, line 363: What else can this be? Given that it has to have the `scheduled_analysis` attribute by now

src/scheduler/Analysis.py
```diff
@@ -383,3 +393,23 @@ def _get_cumulative_remaining_dependencies(self, scheduled_analyses: Set[str]) -
...
+            for parent in self._find_currently_analyzed_parents(fw_object):
+                union = set(fw_object.files_included).union(self.currently_running[parent])
+                self.currently_running[parent] = list(union)
+
+    def _remove_from_current_analyses(self, fw_object: Union[Firmware, FileObject]):
+        for parent in self._find_currently_analyzed_parents(fw_object):
+            self.currently_running[parent] = [uid for uid in self.currently_running[parent] if uid != fw_object.uid]
+            if len(self.currently_running[parent]) == 0:
+                self.currently_running.pop(parent)
+                logging.info('analysis of firmware {} completed'.format(parent))
```
dorpvom, line 412: Capital first letter

src/web_interface/templates/system_health.html
jstucke: if you think that's better ... changed

src/scheduler/Analysis.py
jstucke: the `Firmware` objects are also analyzed and also have the attribute, right?

src/scheduler/Analysis.py
jstucke: changed

PR #411, opened 2020-04-30
System health redesign

dorpvom:
- Added plugin information to intercom object
- Replaced plugin table in sys health with plugin cards
- Replaced machine tables by more compact cards


weidenba:
It looks quite good. Maybe we should visually separate the plugin catalog stuff.

PR #417, opened 2020-05-18
Mongodb docker

jstucke:


src/install/db.py
```diff
@@ -26,31 +14,16 @@ def _get_db_directory():
...
+def main():
     logging.info('Setting up mongo database')
-
-    if distribution in ['xenial', 'debian']:
-        _add_mongo_mirror(distribution)
-        apt_update_sources()
-        apt_install_packages('mongodb-org')
-    else:
-        apt_install_packages('mongodb')
+    _, return_code = execute_shell_command_get_return_code('docker pull mongo:3-xenial')
```
weidenba, line 19: Why xenial with EOL next year?

src/storage/mongodb_docker.py
```diff
@@ -0,0 +1,102 @@
...
+def get_docker_network(docker_client: DockerClient) -> Network:
+    try:
+        network = docker_client.networks.get(FACT_MONGODB_NETWORK)
+    except NotFound:
+        network = create_docker_network(docker_client)
+    return network
+
+
+def create_docker_network(client: DockerClient) -> Network:
+    ipam_pool = docker.types.IPAMPool(subnet='192.168.27.0/24', gateway='192.168.27.1')
```
weidenba, line 87: can this interfere with other local networks? Should we make this IP-Adresses configureable somehow?

src/install/db.py
jstucke: sadly newer versions like bionic are only available with MongoDB 4
I tried to build a custom container but was not able to connect to it (for still unknown reasons)

src/storage/mongodb_docker.py
jstucke: made ip, subnet and gateway configurable

src/install/db.py
jstucke: replaced with a docker image based on debian stretch

jstucke:
replaced by https://github.com/fkie-cad/FACT_docker

PR #426, opened 2020-06-04
Ajax statistics loading

Geierhaas:


src/web_interface/components/ajax_routes.py
```diff
@@ -162,3 +166,27 @@ def _ajax_get_summary(self, uid, selected_analysis):
             firmware = sc.get_object(uid, analysis_filter=selected_analysis)
             summary_of_included_files = sc.get_summary(firmware, selected_analysis)
         return render_template('summary.html', summary_of_included_files=summary_of_included_files, root_uid=uid, selected_analysis=selected_analysis)
+
+    @roles_accepted(*PRIVILEGES['status'])
+    def _get_asynchronous_stats(self):
```
dorpvom, line 171: Name does not fit anymore. AJAX (Asynchronous JavaScript and XML) already implies asynchronicity so a name such as `_get_general_stats` fits better

src/web_interface/templates/home.html
```diff
@@ -67,10 +67,54 @@ <h4 class="mb-3">General FACTs</h4>
...
+                    }
+
+                    running_node.innerHTML = system_data.number_of_running_analyses;
+                });
+            }
+
+            load_ajax_stats();
+            setInterval(function(){
+                load_ajax_stats()
+            }, 10000);
```
dorpvom, line 116: This should be configurable. You can do this in three steps:
1. add config value in **main.cfg**
2. pipe this config value through to template from inside the flask endpoint definition (in **miscellaneous_routes.py**)
3. use the value here with the jinja bracket notation: `{{ timeout }}`

PR #434, opened 2020-06-26
Fixing bug that prevented comparisons

nstarke:
We noticed a bug where comparisons were not being added to the
database properly due to the 'entropy' key not being present in
the 'unpacker' dict, so we added some handling so that we could
process comparisons correctly.


dorpvom:
CI is [running](https://ci.weidenba.de/job/bionic_external/2/). On success I'll merge. Thanks for the contribution!

PR #443, opened 2020-06-30
Bugfix for download link template

nstarke:
The download link template was broken and did not allow
for a user to download the appropriate file.  This commit
fixes the issue

### DIFF
--- a/src/web_interface/templates/generic_view/file_object_macros.html
+++ b/src/web_interface/templates/generic_view/file_object_macros.html
@@ -13,7 +13,7 @@
                 , Type: {{ file_object['mime-type'] }}
             {% endif %}
         </span>
-        <a href="location.href='/download/{{ file_object.uid }}'"><i class="fas fa-download"></i></a>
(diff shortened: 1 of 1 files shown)


src/web_interface/templates/generic_view/file_object_macros.html
```diff
@@ -13,7 +13,7 @@ <h6>{{ file_object.current_virtual_path[0] | replace_uid_with_hid(root_uid=root_
                 , Type: {{ file_object['mime-type'] }}
             {% endif %}
         </span>
-        <a href="location.href='/download/{{ file_object.uid }}'"><i class="fas fa-download"></i></a>
+        <a href="#" onclick="location.href='/download/{{ file_object.uid }}'"><i class="fas fa-download"></i></a>
```
jstucke, line 16: This should be even simpler (I don't know why there was a `location.href` in a hyperlink tag in the first place):
```suggestion
        <a href="/download/{{ file_object.uid }}"><i class="fas fa-download"></i></a>
```

PR #463, opened 2020-07-30
Pie charts

jstucke:


src/web_interface/static/js/fact_statistics.js
```diff
@@ -82,3 +88,42 @@ function create_horizontal_bar_chart(canvas_id, chart_data, link, value_percenta
 
     return BarChart;
 }
+
+function create_pie_chart(canvas_id, chart_data, link, value_percentage_present_flag) {
+    var ctx = document.getElementById(canvas_id);
```
dorpvom, line 93: Here's obvious code duplication. Is there a nice javascript way to de-duplicate this?

src/web_interface/static/js/fact_statistics.js
```diff
@@ -82,3 +88,42 @@ function create_horizontal_bar_chart(canvas_id, chart_data, link, value_percenta
...
+                    labels: {
+                        boxWidth: 20,
+                        fontSize: 10
+                    }
+            }
+        }
+        }
+    );
+
+    document.getElementById(canvas_id).onclick = function(evt){
```
dorpvom, line 121: Code duplication. See above.

src/helperFunctions/web_interface.py
```diff
@@ -16,12 +17,17 @@
...
 
+def get_color_list(number, limit=10):
+    color_map = cm.get_cmap('rainbow')
+    color_list = [colors.rgb2hex(color_map(i)) for i in range(32, 256, 22)]
+    return color_list[:min(number, limit)]
 
-def get_color_list(number, limit=15):
-    return ([BS_PRIMARY, BS_SECONDARY, ] * 8)[:number if number <= limit else limit]
+
+def get_alternating_color_list(number, limit=10):
```
dorpvom, line 27: Maybe in this case, a comment actually makes sense. Without context, the magic number 8 (7) is rather confusing.

src/helperFunctions/web_interface.py
```diff
@@ -4,6 +4,7 @@
 from datetime import timedelta
 
 from common_helper_files import get_binary_from_file
+from matplotlib import cm, colors
```
dorpvom, line 7: We have a matplotlib dependency elsewhere, right? Otherwise this needs an addition to the frontend install script (while keeping in mind that using matplotlib for only this color map would be a rather big overhead).

jstucke:
The links on the new donut charts seem to be broken:
```json
{
  "processed_analysis.exploit_mitigations.summary": {
    "$regex": "Router"
  }
}
``` 
On closer inspection, every search is on the `exploit_mitigations.summary` field (and not e.g. `device_class` in this case)

dorpvom:
I've found a strange behaving plot:

pie_chart

Why would squashFS be listed in the legend and not be shown in the plot?

PR #470, opened 2020-08-20
Sd425 helper sphinx docs

jstucke:
New sphinx docs
- install requirements in `docsrc/doc_requirements.txt`
- build with `docsrc/build.sh` or `make -C docsrc html`


docsrc/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
dorpvom, line 1: Is this not outdated now? I think we have the github actions for that

docsrc/conf.py
```diff
@@ -0,0 +1,54 @@
+# Configuration file for the Sphinx documentation builder.
```
dorpvom, line 1: No idea how to review this ..

src/helperFunctions/uid.py
```diff
@@ -1,46 +1,47 @@
...
 
 from helperFunctions.dataConversion import make_bytes
 from helperFunctions.hash import get_sha256
 
 
-def create_uid(input_data):
+def create_uid(input_data: AnyStr) -> str:
     '''
-    creates an unique identifier: SHA256_SIZE
+    generate a UID (unique identifier) SHA256_SIZE for a string containing data
```
dorpvom, line 10: Hm. I thought we'd encounter bytes here not strings. What would the sense be to generate this based on a string?

dorpvom (changes requested):
Maybe there's already legacy stuff here? Do we even need the stuff in docs (aside from the pdfs) anymore?

src/helperFunctions/uid.py
jstucke: this is actually used e.g. in `db_interface_frontend_editing.py` with a string but changing that is probably out of scope. Changed the doc string

dorpvom (approved):

PR #473, opened 2020-08-21
removed entropy and documented object_storage

dorpvom:


docsrc/modules/helperFunctions.rst
```diff
@@ -5,3 +5,4 @@ helperFunctions
    :maxdepth: 4
 
    helperFunctions.uid
+   helperFunctions.object_storage
```
jstucke, line 8: nitpick: we should order the items alphabetically

PR #476, opened 2020-08-24
fixed typing and added docs

jstucke:


src/helperFunctions/compare_sets.py
```diff
@@ -1,7 +1,16 @@
-from typing import Iterable, List, T, Tuple
+from typing import Iterable, Iterator, List, Tuple, TypeVar
 
+# Generics (not intended for export)
+_T = TypeVar('_T')
```
dorpvom, line 4: Why is this useful or rather more useful then just using `list` in the definitions above?

src/helperFunctions/compare_sets.py
```diff
@@ -10,14 +19,37 @@ def remove_duplicates_from_unhashable(unhashable_list: List[T]) -> List[T]:
 
 
 def remove_duplicates_from_list(list_):
+    '''
```
dorpvom, line 22: Actually why do we have two of these? We should really only have this function and call the unhashable variant on an error regarding unhashability ...

src/helperFunctions/compare_sets.py
```diff
@@ -10,14 +19,37 @@ def remove_duplicates_from_unhashable(unhashable_list: List[T]) -> List[T]:
...
+    Remove duplicates from a list (by converting it to a set and back).
+
+    .. caution::
+        Only works with lists of hashable objects!
+
+    :param list_: the input list (possibly) containing duplicates
```
dorpvom, line 28: We don't lint these (yet), but from what I've read, ending on punctuation is sort of best practice.

src/helperFunctions/compare_sets.py
```diff
@@ -10,14 +19,37 @@ def remove_duplicates_from_unhashable(unhashable_list: List[T]) -> List[T]:
...
+    '''
+    Iterate over each element of an iterable object (e.g. a list) and also get all other (remaining) elements
+    from the object.
+
+    :param iterable: the object that will be iterated over
+    :return: a generator with tuples containing the item and the rest (all other elements)
+    '''
     for element in iterable:
         rest = [e for e in iterable if e != element]
         yield element, rest
```
dorpvom, line 55: This should be collapsed to one line (while you're at it)

dorpvom (changes requested):

src/helperFunctions/compare_sets.py
jstucke: Without this, it would not be clear that the output list contains objects of the same type as the input (also for the IDE)

src/helperFunctions/compare_sets.py
jstucke: I agree -> changed

src/helperFunctions/compare_sets.py
jstucke: changed and also added capital letters at the beginning

src/helperFunctions/compare_sets.py
jstucke: reduced it to two lines but I would prefer to keep it a generator, because the lists may be big and it's more efficient this way

dorpvom: That's what I meant. One (yield) line inside the for-loop.

dorpvom (approved):

PR #478, opened 2020-08-25
Sd1240 plugin doc js

jstucke:
- added readme for building the docs
- added helperFunctions.mongo_task_conversion module docs

### DIFF
--- a/docsrc/README.md
+++ b/docsrc/README.md
@@ -1,0 +1,28 @@
+# Building the FACT Documentation Locally
+
+## Installing the Requirements
+
+```shell script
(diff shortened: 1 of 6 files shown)


dorpvom (approved):

dorpvom:
I have only general change requests. Might be handled in extra PR: 
- `get_uid_of_analysis_task` and `get_uploaded_file_binary` seem to be interal functions. Don't see why they should be public.
- `is_sanitized_entry` does not seem to fit in this module. Better in _helperFunctions/web_interface.py_?

Issue #477, opened 2020-08-25
How to repair/uninstall failed installation?


jahatfi:
I tried to install FACT on Kali 2020 but it's stuck on the "Install static jstree content" step.  If I can't get it to work I'd like to clean up all the files to recover my storage space, but I don't see an uninstall script.  Where would I find that?

dorpvom:
Good question / issue. We haven't thought about an uninstall script yet, and as we have a lot of dependencies, there is no quick way of constructing one.
I'll put that on the to do list, but I might be a while before we get to it.

Meanwhile, can you expand on the jstree problem? 
The internal steps are: (from FACT_core root folder)

```bash
cd src/web_interface/static
wget https://github.com/vakata/jstree/zipball/3.3.9
unzip 3.3.9
rm 3.3.9
rm -rf ./web_js/jstree/vakata*
mv vakata* web_js/jstree
```

jahatfi:
It's stuck on the step as mentioned.  I killed it with ctrl+C and here's the output in case it helps:
![Capture2](https://user-images.githubusercontent.com/12942571/91317452-10b3aa80-e788-11ea-8475-91c4568d6349.PNG)

`

dorpvom:
Quick question: Did you use `root` user for installation and swiched to `user` user afterwards? That would possibly explain the behavior, since the `rm -rf ` would then probably fail. On the other hand this should lead to an error and not a _hanging-idefinitely_.
I'd be interested in which of the commands listed above (wget, unzip, ...) exactly leads to the _hanging_. That's sadly not transparent from the error log due to the implementation of the installation process.

jahatfi:
I think I tried both with and without *sudo*.  I think the instructions explicitly say NO sudo, but I tried it after it failed the first time.  I can try to dig into the installation commands later, but at first glance may I recommend including a timeout for your *execute_shell_command* calls?  That should prevent such commands from hanging like this.

dorpvom:
A timeout would indeed be help. Especially in this case, where it would offer the information on which command it fails and why. The `execute_shell_command` function does support it. Another item for the to do list (which might be done rather shortly).

jahatfi:
I just made those changes and will submit a PR shortly to save you the trouble.

dorpvom:
[...]

PR #479, opened 2020-08-26
Added timeouts to prevent hanging

jahatfi:
Hi folks, I noticed my installation was hanging so I added 10 second timeouts to each `execute_shell` command to force it to fail rather than hang-see #477. Unless you'd prefer a different/modifiable hangout time?


jstucke (commented):
It seems you accidentally added the jstree files to the pull request. A quick `git rm -r --cached src/web_interface/static/3.3.9 src/web_interface/static/vakata-jstree-a7f2242` should fix that.

jahatfi:
All, I apologize, my initial PR had a bug in it I had to scroll way down to see.  An issue with my find/replace implementation.  I will try to submit a correct one within a week.  Thanks for your patience!

PR #480, opened 2020-08-27
added unique count stats

jstucke:
Added stats for unique count of device classes, vendors and device names

### DIFF
--- a/src/web_interface/templates/show_statistic.html
+++ b/src/web_interface/templates/show_statistic.html
@@ -69,9 +69,9 @@
                     {{ macros.stats_table_row("unique included files average size", stats["general_stats"]["average_file_size"] | number_format) }}
                 {% endif %}
                 {% if "device_class_count" in stats["general_stats"] %}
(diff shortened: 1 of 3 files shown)


src/web_interface/templates/show_statistic.html
```diff
@@ -68,6 +68,11 @@ <h3 class="mb-3">Firmware Statistics</h3>
                     {{ macros.stats_table_row("unique included files total size", stats["general_stats"]["total_file_size"] | number_format) }}
                     {{ macros.stats_table_row("unique included files average size", stats["general_stats"]["average_file_size"] | number_format) }}
                 {% endif %}
+                {% if "device_class_count" in stats["general_stats"] %}
```
dorpvom, line 71: The _unique_ before each stat is unnecessary. If they were not unique, then all values would just be == firmware samples

dorpvom (commented):
Code is okay. I'm still unsure, if I endorse to render this to the WebUI permanently.
We'll discuss it.

src/web_interface/templates/show_statistic.html
jstucke: changed

PR #481, opened 2020-08-27
Added 10s timeouts in the installation scripts to prevent hanging

### LINKED ISSUE #477: How to repair/uninstall failed installation?

jahatfi:
Second attempt to fix #477.   Only added timeouts to 5 installation files.  Timeouts are hardcoded at 10s at this time.  This time it throws an error while trying to `Install static jstree content` instead of hanging:
timeout
As a result I can see that the problem was caused by a permission error, which is much more helpful than getting stuck. Still haven't found the solution, but I have better clues now.

### DIFF
--- a/src/install/frontend.py
+++ b/src/install/frontend.py
@@ -17,7 +17,7 @@ COMPOSE_VENV = Path(__file__).parent.absolute() / 'compose-env'
 def execute_commands_and_raise_on_return_code(commands, error=None):  # pylint: disable=invalid-name
     for command in commands:
         bad_return = error if error else 'execute {}'.format(command)
-        output, return_code = execute_shell_command_get_return_code(command)
+        output, return_code = execute_shell_command_get_return_code(command, timeout=10)
         if return_code != 0:
             raise InstallationError('Failed to {}\n{}'.format(bad_return, output))
 
@@ -25,11 +25,11 @@ def execute_commands_and_raise_on_return_code(commands, error=None):  # pylint: 
 def wget_static_web_content(url, target_folder, additional_actions, resource_logging_name=None):
     logging.info('Install static {} content'.format(resource_logging_name if resource_logging_name else url))
     with OperateInDirectory(target_folder):
-        wget_output, wget_code = execute_shell_command_get_return_code('wget -nc {}'.format(url))
+        wget_output, wget_code = execute_shell_command_get_return_code('wget -nc {}'.format(url), timeout=10)
         if wget_code != 0:
             raise InstallationError('Failed to fetch resource at {}\n{}'.format(url, wget_output))
         for action in additional_actions:
-            action_output, action_code = execute_shell_command_get_return_code(action)
+            action_output, action_code = execute_shell_command_get_return_code(action, timeout=10)
             if action_code != 0:
                 raise InstallationError('Problem in processing resource at {}\n{}'.format(url, action_output))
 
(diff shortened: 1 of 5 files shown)


src/install/frontend.py
```diff
@@ -17,19 +17,19 @@
 def execute_commands_and_raise_on_return_code(commands, error=None):  # pylint: disable=invalid-name
     for command in commands:
         bad_return = error if error else 'execute {}'.format(command)
-        output, return_code = execute_shell_command_get_return_code(command)
+        output, return_code = execute_shell_command_get_return_code(command, timeout=10)
```
jstucke, line 20: It would probably better to make the timeout a parameter of the function `execute_commands_and_raise_on_return_code` so it can be adjusted as needed. E.g. the call to this function in `_build_highlight_js` should have a higher timeout as something gets downloaded  (maybe 30 seconds or 60) and for the other calls 10 seconds is probably enough.

src/install/frontend.py
```diff
@@ -17,19 +17,19 @@
...
         if return_code != 0:
             raise InstallationError('Failed to {}\n{}'.format(bad_return, output))
 
 
 def wget_static_web_content(url, target_folder, additional_actions, resource_logging_name=None):
     logging.info('Install static {} content'.format(resource_logging_name if resource_logging_name else url))
     with OperateInDirectory(target_folder):
-        wget_output, wget_code = execute_shell_command_get_return_code('wget -nc {}'.format(url))
+        wget_output, wget_code = execute_shell_command_get_return_code('wget -nc {}'.format(url), timeout=10)
```
jstucke, line 28: As again stuff gets downloaded here, the timeout should be less restrictive here. 30 seconds should be enough, though as the downloaded are not that big.

src/install/frontend.py
```diff
@@ -175,19 +175,19 @@ def main(radare, nginx):
     if radare:
         logging.info('Initializing docker container for radare')
 
-        execute_shell_command_get_return_code('virtualenv {}'.format(COMPOSE_VENV))
-        output, return_code = execute_shell_command_get_return_code('{} install -U docker-compose'.format(COMPOSE_VENV / 'bin' / 'pip'))
+        execute_shell_command_get_return_code('virtualenv {}'.format(COMPOSE_VENV), timeout=10)
+        output, return_code = execute_shell_command_get_return_code('{} install -U docker-compose'.format(COMPOSE_VENV / 'bin' / 'pip'), timeout=10)
```
jstucke, line 179: 30 seconds should be safe here

src/install/frontend.py
```diff
@@ -175,19 +175,19 @@ def main(radare, nginx):
...
         with OperateInDirectory('radare'):
-            output, return_code = execute_shell_command_get_return_code('{} build'.format(COMPOSE_VENV / 'bin' / 'docker-compose'))
+            output, return_code = execute_shell_command_get_return_code('{} build'.format(COMPOSE_VENV / 'bin' / 'docker-compose'), timeout=10)
             if return_code != 0:
                 raise InstallationError('Failed to initialize radare container:\n{}'.format(output))
 
     # pull pdf report container
     logging.info('Pulling pdf report container')
-    output, return_code = execute_shell_command_get_return_code('docker pull fkiecad/fact_pdf_report')
+    output, return_code = execute_shell_command_get_return_code('docker pull fkiecad/fact_pdf_report', timeout=10)
```
jstucke, line 190: This again loads a large docker container so I would omit the timeout.

jstucke (changes requested):
Thank you for your contribution. The timeouts must be carefully balanced to not prevent users from installing.

jahatfi:
@jstucke Thanks for the great feedback!  I apologize, I'm clearly making more work for you.  I will go through one last time and make the changes with your feedback.

PR #486, opened 2020-09-03
Sd1240 plugin doc js

jstucke:
- added docs for `helperFunctions.process`
- removed `nop` function (intercom)
- removed `timeout()` (only used in its own test)
- refactoring

### DIFF
--- a/src/helperFunctions/process.py
+++ b/src/helperFunctions/process.py
@@ -13,12 +13,6 @@ import psutil
 from helperFunctions.logging import TerminalColors, color_string
 
 
-def no_operation(*_):
-    '''
-    No Operation
-    '''
-
-
 def complete_shutdown(message=None):
     if message is not None:
         logging.error(message)
--- a/src/helperFunctions/process.py
+++ b/src/helperFunctions/process.py
@@ -1,47 +1,53 @@
-import functools
 import logging
 import os
-import signal
 import traceback
 from configparser import ConfigParser
 from contextlib import suppress
-from multiprocessing import Pipe, Process, pool
-from typing import Callable, List, Optional
+from multiprocessing import Pipe, Process
+from signal import SIGKILL, SIGTERM
+from typing import Callable, List, Optional, Tuple
 
 import psutil
 
 from helperFunctions.logging import TerminalColors, color_string
 
 
-def complete_shutdown(message=None):
+def complete_shutdown(message: Optional[str] = None) -> None:
+    '''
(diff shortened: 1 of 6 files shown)


src/helperFunctions/process.py
```diff
@@ -88,23 +108,43 @@ def start_single_worker(process_index, label: str, function: Callable) -> Except
...
+    :return: ``True`` if an exception occurred in any process and ``throw_exceptions`` in the FACT configuration is
+             set to `true` and ``False`` otherwise.
+    '''
     return_value = False
     for worker_process in process_list:
         if worker_process.exception:
             logging.error(color_string('Exception in {} process'.format(worker_label), TerminalColors.FAIL))
-            logging.error(worker_process.exception[1])
-            terminate_process_and_childs(worker_process)
+            _, stack_trace = worker_process.exception
```
dorpvom, line 129: Not new code, I know, but would it not make sense to include the stack trace directly in the first error. Then it's also highlighted red, what I would say makes sense in this case.

src/helperFunctions/process.py
```diff
@@ -1,53 +1,53 @@
...
-    No Operation
-    '''
-
+    Shutdown all FACT processes (of the currently running component) by sending a signal to the process group.
 
-def complete_shutdown(message=None):
+    :param message: Optional message to be displayed before the shutdown.
+    '''
     if message is not None:
         logging.error(message)
```
dorpvom, line 22: I'd say this is a warning. But not something we have to urgently change

src/helperFunctions/process.py
```diff
@@ -88,23 +108,43 @@ def start_single_worker(process_index, label: str, function: Callable) -> Except
...
     for worker_process in process_list:
         if worker_process.exception:
             logging.error(color_string('Exception in {} process'.format(worker_label), TerminalColors.FAIL))
-            logging.error(worker_process.exception[1])
-            terminate_process_and_childs(worker_process)
+            _, stack_trace = worker_process.exception
+            logging.error(stack_trace)
+            terminate_process_and_children(worker_process)
             process_list.remove(worker_process)
             if config is None or config.getboolean('ExpertSettings', 'throw_exceptions'):
```
dorpvom, line 133: Reading your apt description of this behavior I'm asking myself if the decision to test this here is correct. Nothing for this PR anyway.

dorpvom (approved):
Decide yourself if the requested changes are worth an iteration

dorpvom (approved):

PR #488, opened 2020-09-07
Add route decorator

jstucke:
- added decorator to imitate the Flask `@app.route()` decorator
- used the decorator on all analysis routes
- refactoring + test fixes


src/web_interface/components/component_base.py
```diff
@@ -1,16 +1,41 @@
...
         self._init_component()
 
-    @abstractmethod
     def _init_component(self):
-        pass
+        for attribute in dir(self):
+            method = getattr(self, attribute)
+            if _is_view_function(method):
+                for route in getattr(method, ROUTES_ATTRIBUTE):  # type: Route
+                    self._app.add_url_rule(rule=route.rule, view_func=method, methods=route.methods)
```
dorpvom, line 37: Does this automagically set the `view_func.__name__` as target that can be used by `url_for()`?

src/web_interface/components/analysis_routes.py
```diff
@@ -165,5 +168,8 @@ def _schedule_re_analysis_task(self, uid, analysis_task, re_do):
             sc.add_re_analyze_task(fw)
 
     @roles_accepted(*PRIVILEGES['delete'])
+    @AppRoute('/admin/re-do_analysis/<uid>', GET, POST)
     def _re_do_analysis(self, uid):
-        return self._update_analysis(uid, re_do=True)
+        if request.method == POST:
```
dorpvom, line 173: I though such distinction could now be omitted by directly labeling the functions with correct HTTP method?

src/web_interface/components/analysis_routes.py
```diff
@@ -90,7 +84,12 @@ def _get_correct_template(self, selected_analysis: str, fw_object: Union[Firmwar
...
         return get_template_as_string('show_analysis.html')
 
-    def _start_single_file_analysis(self, uid):
+    @roles_accepted(*PRIVILEGES['submit_analysis'])
+    @AppRoute('/analysis/<uid>', POST)
+    @AppRoute('/analysis/<uid>/ro/<root_uid>', POST)
+    @AppRoute('/analysis/<uid>/<selected_analysis>', POST)
+    @AppRoute('/analysis/<uid>/<selected_analysis>/ro/<root_uid>', POST)
+    def _start_single_file_analysis(self, uid, selected_analysis=None, root_uid=None):
         if user_has_privilege(current_user, privilege='submit_analysis'):
```
dorpvom, line 93: This should not be necessary anymore, does it?

dorpvom (changes requested):
I think I understand what happens. Anyway, some minor changes requested and questions asked.

src/web_interface/components/component_base.py
```diff
@@ -1,16 +1,41 @@
...
 
+GET = 'GET'
+POST = 'POST'
 
-class ComponentBase(metaclass=ABCMeta):
+
+Route = NamedTuple('Route', [('rule', str), ('methods', Tuple[str, ...])])
+
+
+class AppRoute:
```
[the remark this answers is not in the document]
jstucke: added docstring

src/web_interface/components/component_base.py
jstucke: Yes, `url_for()` works with the endpoint name and if no custom endpoint name is used, the function name (i.e. `view_func.__name__`) is used (e.g. `url_for('show_analysis')`)

src/web_interface/components/analysis_routes.py
jstucke: Yes we could do that, but in this special case I found it to be more compact this way, because this is more or less only a redirect to the update analysis view functions. I could change it if you want, though.

src/web_interface/components/analysis_routes.py
jstucke: You are totally right, I didn't even notice that -> changed

src/web_interface/components/analysis_routes.py
dorpvom: I get what you are saying. I'm looking at it right now and don't really feel one is much better than the other. So this can stay as it is.

dorpvom (approved):

PR #496, opened 2020-09-22
IP and URI filter plugin

any333:


src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+        blacklisted = self.blacklist_ip_and_uris(BLACKLIST, uris_dict)
+        whitelisted = self.whitelist_ip_and_uris(WHITELIST, blacklisted)
+        file_object.processed_analysis[self.NAME] = {"whitelisted": whitelisted, "summary": whitelisted}
+        return file_object
+
+    @staticmethod
+    def blacklist_ip_and_uris(deny_list: list, ip_and_uri_list: list) -> list:
+        for ip_uri in ip_and_uri_list:
+            for entry in deny_list:
+                if entry in ip_uri.lower():
```
jstucke, line 40: This is pretty much exactly what the helper function `substring_is_in_list` does, so we should use it here:

```suggestion
            if substring_is_in_list(ip_uri.lower(), deny_list):
```

This also fixes an error when multiple blacklist elements occur in the element

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+                    ip_and_uri_list.remove(ip_uri)
+        return ip_and_uri_list
+
+    @staticmethod
+    def whitelist_ip_and_uris(allow_list: list, ip_and_uri_list: list) -> list:
+        clean_api_and_uri_list = []
+        for ip_uri in ip_and_uri_list:
+            for entry in allow_list:
+                if entry in ip_uri.lower():
+                    if ip_uri not in clean_api_and_uri_list:
```
jstucke, line 50: `substring_is_in_list` can be used here, too:

```suggestion
            if substring_is_in_list(ip_uri.lower(), allow_list):
```

This should also make the check for duplicates redundant

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+                        clean_api_and_uri_list.append(ip_uri)
+        return clean_api_and_uri_list
+
+    @staticmethod
+    def valid_ip_address(ip: str) -> str:
+        try:
+            return "IPv4" if type(ip_address(ip)) is IPv4Address else "IPv6"
+        except ValueError:
+            return "0"
```
jstucke, line 59: Since the type of the ip address doesn't matter in the function below, we can just check if it is valid:
```suggestion
    def is_valid_ip_address(ip: str) -> bool:
        try:
            ip_address(ip)
            return True
        except ValueError:
            return False
```

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def remove_ip_v4_v6_addresses(self, ip_and_uri_list: list) -> list:
+        for ip_uri in ip_and_uri_list:
+            if self.valid_ip_address(ip_uri) == 'IPv4' or self.valid_ip_address(ip_uri) == 'IPv6':
```
jstucke, line 63:
```suggestion
            if self.is_valid_ip_address(ip_uri):
```

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @staticmethod
+    def valid_ip_address(ip: str) -> str:
+        try:
+            return "IPv4" if type(ip_address(ip)) is IPv4Address else "IPv6"
+        except ValueError:
+            return "0"
+
+    def remove_ip_v4_v6_addresses(self, ip_and_uri_list: list) -> list:
+        for ip_uri in ip_and_uri_list:
```
jstucke, line 62: Since the list is changed during iteration (with `.remove()`) we should iterate over a copy instead of the actual list in order to prevent errors
```suggestion
        for ip_uri in ip_and_uri_list[:]:
```

jstucke (changes requested):
This still needs some tests

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
```diff
@@ -0,0 +1,65 @@
...
+    def process_object(self, file_object):
+        list_of_ips_and_uris = file_object.processed_analysis['ip_and_uri_finder']['summary']
+        uris_dict = self.remove_ip_v4_v6_addresses(list_of_ips_and_uris)
+        blacklisted = self.blacklist_ip_and_uris(BLACKLIST, uris_dict)
+        whitelisted = self.whitelist_ip_and_uris(WHITELIST, blacklisted)
+        file_object.processed_analysis[self.NAME] = {"whitelisted": whitelisted, "summary": whitelisted}
+        return file_object
+
+    @staticmethod
+    def blacklist_ip_and_uris(deny_list: list, ip_and_uri_list: list) -> list:
```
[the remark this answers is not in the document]
jstucke: The same goes for `allow_list` below

src/plugins/analysis/ip_and_uri_finder_filter/code/ip_and_uri_filter.py
jstucke: This could also be a problem in line 38

src/plugins/analysis/interesting_uris/code/interesting_uris.py
```diff
@@ -0,0 +1,65 @@
+from ipaddress import ip_address, IPv4Address
```
jstucke, line 1: import `IPv4Address` is no longer used

jstucke (commented):
Has an unused import which makes the CI fail. Looks fine otherwise

jstucke (approved):

jstucke (approved):

PR #505, opened 2020-10-26
fix insufficient info bug

jstucke:
should fix the insufficient information bug when clicking on an element in the file tree of a file object comparison

### DIFF
--- a/src/objects/file.py
+++ b/src/objects/file.py
@@ -219,15 +219,11 @@ class FileObject:  # pylint: disable=too-many-instance-attributes
         :param root_uid: (Optional) root uid to get vfps for.
         :return: List of virtual paths.
         '''
-        try:
-            file_paths = self.get_virtual_file_paths()
-            req_root_uid = root_uid or self.root_uid
-            if req_root_uid is None:
-                return get_value_of_first_key(file_paths)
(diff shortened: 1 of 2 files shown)


src/objects/file.py
```diff
@@ -219,15 +219,11 @@ def get_virtual_paths_for_one_uid(self, root_uid: str = None) -> List[str]:
...
-            if req_root_uid is None:
-                return get_value_of_first_key(file_paths)
+        file_paths = self.get_virtual_file_paths()
+        req_root_uid = root_uid or self.root_uid
+        if req_root_uid in file_paths:
             return file_paths[req_root_uid]
-        except (AttributeError, IndexError, KeyError, TypeError):
-            logging.error('Error on virtual file path retrieval. This should be fixed')
-            return ["insufficient information: firmware analysis not complete"]
+        return get_value_of_first_key(file_paths)  # fallback
```
dorpvom, line 226: Hm. Do we still expect this to happen from time to time? If so we should fail here in some sensible way. I know that the previous way is misleading, but we might be able to create a flash message or fail in another way that is not "the frontend breaking".

dorpvom (approved):

PR #508, opened 2020-10-27
Linter linguist

any333:


src/plugins/analysis/linter/code/source_code_analysis.py
```diff
@@ -37,37 +40,37 @@ class AnalysisPlugin(AnalysisBasePlugin):
...
 
-            first_line = file_object.binary.decode().splitlines(keepends=False)[0]
-            if first_line.find(self.SCRIPT_TYPES[script_type]['shebang']) >= 0:
+    def _get_script_type(self, file_object, linguist_output):
+        if 'language' in linguist_output:
+            script_language = linguist_output.split('language:', 1)[1]
+            script_type = script_language.translate({ord(c): None for c in string.whitespace}).lower()
+            if script_type:
+                file_object.processed_analysis['file_type']['linguist'] = script_type
                 return script_type
```
jstucke, line 58: This will not work correctly if there is additional linguist output after the "language" line (e.g. "appears to be generated source code")
In my opinion it would be better to store the full linguist output in the file_type analysis (because there could be more interesting information in it for the user).
Also this seems unnecessary complicated and could be easier parsed with re.search, like e.g.
```suggestion
            file_object.processed_analysis['file_type']['linguist'] = linguist_output
            match = re.search(r'language:\s*(\w+)', linguist_output)
            if match:
                return match.groups()[0].lower()
```

src/plugins/analysis/linter/test/test_source_code_analysis.py
```diff
@@ -32,45 +32,22 @@ def stub_plugin(test_config, monkeypatch):
...
-
-    assert stub_plugin._determine_script_type(test_object) == script_type
-
-
 @pytest.mark.parametrize('shebang_and_type', [
-    (b'#!/usr/bin/env python3', 'python'),
-    (b'#!/usr/bin/python', 'python'),
-    (b'#!/bin/bash', 'shell'),
-    (b'#!/bin/sh', 'shell')
+    (b'#!/usr/bin/env python3', b'testfile1:4 lines(4 sloc)\r\n  type:Text\r\n  mime type:text/plain\r\n  language:Python\r\n', 'python'),
```
jstucke, line 36: where did all those carriage returns ("\r") come from? I couldn't find any when I tested it :thinking:

jstucke:
Also there seems to be a PEP8 issue on the CI:
```
22:38:32 /home/jenkins/workspace/bionic/src/plugins/analysis/linter/code/source_code_analysis.py:65:67: W291 trailing whitespace
```

jstucke (approved):

dorpvom (approved):
Checked plugin functionality and found it to be quite useful

PR #510, opened 2020-11-06
1299: Correctly render YARA search queries

dorpvom:


src/web_interface/filter.py
```diff
@@ -400,3 +401,11 @@ def elapsed_time(start_time: float) -> int:
 
 def format_duration(duration: float) -> str:
     return str(timedelta(seconds=duration))
+
+
+def render_query_title(query_title: str):
```
jstucke, line 406: Is `query_title` really of type `str` when there are checks whether it is `None` or `dict`?

jstucke (commented):
Seems to work well. Should there be a test for the new filter?

src/web_interface/filter.py
dorpvom: Good point. It's either None, a string or a dict

dorpvom: Though no idea if the None case actually exists, but it was handled before.

jstucke (approved):

PR #511, opened 2020-11-06
1239: Added docstrings to install helper and moved crypto helper

dorpvom:


src/helperFunctions/install.py
```diff
@@ -3,17 +3,26 @@
...
+    '''
+    Class representing all expected errors that happen during installation, such as timeouts on remote hosts.
+    '''
 
 
 class OperateInDirectory:
-    def __init__(self, target_directory: Union[str, Path], remove=False):
+    '''
+    Context manager allowing to execute a number of commands in a given directory. On exit, the working directory is
+    restored to its previous state.
```
jstucke, line 20: IMHO this could be misunderstood as "changes inside the directory are reverted" when you actually mean that the current working directory is reset to the previous directory.

src/helperFunctions/install.py
```diff
@@ -53,89 +74,135 @@ def run_shell_command_raise_on_return_code(command: str, error: str, add_output_
...
 
 def dnf_update_sources():
-    return run_shell_command_raise_on_return_code('sudo dnf update -y', 'Unable to update')
+    '''
+    Update package lists on Fedora / RedHat / Cent systems.
+    '''
+    return _run_shell_command_raise_on_return_code('sudo dnf update -y', 'Unable to update')
+
 
+def dnf_install_packages(*args: str):
```
jstucke, line 83: I don't think `args` is the best name for this variable. Maybe "packages"? (applies to other functions as well)

src/helperFunctions/install.py
```diff
@@ -160,7 +227,12 @@ def _checkout_github_project(github_path, folder_name):
         raise InstallationError('Repository creation failed on folder {}\n {}'.format(folder_name, clone_url))
 
 
-def load_main_config():
+def load_main_config() -> configparser.ConfigParser:
```
jstucke, line 230: maybe this isn't the right place to discuss this but why isn't this function in helperFunctions/config.py or why isn't `load_config()` from that module used? Seems a bit redundant.

jstucke (commented):
I think moving `parser.py` was a good idea.

src/helperFunctions/install.py
dorpvom: I had this in mind already when writing this. I guess it's back to the drawing board...

src/helperFunctions/install.py
dorpvom: Keep in mind that we are still setting up python dependencies at this point

dorpvom: Indeed, the config helper depends on the process helper which itself depends on _psutil_. _psutil_ though is not present at link time of the installation on a clean system.

jstucke: that is unfortunate indeed

jstucke (approved):
looks fine

PR #513, opened 2020-11-09
Orphaned objects

jstucke:
- added "orphaned objects" to the missing analyses page, showing file objects whose parent firmware is missing from the database

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -277,6 +277,24 @@ class FrontEndDbInterface(MongoInterfaceCommon):
                 parent_to_included.pop(parent_uid)
         return parent_to_included
 
+    def find_orphaned_objects(self):
+        ''' find File Objects whose parent firmware is missing '''
+        query_result = self.file_objects.aggregate([
+            {'$project': {'parent_firmware_uids': 1}},
+            {'$unwind': '$parent_firmware_uids'},
+            {'$lookup': {
+                'from': 'firmwares',
+                'localField': 'parent_firmware_uids',
+                'foreignField': '_id',
+                'as': 'firmware'
+            }},
+            {'$match': {'firmware': {'$size': 0}}}
+        ], allowDiskUse=True)
+        orphans_by_parent = {}
+        for entry in query_result:
+            orphans_by_parent.setdefault(entry['parent_firmware_uids'], []).append(entry['_id'])
+        return orphans_by_parent
+
     def find_missing_analyses(self):
         missing_analyses = {}
         query_result = self.firmwares.aggregate([
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -279,21 +279,25 @@ class FrontEndDbInterface(MongoInterfaceCommon):
 
     def find_orphaned_objects(self):
         ''' find File Objects whose parent firmware is missing '''
-        query_result = self.file_objects.aggregate([
-            {'$project': {'parent_firmware_uids': 1}},
-            {'$unwind': '$parent_firmware_uids'},
-            {'$lookup': {
-                'from': 'firmwares',
-                'localField': 'parent_firmware_uids',
-                'foreignField': '_id',
-                'as': 'firmware'
-            }},
-            {'$match': {'firmware': {'$size': 0}}}
-        ], allowDiskUse=True)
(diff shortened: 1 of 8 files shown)


src/web_interface/components/miscellaneous_routes.py
```diff
@@ -78,6 +78,7 @@ def _app_delete_firmware(self, uid):
     def _app_find_missing_analyses(self):
         template_data = {
             'missing_files': self._find_missing_files(),
+            'orphaned_files': self._find_orphaned_files(),
```
dorpvom, line 81: This (and also `failed_analyses`) is missing in the corresponding REST API endpoint.

src/storage/db_interface_frontend.py
```diff
@@ -277,6 +277,28 @@ def find_missing_files(self):
                 parent_to_included.pop(parent_uid)
         return parent_to_included
 
+    def find_orphaned_objects(self):
+        ''' find File Objects whose parent firmware is missing '''
+        orphans_by_parent = {}
+        query_result = list(self.file_objects.aggregate([
```
dorpvom, line 283: `query_result` does not tell me anything about it's content. Since no one - aside from you, dear Sir - understands aggregate queries, you should rename the variable.

src/storage/db_interface_frontend.py
```diff
@@ -277,6 +277,28 @@ def find_missing_files(self):
...
+            if missing_uids:
+                query_result = self.file_objects.find({'parent_firmware_uids': {'$in': list(missing_uids)}})
+                for fo_entry in query_result:
+                    for uid in missing_uids:
+                        if uid in fo_entry['parent_firmware_uids']:
+                            orphans_by_parent.setdefault(uid, []).append(fo_entry['_id'])
+        return orphans_by_parent
+
+    def _get_all_firmware_uids(self) -> List[str]:
+        query_result = list(self.firmwares.aggregate([{'$group': {'_id': 0, 'firmware_uids': {'$push': '$_id'}}}]))
```
dorpvom, line 299: Why use aggregate here? Does it bring a speed up over sth. like 
```python3
    def _get_all_firmware_uids(self) -> List[str]:
        return [
            firmware['_id'] for firmware in self.firmwares.find({}, {' _id': 1})
        ]
```

src/storage/db_interface_frontend.py
```diff
@@ -277,6 +277,28 @@ def find_missing_files(self):
...
+        orphans_by_parent = {}
+        query_result = list(self.file_objects.aggregate([
+            {'$unwind': '$parent_firmware_uids'},
+            {'$group': {'_id': 0, 'all_parent_uids': {'$addToSet': '$parent_firmware_uids'}}}
+        ], allowDiskUse=True))
+        if query_result:
+            fo_parent_firmware = set(query_result[0]['all_parent_uids'])
+            missing_uids = fo_parent_firmware.difference(self._get_all_firmware_uids())
+            if missing_uids:
+                query_result = self.file_objects.find({'parent_firmware_uids': {'$in': list(missing_uids)}})
```
dorpvom, line 291: see above

dorpvom (changes requested):

src/web_interface/components/miscellaneous_routes.py
jstucke: added both to REST and added tests

src/storage/db_interface_frontend.py
jstucke: renamed to `fo_parent_uids`

src/storage/db_interface_frontend.py
jstucke: I was unsure about the speed and efficiency, so I tried both with these results:

> find: 0.001560211181640625 seconds
> aggregate: 0.0003306865692138672 seconds

So the aggregate seems actually to be faster (about 4.7x) (probably because the iteration is done in MongoDB and not in Python).

src/storage/db_interface_frontend.py
jstucke: inlined

src/storage/db_interface_frontend.py
dorpvom: Fancy shit. So I have to accept this.

dorpvom (approved):

PR #515, opened 2020-11-11
no more sanitize db duplicate entries

jstucke:
- duplicate entries in the "sanitize_db" were created each time `sanitize_analysis()` was called
- fix: only create entry if there is none or if it has changed and delete old ones

### DIFF
--- a/src/storage/db_interface_common.py
+++ b/src/storage/db_interface_common.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import pickle
+from hashlib import md5
 from typing import List, Set
 
 import gridfs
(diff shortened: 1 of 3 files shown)


src/storage/db_interface_common.py
```diff
@@ -178,12 +179,22 @@ def _extract_binaries(self, analysis_dict, key, uid):
...
             else:
                 tmp_dict[analysis_key] = analysis_dict[key][analysis_key]
         return tmp_dict
 
+    def _store_in_sanitize_db(self, content: bytes, file_name: str):
+        if self.sanitize_fs.exists({'filename': file_name}):
+            md5_hash = md5(content).hexdigest()
+            if self.sanitize_fs.exists({'md5': md5_hash, 'filename': file_name}):
+                return  # there is already an up to date entry -> do nothing
+            for old_entry in self.sanitize_fs.find({'filename': file_name}):  # delete old entries first
```
dorpvom, line 193: Is there a query filter for GridFS? Like can you only request the `_id` such as is possible with regular MongoDB queries.

dorpvom (changes requested):

src/storage/db_interface_common.py
```diff
@@ -178,12 +179,22 @@ def _extract_binaries(self, analysis_dict, key, uid):
...
+                self._store_in_sanitize_db(pickle.dumps(analysis_dict[key][analysis_key]), file_name)
                 tmp_dict[analysis_key] = file_name
             else:
                 tmp_dict[analysis_key] = analysis_dict[key][analysis_key]
         return tmp_dict
 
+    def _store_in_sanitize_db(self, content: bytes, file_name: str):
+        if self.sanitize_fs.exists({'filename': file_name}):
+            md5_hash = md5(content).hexdigest()
+            if self.sanitize_fs.exists({'md5': md5_hash, 'filename': file_name}):
```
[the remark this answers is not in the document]
jstucke: Both queries take about 0.001 seconds according to Robo 3T so it probably doesn't make much of a difference. You are totally right that md5 should suffice, though.

src/storage/db_interface_common.py
jstucke: It seems like there is no "projection" parameter for the `find()` method in GridFS.

dorpvom (approved):

PR #518, opened 2020-11-23
added elf executable stats

jstucke:
- added ELF executable stats


src/web_interface/static/js/fact_statistics.js
```diff
@@ -70,7 +70,19 @@ function set_links(canvas_id, any_chart, link) {
...
 }
 
-function create_horizontal_bar_chart(canvas_id, chart_data, link, value_percentage_present_flag) {
+function set_links_from_data(canvas_id, chart) {
+
+    document.getElementById(canvas_id).onclick = function(evt){
+        var points = chart.getElementsAtEvent(evt);
+        if (chart.data.datasets[0].links !== undefined) {
+            var key = chart.data.datasets[0].links[points[0]._index];
+            window.location = link.replace("PLACEHOLDER", key);
```
dorpvom, line 79: Where is `link` set? Is there some implicit inheritance?

src/test/integration/statistic/test_update.py
```diff
@@ -185,3 +189,36 @@ def test_fetch_mitigations(self):
...
+            fo.uid = str(i)
+            self.db_backend_interface.add_file_object(fo)
+
+        stats = self.updater.get_executable_stats().get('executable_stats')
+        expected = [
+            ('big endian', 1, 0.25), ('little endian', 3, 0.75), ('stripped', 1, 0.25), ('not stripped', 2, 0.5),
+            ('32-bit', 1, 0.25), ('64-bit', 3, 0.75), ('dynamically linked', 2, 0.5), ('statically linked', 1, 0.25),
+            ('missing section info', 1, 0.25)
+        ]
+        for expected_item, stat_item in zip(expected, stats):
```
dorpvom, line 223: This construction, I don't like. No for real: Can you not do sth. like 
```python
for label, count, percentage in expected:
    _, db_count, db_percentage, __ = stats[label]
    assert count == db_count
    assert percentage == db_percentage
```
I get that that's less compact compared to your solution, but it might be more readable.

dorpvom (changes requested):

src/web_interface/static/js/fact_statistics.js
jstucke: The links are passed to `BarChart` in `create_horizontal_bar_chart()` as part of (chart_)data (which in turn comes from the filter `data_to_chart_with_value_percentage_pairs`)

dorpvom (approved):

PR #520, opened 2020-11-26
added yara_binary_search docs

jstucke:


src/helperFunctions/yara_binary_search.py
```diff
@@ -75,25 +85,37 @@ def get_binary_search_result(self, task: Tuple[bytes, Optional[str]]):
...
+    '''
+    Get the exception that is caused by trying to compile ``rules_file`` with yara or ``None`` if there is none.
 
-def get_yara_error(rules_file):
+    :param rules_file: A string containing yara rules.
+    :result: The exception if compiling the rules causes an exception or ``None`` otherwise.
+    '''
     if isinstance(rules_file, bytes):
         rules_file = rules_file.decode()
     try:
```
dorpvom, line 121: Is there a specific reason why you haven't documented `class YaraBinarySearchScannerDbInterface` ?

src/helperFunctions/yara_binary_search.py
```diff
@@ -12,23 +13,30 @@
 
 
 class YaraBinarySearchScanner:
+    '''
+    A scanner for searching with YARA rules on the binaries belonging to the firmware and file objects contained
```
dorpvom, line 17: It think we can improve the wording here. How 'bout:

This class provides functionality to scan files in the database for [yara](http://virustotal.github.io/yara/) patterns. The public method allows to either match a given set of patterns on all files in the database or focus only on files included in a single firmware.

dorpvom (changes requested):

src/helperFunctions/yara_binary_search.py
jstucke: Yes: it is only used internally by the other functions (I could also have renamed it to `_YaraBinarySearchScannerDbInterface` but that looked ugly)

src/helperFunctions/yara_binary_search.py
jstucke: I agree → changed

src/helperFunctions/yara_binary_search.py
dorpvom: Haha. Yes it does

dorpvom (approved):

PR #524, opened 2020-12-08
Feedback modal

any333:


dorpvom:
Can you add a `mallto:` link to our mail address (firmware-security@fkie.fraunhofer.de)?

src/web_interface/templates/base.html
```diff
@@ -104,123 +108,179 @@
...
+                            class="fas fa-heartbeat"></i> System</a>
+                    <a class="dropdown-item" href="/about" style="font-size: 14px"><i class="fas fa-beer"></i> About</a>
+                </div>
+            </li>
+            {% elif caption == "Feedback" %}
+            <li class="pr-3">
+                <a {% if caption== active_page %}style="color: #fff; font-weight: 500;" {% endif %}
+                    data-toggle="modal" data-target="#staticBackdrop"> <i class="fas fa-bullhorn"></i> Feedback </a>
+            </li>
+            <div class="modal fade" id="staticBackdrop" data-backdrop="static" data-keyboard="false" tabindex="-1"
```
jstucke, line 171: we could probably find a better name for this element than "staticBackdrop". How about "feedbackModal"?
Also this should be outside the `<ul>` Tag (at least PyCharm is complaining that `<div>`s aren't allowed here)

jstucke (commented):
Looks nice and seems to work well.

jstucke (approved):

PR #529, opened 2021-01-18
dockerized fs metadata plugin

jstucke:
- moved file system metadata plugin mounting into docker container to get rid of sudo
- refactored os in tests to pathlib


src/plugins/analysis/file_system_metadata/docker/mount.py
```diff
@@ -0,0 +1,54 @@
...
+
+def _get_mounted_file_stats():
+    result = []
+    for file in MOUNT_DIR.rglob("*"):
+        try:
+            if not file.is_symlink() and file.is_file():
+                result.append((file.name, str(file.relative_to(MOUNT_DIR)), file.lstat()))
+        except PermissionError:
+            continue
+    return result
```
dorpvom, line 45: There does not appear to be a particular reason to use pickle here instead of json (other than us having to convert `Path.lstat` to a dict. We should try to avoid pickle as much as possible.

src/plugins/analysis/file_system_metadata/docker/mount.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        check_call(f'mount -o ro,loop {input_file} {MOUNT_DIR}', shell=True)
+        yield
+    finally:
+        check_call(f'umount {MOUNT_DIR}', shell=True)
+
+
+def main():
+    try:
+        stats = INPUT_FILE.lstat()
```
dorpvom, line 27: I've to admit. This is clever ; )

src/plugins/analysis/file_system_metadata/docker/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM fkiecad/fact_extractor:latest
```
dorpvom, line 1: I guess you use the extractor as it supposedly has additional filesystem support?

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -98,31 +75,45 @@ def _parent_has_file_system_metadata(self, file_object: FileObject) -> bool:
...
-            with mount(file_object.file_path, type_parameter, get_temp_dir_path(self.config)) as mounted_path:
-                self._analyze_metadata_of_mounted_dir(mounted_path)
+    def _extract_metadata_from_file_system(self, file_object: FileObject):
+        with TemporaryDirectory() as tmp_dir:
+            input_file = Path(tmp_dir) / "input.img"
+            input_file.write_bytes(file_object.binary or Path(file_object.file_path).read_bytes())
+            output = self._mount_in_docker(tmp_dir)
+            output_file = Path(tmp_dir) / "output.pickle"
+            if output_file.is_file():
+                self._analyze_metadata_of_mounted_dir(pickle.loads(output_file.read_bytes()))
```
dorpvom, line 92: See other comment regarding pickle.

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -98,31 +75,45 @@ def _parent_has_file_system_metadata(self, file_object: FileObject) -> bool:
...
-                self._enter_results_for_mounted_file(file_)
+    def _analyze_metadata_of_mounted_dir(self, docker_results: Tuple[str, str, os.stat_result]):
+        for file_name, file_path, file_stats in docker_results:
+            self._enter_results_for_mounted_file(file_name, file_path, file_stats)
 
-    def _enter_results_for_mounted_file(self, file_: Path):
-        result = self.result[b64encode(file_.name.encode()).decode()] = {}
-        stats = os.lstat(str(file_))
+    def _enter_results_for_mounted_file(self, file_name: str, file_path: str, stats: os.stat_result):
+        result = self.result[b64encode(file_name.encode()).decode()] = {}
```
dorpvom, line 112: What is the historical explanation behind this usage of base64?

src/plugins/analysis/file_system_metadata/test/test_plugin_file_system_metadata.py
```diff
@@ -52,33 +57,33 @@ def setUp(self):
         self.analysis_plugin = plugin.AnalysisPlugin(self, config=config)
         plugin.FsMetadataDbInterface.__bases__ = (DatabaseMock,)
         plugin.ConnectTo.__enter__ = mock_connect_to_enter
```
dorpvom, line 59: Why isn't this mocking done with a mock library? Overwriting parameters is generally bad style.

src/plugins/analysis/file_system_metadata/test/test_plugin_file_system_metadata.py
```diff
@@ -1,23 +1,28 @@
...
 
 from test.common_helper import TEST_FW, TEST_FW_2, DatabaseMock, create_test_file_object
+from test.mock import mock_patch
 from test.unit.analysis.analysis_plugin_test_class import AnalysisPluginTest
+
 from ..code import file_system_metadata as plugin
 from ..code.file_system_metadata import FsKeys
 
-TEST_DATA_DIR = os.path.join(get_dir_of_file(__file__), 'data')
+PLUGIN_NAME = 'file_system_metadata'
```
dorpvom, line 15: Can't you take this from the `NAME` attribute of the plugin?

dorpvom (changes requested):

src/plugins/analysis/file_system_metadata/docker/Dockerfile
jstucke: exactly. No point in setting up another container when the extractor already supports mounting all of them for extraction purposes.

src/plugins/analysis/file_system_metadata/docker/mount.py
jstucke: changed

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
jstucke: changed to json

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
jstucke: there was a problem with certain (special) characters in file names but I can't remember exactly what is was

src/plugins/analysis/file_system_metadata/test/test_plugin_file_system_metadata.py
jstucke: changed to `mock.patch`

src/plugins/analysis/file_system_metadata/test/test_plugin_file_system_metadata.py
jstucke: In general: yes. But it is used by the base class `AnalysisPluginTest` (which means we can't really get rid of it except if we change all plugin tests)

dorpvom (approved):

PR #531, opened 2021-01-19
changed Ghidra version to 9.2.1

jstucke:
- updated ghidra docker image
- updated `format_string_resolver` Dockerfile
- fixed bug in lief install


Enkelmann:
The build is still failing on xenial. Do we still officially support xenial? If yes, I would suggest installing v0.10.1 of `lief` instead of v0.11.0 (which was released just yesterday) or the Github master branch (could spell a lot of trouble in the future).

jstucke:
> Do we still officially support xenial?

We do support Xenial but not Python 3.5 (which reached its end of life). The problem is: the Xenial CI instance is running Python 3.5. The plan is to upgrade the Python version (but we will see if we get to access the physical machine running the CI under the current circumstances before Xenial hits its end of life in April).

Enkelmann (approved):
Looks good to me if the CI is also happy with the pinned lief version. :+1:

PR #533, opened 2021-01-25
Fedora33 support

jstucke:
- experimental support for Fedora 32 and 33
- updated docker installation


dorpvom (approved):
How was this tested? It looks sound but lacking CI builds for all three affected Fedora versions (31, 32, 33) I need some assurance that this works

jstucke:
I tested it when I needed to debug the Fedora build for the FS metadata docker PR where the tests were seemingly randomly failing on Fedora (and Debian but not Ubuntu). For that I needed a working installation for Fedora 33.

PR #536, opened 2021-01-29
made max chart elements configurable

jstucke:
- made the number of displayed elements in the donut / pie charts on the stats page configurable (through the main config)


src/web_interface/components/jinja_filter.py
```diff
@@ -134,15 +134,38 @@ def _split_user_and_password_type_entry(result: dict):
...
 
+    def data_to_chart_limited(self, data, limit: Optional[int] = None, color_list=None):
+        limit = self._config.getint('statistics', 'max_elements_per_chart', fallback=10) if limit is None else limit
+        try:
+            label_list, value_list = [list(d) for d in zip(*data)]
+        except ValueError:
+            return None
+        label_list, value_list = flt.set_limit_for_data_to_chart(label_list, limit, value_list)
+        color_list = flt.set_color_list_for_data_to_chart(color_list, value_list, limit)
+        result = {
```
dorpvom, line 145: `return {` ...

dorpvom:
I'm waiting until a working CI run for further review

src/web_interface/components/jinja_filter.py
```diff
@@ -134,15 +134,32 @@ def _split_user_and_password_type_entry(result: dict):
     def check_auth(self, _):
         return self._config.getboolean('ExpertSettings', 'authentication')
 
+    def data_to_chart_limited(self, data, limit: Optional[int] = None, color_list=None):
+        limit = self._config.getint('statistics', 'max_elements_per_chart', fallback=10) if limit is None else limit
```
dorpvom, line 138: handle above mention error here

src/web_interface/filter.py
```diff
@@ -256,9 +237,9 @@ def data_to_chart_with_value_percentage_pairs(data, limit=10):  # pylint: disabl
     return result
 
 
-def set_color_list_for_data_to_chart(color_list, value_list):
+def set_color_list_for_data_to_chart(color_list, value_list, limit: int):
```
dorpvom, line 240: Hm. Not sure I like this being here (since it's only use is in `jinja_filter.py`. But if we view this module as a de facto helper module it's fine I guess.

dorpvom (changes requested):
Minor changes including some exception handling requested.

src/helperFunctions/web_interface.py
```diff
@@ -40,7 +40,7 @@ def get_alternating_color_list(number: int, limit: int = 10) -> List[str]:
     :param limit: The maximum number of returned colors.
     :return: A list of alternating hex color values.
     '''
-    color_list = get_color_list(8)
+    color_list = get_color_list(8, limit=10)
```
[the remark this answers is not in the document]
jstucke: It seems redundant at first but the value of limit determines which colors `[color_list[0], color_list[7]]` (the actual alternating colors) are. If limit was very large, they could be indistinguishable. So to keep them constant, the value of limit must also be constant.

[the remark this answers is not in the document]
jstucke: > You wanna make sure no change in get_color_list breaks this?

Yes no future surprises :sweat_smile:

src/web_interface/filter.py
jstucke: The split between `filter.py` and `jinja_filter.py` seems rather arbitrary to me anyway. Maybe we find better ways to split them up.

jstucke: Anyway ... moved to jinja_filter.py

src/web_interface/components/jinja_filter.py
jstucke: changed as suggested

dorpvom (approved):

dorpvom (approved):

PR #537, opened 2021-02-05
Improved password regexes

jstucke:
- improved unix / `/etc/shadow` password regex to reduce false positive matches
- added htpasswd and DES regexes 
- refactoring
- added test cases


Enkelmann (changes requested):
Could you add a regression test for one of the false positives that should be prevented now? The rest looks fine to me.

Enkelmann (approved):
Thanks!

Issue #538, opened 2021-02-08
Install on Ubuntu 20.04.2 Not functional - Internal Server Error 


sperglord8008s:
GUI works and can move around the interface but Upload and tools dont work. 
Currently running Ubuntu in VM from Windows machine. 

~/FACT_core/start_all_installed_fact_components
[2021-02-07 16:45:09][start_all_installed_fact_components][INFO]: starting db
[2021-02-07 16:45:09][MongoMgr][INFO]: start local mongo database
[2021-02-07 16:45:11][start_all_installed_fact_components][INFO]: starting frontend
[2021-02-07 16:45:11][start_all_installed_fact_components][INFO]: starting backend

**[2021-02-07 16:45:11][plugin][WARNING]: Plugin has no code directory: /home/user/FACT_core/src/plugins/analysis/__pycache__
Traceback (most recent call last):**
  File "/home/user/FACT_core/src/../start_fact_backend", line 54, in <module>
    analysis_service = AnalysisScheduler(config=config)
  File "/home/user/FACT_core/src/scheduler/Analysis.py", line 37, in __init__
    self.load_plugins()
  File "/home/user/FACT_core/src/scheduler/Analysis.py", line 212, in load_plugins
    plugin = source.load_plugin(plugin_name)
  File "/usr/local/lib/python3.8/dist-packages/pluginbase.py", line 301, in load_plugin
    return __import__(self.base.package + '.' + name,
  File "/usr/local/lib/python3.8/dist-packages/pluginbase.py", line 438, in plugin_import
    return self._system_import(import_name, globals, locals,
  File "/home/user/FACT_core/src/plugins/analysis/cve_lookup/code/cve_lookup.py", line 12, in <module>
    from pyxdameraulevenshtein import damerau_levenshtein_distance as distance  # pylint: disable=no-name-in-module
  File "/usr/local/lib/python3.8/dist-packages/pluginbase.py", line 438, in plugin_import
    return self._system_import(import_name, globals, locals,
  File "pyxdameraulevenshtein.pyx", line 1, in init pyxdameraulevenshtein

ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
[uWSGI] getting INI configuration from /home/user/FACT_core/src/config/uwsgi_config.ini
*** Starting uWSGI 2.0.19.1 (64bit) on [Sun Feb  7 16:45:12 2021] ***
[...]

PR #539, opened 2021-02-08
fixed bug in live stats

jstucke:
this should fix a bug with the live stats on the "home" page when no stats can be retrieved from the DB (see #538)


Enkelmann (approved):
LGTM

PR #547, opened 2021-02-25
python 3.5 + xenial deprecation

jstucke:


src/install/common.py
```diff
@@ -113,3 +97,13 @@ def main(distribution):  # pylint: disable=too-many-statements
         Path('start_all_installed_fact_components').symlink_to('src/start_fact.py')
 
     return 0
+
+
+def _update_package_sources(distribution):
+    if distribution == 'fedora':
+        logging.info('Updating system')
```
dorpvom, line 104: Yeah I know - picky - but ... why keep the code duplication?

src/start_fact_db.py
```diff
@@ -18,43 +18,42 @@
...
+        self.run = False
+
+    def main(self):
+        while self.run:
+            self.work_load_stat.update()
+            sleep(5)
+            if self.args.testing:
+                break
+
+    def shutdown(self):
```
dorpvom, line 52: You should also use this function. Not only define it ;)
I'm suprised pylint didn't blink on that...

src/start_fact_frontend.py
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
 '''
     Firmware Analysis and Comparison Tool (FACT)
-    Copyright (C) 2015-2020  Fraunhofer FKIE
+    Copyright (C) 2015-2021  Fraunhofer FKIE
```
dorpvom, line 4: Might I ask why you did not refactor this module akin to your refactoring of backend and db?

dorpvom (changes requested):

src/install/common.py
jstucke: fixed

src/start_fact_db.py
jstucke: Oops... fixed

src/start_fact_frontend.py
jstucke: because pylint did not object (there were already some "pylint: disable" comments)
changed and refactored common code to new base class

dorpvom (approved):

PR #551, opened 2021-03-10
pre_install.sh: Check if CODENAME is nonempty

maringuu:
I didn't have `lsb-release` installed and then wondered why I got `E: Malformed entry 50 in list file /etc/apt/sources.list` later. Now the `pre_install.sh` will print an error when CODENAME is empty.


jstucke (approved):
Hi and thank you for your contribution! This fix should work fine.

PR #553, opened 2021-03-16
bugfix: pinned sqlalchemy

jstucke:
- pinned version of `sqlalchemy` because of errors in combination with `flask_sqlalchemy`


Enkelmann (approved):
LGTM

PR #574, opened 2021-05-06
Lazy loading analysis tags

dorpvom:
As of yet, there is an extra scheduler that keeps track of new analysis tags and propagates them to the parent firmware.

This new mechanism lazy loads the analysis tags whenever an object is retrieved from the database, thus making a central storage of all tags unnecessary.
This also fixes some unwanted behavior related to the tag _scheduling_ such as
- outdated analysis tags are not removed,
- files that live in multiple firmware containers propagate tags only to one of them,
- tags that can come with multiple different values are overwritten

As trade-off to this slightly slower loading of results, the removal of tag _scheduling_ improves the speed of the analysis process and removes one process that runs at all times.

Plus: One super naggy test should run stable now (test_full_cycle_with_tags).


src/storage/db_interface_common.py
```diff
@@ -310,3 +311,50 @@ def release_unpacking_lock(self, uid):
...
+                        self.file_objects,
+                        '$_id',
+                        match={
+                            'virtual_file_path.{}'.format(uid): {'$exists': 'true'},
+                            f'processed_analysis.{plugin}.tags': {'$exists': 'true'}
+                        }
+                    )
+                )
+            )
+        return self.get_objects_by_uid_list(uids, analysis_filter=plugins_with_tag_propagation)
```
jstucke, line 350: If at some point the performance should prove to be a problem, we should think about a solution with fewer DB queries. The performance seems to be fine, though

src/storage/db_interface_common.py
```diff
@@ -310,3 +311,50 @@ def release_unpacking_lock(self, uid):
...
+        ]
+        uids = set()
+        for plugin in plugins_with_tag_propagation:
+            uids.update(
+                set(
+                    get_list_of_all_values(
+                        self.file_objects,
+                        '$_id',
+                        match={
+                            'virtual_file_path.{}'.format(uid): {'$exists': 'true'},
```
jstucke, line 344: we could also use the field "parent_firmware_uids" (could be a bit more intuitive)

```suggestion
                            'parent_firmware_uids': uid,
```

src/storage/db_interface_common.py
```diff
@@ -310,3 +311,50 @@ def release_unpacking_lock(self, uid):
...
+
+    def _collect_analysis_tags_from_children(self, uid):
+        children = self._fetch_children_with_tags(uid)
+        unique_tags = {}
+        for child in children:
+            self._collect_analysis_tags(child, unique_tags)
+        return unique_tags
+
+    def _collect_analysis_tags(self, file_object, analysis_tags):
+        for name, analysis in [(n, a) for n, a in file_object.processed_analysis.items() if 'tags' in a]:
```
jstucke, line 323: You had to squeeze that into one line, didn't you? :sweat_smile: 
Converting it to a generator could be a bit more efficient (omits the intermediate list creation)

```suggestion
        for name, analysis in ((n, a) for n, a in file_object.processed_analysis.items() if 'tags' in a):
```

src/storage/db_interface_common.py
```diff
@@ -310,3 +311,50 @@ def release_unpacking_lock(self, uid):
...
+            if 'file_system_flag' in analysis and analysis['file_system_flag']:
+                analysis = self.retrieve_analysis(file_object.processed_analysis, analysis_filter=[name, ])[name]
```
jstucke, line 325: why `[name, ]` and not just `[name]`?

src/storage/db_interface_common.py
```diff
@@ -310,3 +311,50 @@ def release_unpacking_lock(self, uid):
...
+                    )
+                )
+            )
+        return self.get_objects_by_uid_list(uids, analysis_filter=plugins_with_tag_propagation)
+
+
+def append_unique_tag(unique_tags: Dict[str, dict], tag: dict, plugin_name: str, tag_type: str) -> None:
+    if plugin_name in unique_tags:
+        if tag_type in unique_tags[plugin_name] and tag not in unique_tags[plugin_name].values():
+            unique_tags[plugin_name][f'{tag_type}-alt-{random.randint(0, 1000)}'] = tag
```
jstucke, line 356: Is this necessary for multiple results of the same kind?
I don't like the random(9999) solutions too much we could also use the length:
```suggestion
            unique_tags[plugin_name][f'{tag_type}-alt-{len(unique_tags[plugin_name])}'] = tag
```

src/storage/db_interface_common.py
dorpvom: Agreed

src/storage/db_interface_common.py
dorpvom: Doing that some tests fail. Where is this set anyway? I grepped it and found only a retrieval from previously stored database entries (and tests):

```
➜  FACT_core git:(poc-lazy-tagging) ✗ grep 'parent_firmware_uids =' -r src 
src/storage/db_interface_common.py:        file_object.parent_firmware_uids = set(entry['parent_firmware_uids'])
src/objects/file.py:        self.parent_firmware_uids = set()
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/scheduler/test_current_analyses.py:        fo.parent_firmware_uids = {'parent_uid'}
src/test/unit/helperFunctions/test_object_conversion.py:    fo.parent_firmware_uids = ['parent_uid']
src/test/integration/storage/test_db_interface_frontend.py:        child_fo.parent_firmware_uids = [uid]
src/test/integration/storage/test_db_interface_frontend.py:        test_fo.parent_firmware_uids = ['missing_parent_uid']
src/test/integration/storage/test_db_interface.py:            test_fo.parent_firmware_uids = parent_uids
src/test/integration/web_interface/rest/test_rest_missing_analyses.py:        test_fo.parent_firmware_uids = ['missing_uid']
src/test/acceptance/test_advanced_search.py:        self.child_fo.parent_firmware_uids = [uid]
```

src/storage/db_interface_common.py
dorpvom: Sure I did. Otherwise our guy McCabe would have come knocking :wink:

src/storage/db_interface_common.py
dorpvom: Nagging insecurity. Once I defined sth. similar to `['string']` and when I accessed it's first element I got `'s'`.
You want me to change it?

src/storage/db_interface_common.py
dorpvom: This does not solve anything. Two _OS version_ tags would still generate the same key when they e.g. have values _3.10.20_ and _3.10.35_. In case you only have two it works due to the `-alt-` but for three linux versions you overwrite a value. Dicts are not hashable, aye? The _safe_ but nasty solution is an iterating while / break to identify the next unused key starting at _-alt-0_.

src/storage/db_interface_common.py
```diff
@@ -167,20 +165,20 @@ def retrieve_analysis(self, sanitized_dict, analysis_filter=None):
...
                 if sanitized_dict[key]['file_system_flag']:
-                    logging.debug('Retrieving stored file {}'.format(key))
+                    logging.debug(f'Retrieving stored file {key}')
                     sanitized_dict[key].pop('file_system_flag')
                     sanitized_dict[key] = self._retrieve_binaries(sanitized_dict, key)
                 else:
                     sanitized_dict[key].pop('file_system_flag')
             except (KeyError, IndexError, AttributeError, TypeError, pickle.PickleError) as error:
-                logging.debug('Could not retrieve information: {} {}'.format(type(error), error))
+                logging.debug(f'Could not retrieve information: {type(error)} {error}')
```
jstucke, line 174: While we are at it, we should probably change this to `error()` or at least `warning()` and add `exc_info=True` (and maybe remove one of the `{error}`s or both)

src/storage/db_interface_common.py
```diff
@@ -33,9 +34,7 @@ def _setup_database_mapping(self):
     def existence_quick_check(self, uid):
         if self.is_firmware(uid):
             return True
-        if self.is_file_object(uid):
-            return True
-        return False
+        return self.is_file_object(uid)
```
jstucke, line 37: even shorter: `return self.is_firmware(uid) or self.is_file_object(uid)`

src/storage/db_interface_common.py
```diff
@@ -269,7 +267,7 @@ def _get_summary_of_one(file_object, selected_analysis):
                 for item in file_object.processed_analysis[selected_analysis]['summary']:
                     summary[item] = [file_object.uid]
         except (AttributeError, KeyError) as err:
-            logging.warning('Could not get summary: {} {}'.format(type(err), err))
+            logging.warning(f'Could not get summary: {type(err)} {err}')
```
jstucke, line 270: see above

jstucke (approved):

jstucke (approved):

PR #594, opened 2021-06-15
string wrap problem fix unpacker.html

Mic27M:


jstucke (changes requested):
When I tried it, it did not work for me (in Firefox and Chromium) -- even though I saw it work for you. It worked when I moved the style attributes to the `<code>` block though so that should hopefully fix it.

jstucke (approved):
Seems to work well

PR #592, opened 2021-06-14
Rework install scripts of plugins

maringuu:
A new enviroment variable `FACT_INSTALLER_SKIP_DOCKER` is introduced. When this variable is set the installer will not pull/build any docker images.

Also two flags are added to the installer to pull/build the required docker images for the backend or frontend.

To make this seperation possible the plugin install scripts are split in two: `install.sh` and `install_docker.sh`.
The latter contains the pulling/building of docker images.

With this seperation it will be possible to use the installer in `Dockerfile`'s.

TODO:
- [x] Consider if an environment variable is the right way to communicate to the installer that no docker images shall be build/pulled
- [x] Documentation

PR #595, opened 2021-06-16
Create Dockerfile

maringuu:
Depends on #592 

This creats a `Dockerfile` which contains a FACT installation which should work out of the box.

I tested the analysis with a simple text file and the analysis seems to work.
I had to introduce a hacky workaround that I am not happy about to make this work https://github.com/fkie-cad/FACT_core/blob/1ebac9e9d666dd95db12f37ca83bbc5f5d8d3819/src/unpacker/unpack_base.py#L23-L27

The problem here is that the fact-extractor docker container writes its output to `/tmp/extractor` via a bind mount.
When running FACT in docker the docker daemon will create `/tmp/extractor` on the host; meaning that `/tmp/extractor` is not avaiable to the dockerised FACT. The workaround is to add `-v /tmp:/tmp` to `docker run`.

----
The problem described above is relevevant in two places.
```
# ripgrep "docker run"
src/helperFunctions/pdf.py
30:        f'docker run -m 512m -v {folder}:/tmp/interface --rm fkiecad/fact_pdf_report'

src/unpacker/unpack_base.py
24:            'docker run --privileged -m {}m -v /dev:/dev -v {}:/tmp/extractor --rm fkiecad/fact_extractor --chown {}:{}'.format(


```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -58,6 +58,11 @@ $ src/install.py
 
 More advanced setup options can be viewed using the help function of the installer `$ src/install.py --help` and are explained in some detailed in the following paragraphs.
 
+If the environment variable `FACT_INSTALLER_SKIP_DOCKER` is set the installer
+will skip all pulling/building of docker images.
+This is primarily used for the docker container of FACT but can also be used to
+save some time when you already have the images.
+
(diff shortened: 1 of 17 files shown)


maringuu:
Closing in favor of #607

PR #598, opened 2021-06-23
system health live updates

jstucke:
Changed system health page from reloading to live updates (à la AJAX)

### DIFF
--- a/src/storage/db_interface_statistic.py
+++ b/src/storage/db_interface_statistic.py
@@ -1,5 +1,7 @@
 import logging
-import sys
+from typing import List
+
+from pymongo.errors import PyMongoError
 
 from storage.mongo_interface import MongoInterface
 
@@ -33,8 +35,8 @@ class StatisticDbUpdater(StatisticDb):
             self.statistic.delete_many({'_id': identifier})
             content_dict['_id'] = identifier
             self.statistic.insert_one(content_dict)
-        except Exception as e:
-            logging.error("Could not store statistic {}: {} - {}".format(identifier, sys.exc_info()[0].__name__, e))
+        except PyMongoError as err:
+            logging.error(f"Could not store statistic {identifier} ({err})", exc_info=True)
 
 
 class StatisticDbViewer(StatisticDb):
@@ -46,3 +48,6 @@ class StatisticDbViewer(StatisticDb):
 
     def get_statistic(self, identifier):
         return self.statistic.find_one({'_id': identifier})
+
+    def get_statistics(self, *identifiers: str) -> List[dict]:
+        return list(self.statistic.find({'_id': {'$in': identifiers}}))
--- a/src/storage/db_interface_statistic.py
+++ b/src/storage/db_interface_statistic.py
@@ -49,5 +49,5 @@ class StatisticDbViewer(StatisticDb):
     def get_statistic(self, identifier):
         return self.statistic.find_one({'_id': identifier})
 
-    def get_statistics(self, *identifiers: str) -> List[dict]:
+    def get_stats_list(self, *identifiers: str) -> List[dict]:
         return list(self.statistic.find({'_id': {'$in': identifiers}}))
--- a/src/test/unit/web_interface/test_app_ajax_routes.py
+++ b/src/test/unit/web_interface/test_app_ajax_routes.py
@@ -1,3 +1,5 @@
+# pylint: disable=wrong-import-order
+
 import pytest
 
 from test.common_helper import TEST_FW, TEST_FW_2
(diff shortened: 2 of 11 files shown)


src/storage/db_interface_statistic.py
```diff
@@ -46,3 +48,6 @@ class StatisticDbViewer(StatisticDb):
 
     def get_statistic(self, identifier):
         return self.statistic.find_one({'_id': identifier})
+
+    def get_statistics(self, *identifiers: str) -> List[dict]:
+        return list(self.statistic.find({'_id': {'$in': identifiers}}))
```
rhelmke, line 53: The Levenshtein Distance between these two names is 1, making them visually hard to differentiate (especially when mindlessly using autocompletion). Uncle Bob would say: _make meaningful distinctions_ :-).

In my opinion, from both a functional and computational stance, the plural version **could** fully replace the singular version:

1. `get_statistics` basically covers the functionality of `get_statistic`
2. `find_one` is syntactic sugar for `find().limit(1)`
3. the `$in`-operator should have negligible performance impact when `len(identifiers) == 1`

**BUT** we either need to return a `Union[str, List[str]]`, or touch a modest amount of other files in the FACT sources (a quick symbol search yields 11 files).
I guess both proposals exceed the scope of this PR, which is why I wanted to comment, but not request any changes :-)

src/web_interface/static/js/system_health.js
```diff
@@ -0,0 +1,160 @@
...
+function change_button(button_id) {
+    element = document.getElementById(button_id);
+    if (element.classList.contains("fa-caret-down")) {
+        element.classList.remove("fa-caret-down");
+        element.classList.add("fa-caret-up");
+    } else {
+        element.classList.remove("fa-caret-up");
+        element.classList.add("fa-caret-down");
+    }
+}
```
rhelmke, line 13: According to the HTML template, the classList is statically initialized in a proper state. Then, we could also use toggle:

```suggestion
function change_button(button_id) {
    element = document.getElementById(button_id);
    ["fa-caret-down", "fa-caret-up"].forEach(class_name => element.classList.toggle(class_name));
}
```

src/web_interface/static/js/system_health.js
```diff
@@ -0,0 +1,160 @@
...
+    getSystemHealthData().then(data => data.systemHealth.map(entry => {
+        const statusElement = document.getElementById(`${entry._id}-status`);
+        statusElement.innerText = entry.status;
+        if (entry.status == "offline") {
+            statusElement.classList.add('text-danger');
+            statusElement.classList.remove('text-success');
+        } else {
+            statusElement.classList.remove('text-danger');
+            statusElement.classList.add('text-success');
+            document.getElementById(`${entry._id}-os`).innerText = entry.platform.os;
```
rhelmke, line 30: We could reduce code depth here.

```suggestion
        if (entry.status == "offline") {
            statusElement.classList.add('text-danger');
            statusElement.classList.remove('text-success');
            return;
        }
        classList.remove('text-danger');
        classList.add('text-success');
        getElementById(`${entry._id}-os`).innerText = entry.platform.os;
```

rhelmke (changes requested):

src/storage/db_interface_statistic.py
jstucke: I could rename it for the time being. How about `get_stats_list` or `get_multiple_statistics`?

src/web_interface/static/js/system_health.js
jstucke: Nice that's much better :smile:

src/web_interface/static/js/system_health.js
jstucke: changed

rhelmke (approved):

src/storage/db_interface_statistic.py
rhelmke: That's better, very much :-)

src/test/unit/web_interface/test_app_ajax_routes.py
```diff
@@ -37,6 +38,12 @@ def test_ajax_get_system_stats_error(self):
         assert result['backend_cpu_percentage'] == 'n/a'
         assert result['number_of_running_analyses'] == 'n/a'
 
+    def test_ajax_system_health(self):
+        self.mocked_interface.get_stats_list = lambda *_: [{'foo': 'bar'}]
```
dorpvom, line 42: Why the different patching compared to the test just above? Overwriting without setting it back afterwards is bad style.

dorpvom (approved):

PR #608, opened 2021-07-06
Kernel config hardening check

jstucke:
Added kernel config hardening check to kernel config plugin based on [kconfig-hardened-check](https://github.com/a13xp0p0v/kconfig-hardened-check) and [linux-kernel-defence-map](https://github.com/a13xp0p0v/linux-kernel-defence-map)

### DIFF
--- a/src/plugins/analysis/kernel_config/code/kernel_config.py
+++ b/src/plugins/analysis/kernel_config/code/kernel_config.py
@@ -1,51 +1,23 @@
-import json
-import logging
 import re
 import sys
-from json import JSONDecodeError
 from pathlib import Path
-from tempfile import NamedTemporaryFile
 from typing import List
 
-from common_helper_process import execute_shell_command
-
 from analysis.PluginBase import AnalysisBasePlugin
-from helperFunctions.fileSystem import get_src_dir
 from objects.file import FileObject
 
 try:
     from ..internal.decomp import decompress
+    from ..internal.kernel_config_hardening_check import check_kernel_hardening
+    from ..internal.checksec_check_kernel import check_kernel_config, CHECKSEC_PATH
 except ImportError:
     sys.path.append(str(Path(__file__).parent.parent / 'internal'))
     from decomp import decompress
+    from kernel_config_hardening_check import check_kernel_hardening
+    from checksec_check_kernel import check_kernel_config, CHECKSEC_PATH
 
 
 MAGIC_WORD = b'IKCFG_ST\037\213'
-CHECKSEC_PATH = Path(get_src_dir()) / 'bin' / 'checksec'
-
-KERNEL_WHITELIST = [
-    'kernel_heap_randomization', 'gcc_stack_protector', 'gcc_stack_protector_strong',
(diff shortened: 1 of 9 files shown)


src/plugins/analysis/kernel_config/internal/kernel_config_hardening_check.py
```diff
@@ -0,0 +1,108 @@
...
+        logging.warning('kconfig-hardened-check analysis failed')
+    return []
+
+
+def _add_protection_info(hardening_result: List[List[str]]) -> List[Tuple[str, str, str, str, str, List[str]]]:
+    full_result = []
+    for single_result in hardening_result:
+        config_key = single_result[0]
+        protection_info = PROTECTS_AGAINST[config_key] if config_key in PROTECTS_AGAINST else []
+        full_result.append((*single_result, protection_info))
```
rhelmke, line 107: I'm not quite sure on how to feel about the return structure here. I understand why you did that, but I can't derive its content.

Maybe a `List[namedtuple]` instead of `List[Tuple[str, str, str, str, str, List[str]]]` to improve readability/clarity? E.g.:

```python
from collections import namedtuple

# [...]

Result = namedtuple('Result', ['name', 'desired', 'decision', 'reason', 'result', 'mitigates'])

# [...]

full_result.append(Result(*single_result, protection_info))
```

src/plugins/analysis/kernel_config/code/kernel_config.py
```diff
@@ -1,59 +1,31 @@
...
-    'config_grkernsec_kmem', 'config_grkernsec_io', 'config_grkernsec_modharden',
-    'config_modules', 'config_grkernsec_chroot', 'config_grkernsec_harden_ptrace',
-    'config_grkernsec_randnet', 'config_grkernsec_blackhole', 'config_grkernsec_brute',
-    'config_grkernsec_hidesym'
-]
 
 
 class AnalysisPlugin(AnalysisBasePlugin):
     NAME = 'kernel_config'
     DESCRIPTION = 'Heuristics to find plaintext and image-embedded kernel configurations (IKCONFIG=[y|m])'
```
rhelmke, line 25: Maybe it's time to extend the description here.

```python
    DESCRIPTION = 'Heuristics to find and analyze Linux Kernel configurations via checksec and kconfig-hardened-check'
```

?

src/plugins/analysis/kernel_config/internal/checksec_check_kernel.py
```diff
@@ -0,0 +1,59 @@
...
+def check_kernel_config(kernel_config: str) -> dict:
+    try:
+        with NamedTemporaryFile() as fp:
+            fp.write(kernel_config.encode())
+            fp.seek(0)
+            command = f'{CHECKSEC_PATH} --kernel={fp.name} --output=json 2>/dev/null'
+            result = json.loads(execute_shell_command(command))
+            whitelist_configs(result)
+            return result
+    except (JSONDecodeError, KeyError):
```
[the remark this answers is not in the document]
jstucke: `execute_shell_command` only raises an exception if the parameter `check` is `True` (which it is not by default)

rhelmke (approved):

PR #610, opened 2021-07-07
FACT logs view

any333:


src/helperFunctions/program_setup.py
```diff
@@ -58,14 +60,15 @@ def _get_console_output_level(debug_flag):
     return logging.INFO
 
 
-def _setup_logging(config, args):
+def _setup_logging(config, args, component):
```
jstucke, line 63: A default value for component would make sense and prevent one test from failing
```suggestion
def _setup_logging(config, args, component=None):
```

src/test/unit/web_interface/test_app_find_logs.py
```diff
@@ -0,0 +1,28 @@
...
+class IntercomMock:
+    @staticmethod
+    def get_backend_logs():
+        return ['String1', 'String2', 'String3']
+
+    def shutdown(self):
+        pass
+
+
+class TestShowStatistic(WebInterfaceTest):
```
jstucke, line 16: Should probably be renamed

src/test/unit/web_interface/test_app_find_logs.py
```diff
@@ -0,0 +1,28 @@
...
+    def setUp(self, db_mock=None):
+        super().setUp(db_mock=IntercomMock)
+        self.config['Logging']['logFile'] = 'NonExistentFile'
```
jstucke, line 19: it would be better to move this line to the start of `test_logs_available()` to prevent possible side effects

src/test/unit/web_interface/test_app_find_logs.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_logs_available(self):
```
jstucke, line 21: It would probably make sense to name this something like `test_backend_logs` to be more consistent with the other test

src/web_interface/rest/helper.py
```diff
@@ -5,6 +5,7 @@
 
 import calendar
 import json
+import logging
```
jstucke, line 8: seems to be unused

src/web_interface/templates/logs.html
```diff
@@ -0,0 +1,45 @@
...
+        overflow-y: scroll;
+        overflow-x: hidden;
+        display: block;
+        font-family: monospace;
+        margin: 0 0;
+        color: #f8f8f8;
+        background-color: #333;
+        max-height: 500px;
+    }
+    .logs-td {
```
jstucke, line 20: seems to be unused

src/web_interface/components/miscellaneous_routes.py
```diff
@@ -127,3 +131,17 @@ def _find_failed_analyses(self):
...
+    @roles_accepted(*PRIVILEGES['view_logs'])
+    def _app_find_logs(self):
+        with ConnectTo(InterComFrontEndBinding, self._config) as sc:
+            backend_logs = '\n'.join(sc.get_backend_logs())
+        frontend_logs = '\n'.join(self.get_frontend_logs())
+
+        return render_template('logs.html', backend_logs=backend_logs, frontend_logs=frontend_logs)
+
+    def get_frontend_logs(self):
+        backend_logs = Path(get_log_file_for_component('frontend', self._config))
```
jstucke, line 144: variable should probably be renamed

jstucke (changes requested):
An important part is missing: the entry in the navigation bar (admin drop down menu) to actually get to the new page

The test `test_all_listeners_started` currently fails and needs to be adjusted from 9 to 10 workers

What is more, the logs page looks a bit weird if the logs are empty and it would be better to display some message like "No log entries to show" in this case

jstucke (approved):

PR #612, opened 2021-07-13
Analsysis/Compare: Don't die on failing import

maringuu:
This wraps importing plugins in a try-except block.

Now a plugin can contain invalid code and the fact backend will still start.

### DIFF
--- a/src/compare/compare.py
+++ b/src/compare/compare.py
@@ -78,8 +78,17 @@ class Compare:
     def _init_plugins(self):
         self.source = import_plugins('compare.plugins', 'plugins/compare')  # pylint: disable=attribute-defined-outside-init
         for plugin_name in self.source.list_plugins():
-            plugin = self.source.load_plugin(plugin_name)
-            plugin.ComparePlugin(self, config=self.config, db_interface=self.db_interface)
+            try:
+                plugin = self.source.load_plugin(plugin_name)
+            except Exception as e:
+                # This exception could be caused by invalid syntax
+                # Invalid syntax may be caused by plugin dependencies that got
+                # upgraded to incompatible versions
+                logging.error(f'Could not import plugin {plugin_name} due to Exception: {e}')
+            else:
+                plugin.ComparePlugin(self, config=self.config, db_interface=self.db_interface)
+
+
 
     def register_plugin(self, name, c_plugin_instance):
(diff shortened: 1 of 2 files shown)


src/compare/compare.py
```diff
@@ -78,8 +78,17 @@ def _setup_plugins(self):
...
         self.source = import_plugins('compare.plugins', 'plugins/compare')  # pylint: disable=attribute-defined-outside-init
         for plugin_name in self.source.list_plugins():
-            plugin = self.source.load_plugin(plugin_name)
-            plugin.ComparePlugin(self, config=self.config, db_interface=self.db_interface)
+            try:
+                plugin = self.source.load_plugin(plugin_name)
+            except Exception as e:
+                # This exception could be caused by invalid syntax
+                # Invalid syntax may be caused by plugin dependencies that got
+                # upgraded to incompatible versions
```
jstucke, line 86: The dependencies could also simply be missing (either if new dependencies were added and the user did not re-run the installation or if the user didn't install FACT correctly in the first place)

src/scheduler/Analysis.py
```diff
@@ -206,8 +206,15 @@ def register_plugin(self, name, plugin_instance):
     def load_plugins(self):
         source = import_plugins('analysis.plugins', 'plugins/analysis')
         for plugin_name in source.list_plugins():
-            plugin = source.load_plugin(plugin_name)
-            plugin.AnalysisPlugin(self, config=self.config)
+            try:
+                plugin = source.load_plugin(plugin_name)
+            except Exception as e:
```
jstucke, line 211: see above

src/scheduler/Analysis.py
```diff
@@ -206,8 +206,15 @@ def register_plugin(self, name, plugin_instance):
...
+                # This exception could be caused by invalid syntax
+                # Invalid syntax may be caused by plugin dependencies that got
+                # upgraded to incompatible versions
+                logging.error(f'Could not import plugin {plugin_name} due to Exception: {e}')
```
jstucke, line 215: see above

jstucke (changes requested):
Seems to work fine. Some small issues

maringuu:
Adressed review in #616

PR #613, opened 2021-07-16
Plugin install refactor

maringuu:


jstucke (approved):
Reviewed in #592 and since unchanged

PR #617, opened 2021-07-30
install.md: Mention Dockerfile

maringuu:


jstucke (commented):
It would probably make sense to put this in the `README.md` so that people find it more easily (maybe above or below the vagrant mention with its own header "Docker")

maringuu:
Seems link a good idea.
I didn't put it under its own headline though because we don't have a dockerhub image yet.

jstucke (approved):

PR #619, opened 2021-08-03
objects/firmware: Document Firmware

maringuu:


src/objects/firmware.py
```diff
@@ -6,17 +6,32 @@
...
+        #: The version of the firmware in no specific format
+        self.version: str = None
+        #: The type of device that the firmware runs on.
+        #: E.g. "router"
+        self.device_class: str = None
+        self.vendor: str = None
+        #: The part of the firmware that is this object represents.
+        #: An empty string represents "complete".
+        #: Some possible values are "kernel", "bootloader" and "root-fs"
+        self.part: str = ''
```
rhelmke, line 29: Overall, I'd like a little bit less verbosity, more explanation, and clearer definitions in your comments. To clarify what I mean by that, below a little example :-)

```suggestion
        #: Specifies the parts of an embedded system that are contained in the firmware.
        #: While this meta data string can be freely defined during firmware upload,
        #: FACT provides a preset of frequently used values: 'complete', 'kernel', ' bootloader', and 'root-fs'.
        #: The firmware image is assumed to be 'complete' if the assigned value is an empty string.
        self.part: str = ''
```

src/objects/firmware.py
```diff
@@ -6,17 +6,32 @@
...
+        # The release date of the firmware in the format "YYYY-MM-DD".
+        self.release_date: str = None
+        #: A dict where the keys represent tags and the values represent the
+        #: colors of the tags.
+        #: Possible values are defined in :class:`helperFunctions.tag.TagColor`.
         self.tags = dict()
         self._update_root_id_and_virtual_path()
```
rhelmke, line 37: I understand that the setters below are self-explanatory (and additionally wonder why the trivial ones exist).
Yet, a proper Docstring for all methods would still be nice for the Sphinx documentation.

rhelmke (changes requested):

src/objects/firmware.py
```diff
@@ -6,17 +6,32 @@
 
 class Firmware(FileObject):
     '''
-    This objects represents a firmware
+    This class represents a firmware.
+    It esentially is a FileObject with some more attributes that only make
```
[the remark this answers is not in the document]
rhelmke: *essentially :-). But how about this?

```suggestion
    It is a specialization of :meth:`objects.file.FileObject that adds a variety of firmware-specific attributes.
```

src/objects/firmware.py
```diff
@@ -1,59 +1,79 @@
...
-    def set_firmware_version(self, version):
-        self.version = version
-
-    def set_device_class(self, device_class):
-        self.device_class = device_class
-
     def set_binary(self, binary):
+        '''
+        See :meth:`objects.file.FileObject.set_binary`.
+        Additionally `self` is set to be the root file of the fimware.
```
rhelmke, line 54: This is an implementation detail that is irrelevant here.
```suggestion
```

rhelmke (changes requested):
Looks better. :-) Please apply the requested changes and I can approve.

maringuu:
> Please apply the requested changes

Done.

rhelmke (approved):

maringuu:
Fixed the typo, the CI works now.

src/objects/firmware.py
```diff
@@ -1,56 +1,118 @@
...
-        #: The firmware image is assumed to be 'complete' if the assigned value is an empty string.
+        #: FACT provides a preset of frequently used values: `complete`, `kernel`, `bootloader`, and `root-fs`.
+        #:
+        #: This attribute is **optional**. The firmware image is assumed to be `complete` if the assigned/default value is an empty string.
         self.part: str = ''
-        # The release date of the firmware in the format "YYYY-MM-DD".
+
+        #: Release date string of this firmware version in `ISO 8601 <https://en.wikipedia.org/wiki/ISO_8601>`_ `YYYY-MM-DD` format.
+        #:
+        #: This attribute is **optional** and may be `None`.
```
jstucke, line 87: It would be good to add the info that it is automatically set to `'1970-01-01'` if it is empty

src/objects/firmware.py
```diff
@@ -1,56 +1,118 @@
...
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        #: The name of the device that the firmware runs on
+
+        #: Device name string identifier.
+        #: Can be freely defined during upload.
+        #:
+        #: This attribute is **mandatory** and shall never be `None`.
         self.device_name: str = None
```
jstucke, line 59: The optional parameters should use the type `Optional[]` from `typing` (i.e. `Optional[str]` here)

jstucke (commented):
The Firmware docs look pretty good now :smile: 
Two small issues:

jstucke (approved):

PR #622, opened 2021-08-06
Flask2.0

maringuu:
~Depends on #618 but it could easily be rebased on master~


src/web_interface/frontend_main.py
```diff
@@ -27,10 +27,11 @@ def __init__(self, config=None):
         logging.info("Web front end online")
 
     def _setup_app(self):
+        # TODO replace __name__ see Flask constructor docs
         self.app = Flask(__name__)
         self.app.config.from_object(__name__)
 
-        Flask.secret_key = os.urandom(24)
```
maringuu, line 33: I wonder how this worked, since it does not actually set the `secret_key` in `app.config`

dorpvom:
I've checked this and it doesn't work. Scenario:
1. I create a user and role according to the documentation.
2. I use the current master with activated authentication.
3. I can successfully log into FACT using the create credentials.
4. I switch to this branch and run the installation (after manually removing flask_security)
5. I can't log into FACT using the credentials or even credentials created after switching the branch.

dorpvom:
We have to look for reasons why this behavior occurs. Maybe flask-security-too has a differend handle on the database.

maringuu:
I added alembic to handle database migrations. It might be a bit overkill but it seems to be standart in the flask ecosystem.

To upgrade the database run `alembic upgrade head`.
For downgrading use `alembic downgrade base`.

For me the problem is fixed.

@dorpvom Can you confirm?

dorpvom:
> 
> 
> I added alembic to handle database migrations. It might be a bit overkill but it seems to be standart in the flask ecosystem.
> 
> To upgrade the database run `alembic upgrade head`.
> For downgrading use `alembic downgrade base`.
> 
> For me the problem is fixed.
> 
> @dorpvom Can you confirm?

We'll discuss this tomorrow

maringuu:
Note that the tests fail in the CI the packages are not upgraded in the CI.

src/migrate_database.py
```diff
@@ -0,0 +1,79 @@
...
+    users = [row[0] for row in cur.execute('SELECT id FROM user')]
+    for uid in users:
+        cur.execute('UPDATE user SET fs_uniquifier = ? WHERE user.id = ?', (uuid.uuid4().hex, uid))
+
+    # Due to limitiations in SQLite we have to create a temporary table
+    cur.execute('''
+        CREATE TABLE "user_tmp" (
+            "id"			INTEGER NOT NULL,
+            "api_key"		VARCHAR(255),
+            "email"			VARCHAR(255),
```
jstucke, line 22: shouldn't these be `UNIQUE`?

src/test/unit/web_interface/security/test_user_role_db_interface.py
```diff
@@ -14,6 +14,7 @@ def setUp(self):
         self.test_app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///'
         self.test_app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
         self.test_app.config['SECURITY_PASSWORD_SALT'] = 'salt123'
+        self.test_app.config['SECRET_KEY'] = 'key123'
```
jstucke, line 17: > test/unit/web_interface/security/test_user_role_db_interface.py:7: in <module>
    from web_interface.security.authentication import create_user_interface
E   ImportError: cannot import name 'create_user_interface' from 'web_interface.security.authentication'

src/web_interface/security/authentication.py
```diff
@@ -1,27 +1,35 @@
...
+    # flask-security we change the validation mapper of 'email'.
+    app.config['SECURITY_USER_IDENTITY_ATTRIBUTES'] = [{'email': {"mapper": uia_username_mapper, "case_insensitive": True}}]
 
-    user_interface = create_user_interface(db)
-    security = Security(app, user_interface)
 
-    _add_apikey_handler(security, user_interface)
-    return db, user_interface
+def add_flask_security_to_app(app):
+    # db.Table etc cause errors but are correct
```
jstucke, line 28: I did not observe any errors. Should I see errors?

jstucke (commented):
Everything seems to work. It might be a problem to directly upgrade, because `flask-security` might still be installed. Maybe we make sure that it is uninstalled during the installation?

src/migrate_database.py
```diff
@@ -0,0 +1,79 @@
...
+            "password"		VARCHAR(255),
+            "active"		BOOLEAN,
+            "confirmed_at"	DATETIME,
+            "fs_uniquifier"	VARCHAR(64)	NOT NULL,
+            CHECK(active IN (0,1)),
+            PRIMARY KEY("id")
+        );''')
+    cur.execute('INSERT INTO "user_tmp" SELECT * FROM "user" WHERE true')
+    cur.execute('DROP TABLE "user"')
+    cur.execute('ALTER TABLE "user_tmp" RENAME TO "user"')
```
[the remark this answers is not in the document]
maringuu: The ALTER TABLE statement is very limeted in SQLite. I added an explaining comment above.

src/web_interface/security/authentication.py
maringuu: For me pylint complains that `Table` is not a member of `db`.
Pylint is wrong so I disabled the warnings and added this comment to explain why I disabled them.

src/web_interface/frontend_main.py
```diff
@@ -27,11 +27,13 @@ def __init__(self, config=None):
         logging.info("Web front end online")
 
     def _setup_app(self):
+        # FIXME replace __name__ see Flask constructor docs
```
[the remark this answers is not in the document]
maringuu: I didn't know about `flask_app_wrapper.py`. I think you are right. I removed the comment.

maringuu:
I don't think Lintly is correct here.
When conforming to the standart (I used autopep8) it becomes less readable imo.

src/test/unit/test_manage_users.py
```diff
@@ -51,9 +54,24 @@ def test_integration_try_actions(monkeypatch, action_and_inputs):
...
+        {'data_storage': {
+            # We want an in memory database for testing
+            'user_database': 'sqlite://',
+            'password_salt': "salt"
+            },
+        'ExpertSettings': {
+            'authentication': 'true'
+            },
+        }
+    )
```
jstucke, line 70: indentation fix:
```suggestion
    parser.read_dict({
        'data_storage': {
            # We want an in memory database for testing
            'user_database': 'sqlite://',
            'password_salt': "salt"
        },
        'ExpertSettings': {
            'authentication': 'true'
        },
    })
```

jstucke (approved):

PR #623, opened 2021-08-12
Virtualenv compat

jstucke:
This PR should (hopefully) make FACT compatible to running Python from a venv.

- converted python package dependencies to pip dependencies 
- updated `tlsh` function for newer version 
    - added a test
    - bumped hash/tlsh plugin versions
    - observation: newer tlsh version is backwards compatible to old hashes and can compute hashes for string lenghts >=50 (old: >=256)
        - old tlsh version is not upwards compatible with new hashes (they have a new prefix)
- refactoring

### DIFF
--- a/src/helperFunctions/install.py
+++ b/src/helperFunctions/install.py
@@ -2,6 +2,7 @@ import configparser
 import logging
 import os
 import shutil
+import sys
 from pathlib import Path
 from typing import List, Tuple, Union
 
@@ -239,3 +240,7 @@ def load_main_config() -> configparser.ConfigParser:
         raise InstallationError('Could not load config at path {}'.format(config_path))
     config.read(str(config_path))
     return config
+
+
+def is_virtualenv() -> bool:
+    return sys.prefix != getattr(sys, 'base_prefix', getattr(sys, 'real_prefix', None))
--- a/src/install/common.py
+++ b/src/install/common.py
@@ -6,18 +6,19 @@ from common_helper_process import execute_shell_command_get_return_code
 
 from helperFunctions.install import (
     InstallationError, OperateInDirectory, apt_install_packages, apt_remove_packages, apt_update_sources,
-    dnf_install_packages, dnf_remove_packages, dnf_update_sources, install_github_project, pip3_install_packages
+    dnf_install_packages, dnf_remove_packages, dnf_update_sources, install_github_project, is_virtualenv,
+    pip3_install_packages
 )
 
 BIN_DIR = Path(__file__).parent.parent / 'bin'
 
 
-def install_pip(python_command):
-    logging.info('Installing {} pip'.format(python_command))
-    for command in ['wget https://bootstrap.pypa.io/get-pip.py', 'sudo -EH {} get-pip.py'.format(python_command), 'rm get-pip.py']:
+def install_pip():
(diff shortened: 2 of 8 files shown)


dorpvom:
Fedora build breaks on pip installation of ssdeep

src/plugins/analysis/hash/code/hash.py
```diff
@@ -34,11 +34,11 @@ def process_object(self, file_object):
...
         file_object.processed_analysis[self.NAME] = {}
-        for h in self.hashes_to_create:
-            if h in algorithms_available:
-                file_object.processed_analysis[self.NAME][h] = get_hash(h, file_object.binary)
+        for hash_ in self.hashes_to_create:
+            if hash_ in algorithms_available:
+                file_object.processed_analysis[self.NAME][hash_] = get_hash(hash_, file_object.binary)
             else:
-                logging.debug('algorithm {} not available'.format(h))
+                logging.debug('algorithm {} not available'.format(hash_))
```
dorpvom, line 41: tzz, now here was an obvious refactoring-to-fstring-possibility

src/install/common.py
```diff
@@ -46,28 +47,25 @@ def main(distribution):  # pylint: disable=too-many-statements
...
-        dnf_install_packages('python3-ssdeep')
     else:
         apt_install_packages('libmagic-dev')
         apt_install_packages('libfuzzy-dev')
-        apt_install_packages('python3-tlsh')
-        pip3_install_packages('ssdeep')
 
     pip3_install_packages('git+https://github.com/fkie-cad/fact_helper_file.git')
-    pip3_install_packages('psutil', 'pytest', 'pytest-cov', 'pylint', 'python-magic', 'xmltodict', 'yara-python==3.7.0', 'appdirs', 'flaky')
+    pip3_install_packages('psutil', 'pytest', 'pytest-cov', 'pylint', 'python-magic', 'xmltodict', 'yara-python==3.7.0', 'appdirs', 'flaky', 'python-tlsh', 'ssdeep')
```
dorpvom, line 68: No wonder this breaks on fedora. Why was ssdeep moved out of the conditional block?

src/helperFunctions/install.py
```diff
@@ -239,3 +240,7 @@ def load_main_config() -> configparser.ConfigParser:
         raise InstallationError('Could not load config at path {}'.format(config_path))
     config.read(str(config_path))
     return config
+
+
+def is_virtualenv() -> bool:
```
dorpvom, line 245: Uargh. Short explanation? (StackOverflow post this was copied from suffices)

dorpvom (changes requested):

src/helperFunctions/install.py
jstucke: https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv :sweat_smile:

src/install/common.py
jstucke: I was under the impression that it worked in  #618. Surely, I have been mistaken. -> back to the conditional block

src/plugins/analysis/hash/code/hash.py
jstucke: changed

dorpvom (approved):

PR #628, opened 2021-08-23
Remove x editable

### LINKED ISSUE #625: bootstrap-editable is no longer maintained
A dependency, https://github.com/vitalets/x-editable, is no longer maintained. The library seems to be used in exactly one place, `src/web_interface/templates/user_management/edit_user.html`; but as far as I can tell, I am pretty sure it isn't compatible with bootstrap 4 anyways.

In any case, please consider to remove it.

jstucke:
resolves #625 

- removed x-editable dependency
- added modal for changing user roles

### DIFF
--- a/src/web_interface/components/user_management_routes.py
+++ b/src/web_interface/components/user_management_routes.py
@@ -24,8 +24,8 @@ class UserManagementRoutes(ComponentBase):
         try:
             yield session
             session.commit()
-        except (SQLAlchemyError, TypeError) as exception:
-            logging.error('error while accessing user db: {}'.format(exception))
+        except (SQLAlchemyError, TypeError):
+            logging.error('error while accessing user db:', exc_info=True)
             session.rollback()
             if error_message:
                 flash(error_message)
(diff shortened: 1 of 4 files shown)


src/web_interface/templates/user_management/edit_user.html
```diff
@@ -72,16 +36,53 @@ <h3 class="mb-3">Manage user {{ user.email }}</h3>
...
+
+                </div>
+            </div>
+        </div>
+
+
+        <!-- role help -->
+        <div id="user_role_help_modal" class="modal" role="dialog" data-backdrop="false">
+            <div class="modal-dialog modal-lg">
+                <div class="modal-content">
```
rhelmke, line 85: On a side note: Is it wise to allow superusers to delete themselves? I just locked me out of the instance by doing so :smile:

src/web_interface/components/user_management_routes.py
```diff
@@ -88,45 +87,43 @@ def _change_user_password(self, user_id):
...
         try:
-            user = self._user_db_interface.find_user(email=user_name)
+            user = self._user_db_interface.find_user(id=user_id)
         except SQLAlchemyError:
-            return 'Not found', 400
+            flash('Error when trying to change user roles')
+            return
+        if user is None:
+            flash(f'Error: user with id {user_id} not found')
+            return
```
rhelmke, line 105: I'm quite unhappy about the loss of application protocol information here :-/. Would it still be possible to respond with a 500 (`SQLAlchemyError`) and 404 (`user is None`) with the changes in this PR?

src/web_interface/templates/user_management/edit_user.html
jstucke: Probably not but who are we to disallow them to do it? :sweat_smile:

src/web_interface/components/user_management_routes.py
jstucke: The endpoint changed from a REST(-like?) endpoint to a "regular" one. So I should probably also remove the "400"

jstucke: merged it with the "change password" functionality into a common POST endpoint

src/web_interface/templates/user_management/edit_user.html
rhelmke: Fair enough.

rhelmke (approved):

PR #630, opened 2021-08-25
fixed navbar scaling for medium width

jstucke:
fixed navbar scaling for screen widths around half of FHD or QHD (which was broken since the admin dropdown menu was always displayed)

- removed FACT title from navbar
- reduced size of quick search input
    - made this input grow bigger when being focused on 
- made navbar collapsible for very small screens
- when authentication is turned on, long user names could still be problematic (because of the additional menu entry)


rhelmke (approved):
This looks good to me. As discussed, there is a follow-up fix for more robust scaling in our pipeline.

PR #631, opened 2021-08-26
Information leaks plugin - user paths and git repo

any333:


src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, plugin_administrator, config=None, recursive=True, offline_testing=False):
+        super().__init__(plugin_administrator, config=config, recursive=recursive, plugin_path=__file__,
+                         offline_testing=offline_testing)
+
+    def process_object(self, file_object: FileObject):
+        file_object.processed_analysis[self.NAME]['summary'] = []
+        if file_object.processed_analysis['file_type']['mime'] == 'text/plain':
+            self._find_artifacts(file_object)
+            file_object.processed_analysis[self.NAME]['summary'] = file_object.processed_analysis[self.NAME].keys()
```
jstucke, line 64: `.keys()` does not actually return a list and this could cause issues when trying to store the result in the database
```suggestion
            file_object.processed_analysis[self.NAME]['summary'] = list(file_object.processed_analysis[self.NAME])
```

What is more, line 85 in `_find_paths()` does practically the same, so we can just remove that and move this line down out of the if statement. This also makes line 61 unnecessary.

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _find_artifacts(self, file_object: FileObject):
+        for virtual_path_list in file_object.virtual_file_path.values():
+            for virtual_path in virtual_path_list:
+                path = virtual_path.split('|')[-1]
+                for key_path, artifact in PATH_ARTIFACT_DICT.items():
+                    if path.endswith(key_path):
+                        file_object.processed_analysis[self.NAME][artifact] = file_object.binary.decode()
+                for key_path, artifact in DIRECTORY_DICT.items():
+                    if path == key_path.split('/')[-2]:
```
jstucke, line 78: This should probably be `key_path` instead of `path` here 
```suggestion
                    if key_path == key_path.split('/')[-2]:
```
Seems like we need a test case for this

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,85 @@
...
+                        file_object.processed_analysis[self.NAME][artifact] = path
```
jstucke, line 79: It would make more sense to store the whole file path here, since only the file name could be not very informative (since the thing we matched was the parent directory and not the file itself)
```suggestion
                        file_object.processed_analysis[self.NAME][artifact] = virtual_path
```

jstucke (changes requested):

src/plugins/analysis/information_leaks/code/information_leaks.py
jstucke: also, accessing index `[-2]` could cause a `IndexError`, so we should handle that.

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,85 @@
+import re
+
+from analysis.PluginBase import AnalysisBasePlugin
+from objects.file import FileObject
+
+PATH_REGEX = {'user_paths': re.compile(rb'/home/[^/]+/[^\n \x00]+')}
```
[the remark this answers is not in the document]
lwilms: I added the new suggestions

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,85 @@
...
+    '.pytest_cache': 'pytest_cache_directory',
+    '.subversion': 'svn_user_settings_directory',
+    'subversion': 'svn_settings_directory',
+    '.idea': 'pycharm_config_directory'
+}
+
+
+class AnalysisPlugin(AnalysisBasePlugin):
+    '''
+    This Plugin creates several hashes of the file
```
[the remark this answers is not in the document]
lwilms: The description is now more fitting

src/plugins/analysis/information_leaks/view/information_leaks.html
```diff
@@ -0,0 +1,47 @@
+{% extends "analysis_plugins/general_information.html" %}
+
+{% block analysis_result_details %}
+<style>
+    .information-leaks-pre {
```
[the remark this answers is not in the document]
lwilms: I merged the master and added the code-pre.css into the html sheet

src/plugins/analysis/information_leaks/code/information_leaks.py
lwilms: Almost correctly so.  the two were switched. I also added some logic improvements.

lwilms (approved):

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,91 @@
...
+                self._find_paths(file_object, regex, label)
+        file_object.processed_analysis[self.NAME]['summary'] = list(file_object.processed_analysis[self.NAME])
+        return file_object
+
+    def _find_artifacts(self, file_object: FileObject):
+        for virtual_path_list in file_object.virtual_file_path.values():
+            for virtual_path in virtual_path_list:
+                for key_path, artifact in PATH_ARTIFACT_DICT.items():
+                    if virtual_path.endswith(key_path):
+                        file_object.processed_analysis[self.NAME][artifact] = file_object.binary.decode()
```
jstucke: After testing it again, I have to reconsider: It probably isn't a good idea to simple save the complete file content in the analysis result (and therefore in the database). If the user wants to see the content. he can do that using the preview functionality.

Instead, it is probably better to only store 
```suggestion
                        file_object.processed_analysis[self.NAME]['path_artifacts'] = artifact
```
Sadly, this makes our previous changes to the summary generation obsolete. We still want the `artifact`s in the summary.

jstucke (changes requested):

src/plugins/analysis/information_leaks/code/information_leaks.py
lwilms: Why would it make it obsolete? We would just store the file paths, just like we have done with directories or am I missing something?

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,91 @@
...
+                for key_path, artifact in DIRECTORY_DICT.items():
+                    v_path = virtual_path.split('|')
+                    if len(v_path) > 1:
+                        if v_path[-2] == key_path:
```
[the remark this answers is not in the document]
lwilms: yes, I forgot to change these lines. It is not applied, when calling the functions.

src/plugins/analysis/information_leaks/test/test_plugin_information_leaks.py
```diff
@@ -0,0 +1,78 @@
...
+        fo.processed_analysis['file_type'] = {}
+        fo.processed_analysis['file_type']['mime'] = 'text/plain'
+        fo.virtual_file_path = {
+            1: ['|home|user|project|.git|config',
+                '|home|user|some_path|.pytest_cache|some_file',
+                '|root|some_directory|some_more|.config|Code|User|settings.json',
+                '|some_home|some_user|urandom|42|some_file.uvprojx',
+                'home', '', 'h654qf"§$%74672', 'vuwreivh54r234|', '|vr4242fdsg4%%$'
+                ]
+        }
```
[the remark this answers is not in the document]
lwilms: Did not know, that with the stacking uid, nice to know for the future. Nevertheless, it was fixed in an earlier commit.

I fixed the weirdness with the brackets ;)

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,91 @@
...
+    '.subversion': 'svn_user_settings_directory',
+    'subversion': 'svn_settings_directory',
+    '.idea': 'pycharm_config_directory'
+}
+
+
+class AnalysisPlugin(AnalysisBasePlugin):
+    """
+    This Plugin searches for compilation artifacts in a firmware,
+        e.g., github repositories, IDE configs and special paths
```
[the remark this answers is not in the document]
lwilms: I added the .bash_history and .zsh_history, but for a regex with history it would require some more logic and maybe an additional directory.

I will work on this logic.

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -0,0 +1,91 @@
+import re
+
+from analysis.PluginBase import AnalysisBasePlugin
+from objects.file import FileObject
+
+PATH_REGEX = {'user_paths': re.compile(rb'/home/[^/]+/[^\n \x00]+'),
+              'root_path': re.compile(rb'/root/[^/]+/[^\n \x00]+'),
+              'var_path': re.compile(rb'/var/www/[^/]+/[^\n \x00]+')
+              }
```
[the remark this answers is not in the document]
lwilms: applied the top variant :)

src/plugins/analysis/information_leaks/code/information_leaks.py
jstucke: What I meant was that we were able to use common approach for generating the summary for all cases but now the structure is different in this case if we want to keep `artifact` as summary entry. Therefore, each part again needs its own summary generation.

jstucke (approved):

jstucke (approved):

Issue #633, opened 2021-08-28
Unused Dependency: python3-tk / python3-tkinter


mistressofjellyfish:
As a dependency, FACT installs python3-tk (Ubuntu, Debian) or python3-tkinter (Fedora).

As far as I can tell, this dependency is not used:

```
katharina@lilienfeld ~/Dokumente/fact/FACT_core [FACT_core/master]% ag "import tk"
katharina@lilienfeld ~/Dokumente/fact/FACT_core [FACT_core/master]% ag "tkinter"
src/install/dnf-pkgs-backend.txt
3:python3-tkinter
katharina@lilienfeld ~/Dokumente/fact/FACT_core [FACT_core/master]% 
```

Please consider to remove it.

jstucke:
Resolved in #638

PR #638, opened 2021-09-01
remove unused python tk dep #633

0xricksanchez:


jstucke (approved):
According to the CI, nothing broke when removing the dependency

PR #640, opened 2021-09-08
Yara 3.9 support

### LINKED ISSUE #634: Outdated dependency: Yara
FACT downloads and builds Yara 3.7.1. This version is outdated, security related bugs (memory leaks, buffer overruns, the like) have since been fixed.

both debian buster and bullseye ship newer versions (4.0.5~something), so I'm not entirely sure why the installer builds an outdated version from source.

Please consider removing the source build in favor of distribution-provided packages. If they are too old, it should be the users concern to provide a sufficiently recent version.

jstucke:
Resolves #634 
- changed installation to install the latest yara version
- fixed incompatibility issues with yara versions >= 3.9 (needs new parameter `-C` when using pre-compiled rules)
- added handling for internal server error when trying to upload a compiled rule file in the binary search UI
- refactoring

:exclamation:  https://github.com/fkie-cad/common_helper_yara/pull/4 needs to be merged first

### DIFF
--- a/src/analysis/YaraPluginBase.py
+++ b/src/analysis/YaraPluginBase.py
@@ -24,9 +24,9 @@ class YaraBasePlugin(AnalysisBasePlugin):
         self.config = config
         self.signature_path = self._get_signature_file(plugin_path) if plugin_path else None
         if self.signature_path and not Path(self.signature_path).exists():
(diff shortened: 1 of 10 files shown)


rhelmke (approved):
This seems to work.

Thank you for the workaround here. As we already discussed, it is really unfortunate that our dependency management keeps getting in the way of things. I wish there would be some elegant and more universal solution for us to put all things together.

PR #644, opened 2021-09-09
update without unpacking

jstucke:
- changed "update analysis" so that unpacking is skipped
    - added entry to `about.html`
- fixed "current analysis progress" in "system health" so that progress is displayed correctly for updates
- fixed a bug where certain combinations of unpacking and analysis progress in combination with minimum length of the progress text were causing the progress bar lengths to be displayed wrongly
- refactoring (f-string + pylint fixes)

### DIFF
--- a/src/web_interface/components/analysis_routes.py
+++ b/src/web_interface/components/analysis_routes.py
@@ -169,13 +169,13 @@ class AnalysisRoutes(ComponentBase):
 
     def _schedule_re_analysis_task(self, uid, analysis_task, re_do):
         if re_do:
-            olf_fw_obj = None
+            base_fw = None
             with ConnectTo(AdminDbInterface, self._config) as sc:
                 sc.delete_firmware(uid, delete_root_file=False)
         else:
             with ConnectTo(FrontEndDbInterface, self._config) as db:
-                olf_fw_obj = db.get_firmware(uid)
-        fw = convert_analysis_task_to_fw_obj(analysis_task, base_fw=olf_fw_obj)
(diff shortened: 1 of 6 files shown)


src/web_interface/components/analysis_routes.py
```diff
@@ -167,12 +167,16 @@ def post_update_analysis(self, uid, re_do=False):
...
+            olf_fw_obj = None
             with ConnectTo(AdminDbInterface, self._config) as sc:
                 sc.delete_firmware(uid, delete_root_file=False)
+        else:
+            with ConnectTo(FrontEndDbInterface, self._config) as db:
+                olf_fw_obj = db.get_firmware(uid)
+        fw = convert_analysis_task_to_fw_obj(analysis_task, base_fw=olf_fw_obj)
         with ConnectTo(InterComFrontEndBinding, self._config) as sc:
-            sc.add_re_analyze_task(fw)
+            sc.add_re_analyze_task(fw, unpack=re_do)
```
rhelmke, line 179: Reading the code, I do not quite understand what the `olf` in `olf_fw_obj` means here. Did you mean `old`? Maybe `base_fw` would be more suitable.

rhelmke (changes requested):

src/web_interface/components/analysis_routes.py
jstucke: Yes, that was a typo :rofl: -> fixed

rhelmke (approved):
This works as expected. Thanks!

PR #645, opened 2021-09-10
new plugin for hardware analysis

Mic27M:


src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,115 @@
...
+        if isinstance(kernel_config, str):
+
+            kernel_config_list = kernel_config.splitlines()
+
+            kernel_config = ""
+
+            for line in kernel_config_list:
+
+                if not line.startswith("#"):
+                    kernel_config = kernel_config + line + "\n"
```
jstucke, line 75: we agreed internally to use f-strings (and not the "+" operator) for string formatting
also this could be achieved with a `join()`:
```suggestion
            kernel_config = "\n".join([
                line
                for line in kernel_config_list
                if not line.startswith("#")
            ])
```

But I think it would be better to do the `join()` in the template because it makes using the output from this plugin easier (it would be a JSON array instead of one long string)

There should probably be more filtering, though because many of the kernel config fields are not related to architecture/hardware. Maybe a whitelist would make sense (it doesn't have to be complete at this point). Otherwise, a blacklist filtering out all known standard options could make sense so that we also get all relevant options that we don't know yet (but also false positives).

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,115 @@
...
+        # init the data
+
+        cpu_architecture_dict = file_object.processed_analysis['cpu_architecture']
+        kernel_config_dict = file_object.processed_analysis['kernel_config']
+
+        # getting the cpu architecture (test needed?)
+        cpu_architecture = list(cpu_architecture_dict.keys())[0]
+
+        if cpu_architecture == "summary":
+            cpu_architecture = "no information"
```
jstucke, line 34: it would be better to store `None` here and display `"no information"` in the template

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,115 @@
...
+                file_object.processed_analysis[self.NAME]['summary'] = ['']
+
+            elif (elf_analysis != "no information") & (kernel_config != "no information"):
+                file_object.processed_analysis[self.NAME]['summary'] = ['.modinfo available - kernel configuration available']
+
+            elif (elf_analysis != "no information") & (kernel_config == "no information"):
+                file_object.processed_analysis[self.NAME]['summary'] = ['.modinfo available']
+
+            elif (elf_analysis == "no information") & (kernel_config != "no information"):
+                file_object.processed_analysis[self.NAME]['summary'] = ['kernel configuration available']
```
jstucke, line 113: The summary is a list of strings and we can simply add more entries where it makes sense. In this case it would probably be a good idea to add summary items for each category, i.e.
```suggestion
        if cpu_architecture != "no information":
            file_object.processed_analysis[self.NAME]['summary'].append(f'{cpu_architecture}')
        if elf_analysis != "no information":
            file_object.processed_analysis[self.NAME]['summary'].append('modinfo')
        if kernel_config != "no information":
            file_object.processed_analysis[self.NAME]['summary'].append('kernel configuration')
```

jstucke: On a side note: `&` is a binary logic operator and should only be used for binary logic operations. Instead you could simply use the "regular" logic operator  `and` (the same goes for `|` and `or`)

jstucke (commented):
There are several files in the PR that were probably added by mistake

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+        file_object.processed_analysis[self.NAME]['kernel configuration'] = kernel_config
+
+        # propagate summary to parent objects
+        file_object.processed_analysis[self.NAME]['summary'] = self.make_summary(cpu_architecture, modinfo, kernel_config)
+
+        return file_object
+
+    def cpu_architecture_analysis(self, file_object) -> Optional[str]:
+        cpu_architecture_dict = file_object.processed_analysis['cpu_architecture']
+        cpu_architecture = cpu_architecture_dict["summary"]
```
jstucke, line 41:
```suggestion
        cpu_architecture = file_object.processed_analysis['cpu_architecture']['summary']
```

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if cpu_architecture == []:
+            cpu_architecture = None
+        else:
+            cpu_architecture = cpu_architecture[0]
+
+        return cpu_architecture
```
jstucke, line 48: This can be simplified:
```suggestion
        if cpu_architecture == []:
            return None
        return cpu_architecture[0]
```

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+        # search for important information
+        cpu_architecture = self.cpu_architecture_analysis(file_object)
+        modinfo = self.filter_modinfo(file_object)
+        kernel_config = self.filter_kernel_config(file_object)
+
+        # store the results
+        file_object.processed_analysis[self.NAME] = dict()
+        file_object.processed_analysis[self.NAME]['cpu architecture'] = cpu_architecture
+        file_object.processed_analysis[self.NAME]['*.ko ".modinfo"'] = modinfo
+        file_object.processed_analysis[self.NAME]['kernel configuration'] = kernel_config
```
jstucke, line 32: We can clear this up a bit:
```suggestion
        file_object.processed_analysis[self.NAME] = {
            'cpu architecture': cpu_architecture,
            'modinfo section': modinfo,
            'kernel configuration': kernel_config
        }
```
Also `'*.ko ".modinfo"'` doesn't seem to be very practical for accessing the data so maybe something like "modinfo section" would be better

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+        if cpu_architecture == []:
+            cpu_architecture = None
+        else:
+            cpu_architecture = cpu_architecture[0]
+
+        return cpu_architecture
+
+    def filter_modinfo(self, file_object):
+        # getting the information from the *.ko files .modinfo
+        if "modinfo" in file_object.processed_analysis['elf_analysis'].keys():
```
jstucke, line 52: We internally agreed to only use single quoted strings in the code (there are some more)
```suggestion
        if 'modinfo' in file_object.processed_analysis['elf_analysis'].keys():
```

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+            modinfo = file_object.processed_analysis['elf_analysis']['modinfo']
+        else:
+            modinfo = None
+
+        return modinfo
+
+    def filter_kernel_config(self, file_object):
```
jstucke, line 59: We should filter out kernel config entries that are irrelevant (from hardware perspective). Displaying all entries is not very useful. This could be accomplished either by filtering out entries that are known to be irrelevant (aka "blacklisting") or keeping entries that are known to be useful (aka "whitelisting"). The former is prone to false positives and the latter is prone to false negatives. I guess the former makes more sense in this case, because we can't really know all relevant entries but there are a lot of default entries, that are not very informative.

src/plugins/analysis/hardware_analysis/code/hardware_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+        kernel_config_dict = file_object.processed_analysis['kernel_config']
+        kernel_config = kernel_config_dict.get("kernel_config")
+
+        if isinstance(kernel_config, str):
+            kernel_config_list = kernel_config.splitlines()
+            kernel_config = "\n".join([
```
jstucke, line 65: It would probably make sense to do the `join()` in the template, because if we want to access and use the data for something else, it would be better to have it as a list and not as a long string with line breaks.

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -142,3 +146,22 @@ def _convert_address_values_to_hex(elf_dict):
             for entry in elf_dict[category]:
                 for key in {'virtual_address', 'offset'}.intersection(entry):
                     entry[key] = hex(entry[key])
+
+    def filter_modinfo(self, binary) -> Optional[str]:
```
jstucke, line 150: The type hint doesn't fit the changed code:
```suggestion
    def filter_modinfo(self, binary) -> Optional[List[str]]:
```

src/plugins/analysis/hardware_analysis/view/hardware_analysis.html
```diff
@@ -0,0 +1,22 @@
...
+	<td class="table-head-light"><b>Plug-in</b></td>
+	<td class="table-head-light"><b>Detection</b></td>
+</tr>
+	{% for key, value in firmware.processed_analysis[selected_analysis].items() %}
+		{% if value %}
+			{% if key | is_not_mandatory_analysis_entry %}
+				<tr>
+					<td> {{ key }} </td>
+					{% set entry = "\n".join(value) if value is iterable else value %}
+					<td><pre><code style="word-break: break-all; white-space: pre-wrap;">{{ firmware.processed_analysis[selected_analysis][key] | nice_generic | safe }}</code></pre></td>
```
jstucke, line 16: The `join()` isn't really necessary, since the `nice_generic` filter seems to handle lists just fine, so we could remove it
```suggestion
					<td><pre><code style="word-break: break-all; white-space: pre-wrap;">{{ firmware.processed_analysis[selected_analysis][key] | nice_generic | safe }}</code></pre></td>
```

jstucke:
looks like the new elf_analysis test failed:
> 13:47:04 >           for section in binary.sections:
13:47:04 E           AttributeError: 'LiefResult' object has no attribute 'sections'
13:47:04 
13:47:04 src/plugins/analysis/elf_analysis/code/elf_analysis.py:154: AttributeError

PR #650, opened 2021-09-20
fixed incompatibility with newest highlight.js version

### LINKED ISSUE #649: highlight.js broken in newest version
The current version of highlight.js does not contain `highlight.pack.js` (which is imported in `show_analysis.html`)

> GEThttp://localhost:5000/static/highlight.js/highlight.pack.js
> [HTTP/1.1 404 NOT FOUND 4ms]
> Loading failed for the <script> with source “http://localhost:5000/static/highlight.js/highlight.pack.js”.

What is more, there are deprecation warnings if the library path is changed to `highlight.min.js` and the stylesheet path is also not correct:

> Deprecated as of 10.7.0. highlightBlock will be removed entirely in v12.0 highlight.min.js:105:26
> Deprecated as of 10.7.0. Please use highlightElement now.

jstucke:
Resolves #649 

- fixes library and stylesheet paths
- replaces deprecated function

### DIFF
--- a/src/web_interface/templates/show_analysis.html
+++ b/src/web_interface/templates/show_analysis.html
@@ -24,8 +24,8 @@
     <script type="text/javascript" src="{{ url_for('static', filename='web_js/jstree/dist/jstree.min.js') }}"></script>
 
     {# highlight.js import #}
-    <script src="{{ url_for('static', filename='highlight.js/highlight.pack.js') }}"></script>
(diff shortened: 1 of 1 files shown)


rhelmke (approved):
Just to be safe I grepped our code base against the list of [breaking changes](https://highlightjs.org/#news-142) in highlight.js 11.0. This PR appears to be complete in its intent :-)

PR #651, opened 2021-09-21
Add "Browse Binary Search History" Page

maringuu:


src/web_interface/components/database_routes.py
```diff
@@ -73,6 +74,33 @@ def browse_database(self, query: str = '{}', only_firmwares=False, inverted=Fals
...
 
+    @roles_accepted(*PRIVILEGES['pattern_search'])
+    @AppRoute('/database/browse_binary_search_history', GET)
+    def browse_searches(self):
+        page, per_page, offset = extract_pagination_from_request(request, self._config)
+        all_searches = []
+
+        try:
+            with ConnectTo(FrontEndDbInterface, self._config) as conn:
+                # FIXME Use a proper yara parser
```
jstucke, line 85: Do you think it would make sense to use something like [plyara](https://pypi.org/project/plyara/)?

jstucke (commented):
Seems to work fine

jstucke:
Also we should add one or more tests for the new page

src/web_interface/components/database_routes.py
```diff
@@ -73,6 +74,33 @@ def browse_database(self, query: str = '{}', only_firmwares=False, inverted=Fals
...
+                rule_name_regex = re.compile(r'rule\s+([[a-zA-Z_]\w*)')
+                all_searches = [(r['_id'], r['query_title'], rule_name_regex.findall(r['query_title']))
+                    for r in conn.search_query_cache.find(skip=per_page * (page - 1), limit=per_page)]
+                total = conn.search_query_cache.find().count()
+        except Exception as exception:
+            error_message = 'Could not query database: {} {}'.format(type(exception), str(exception))
+            logging.error(error_message)
```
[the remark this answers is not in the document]
maringuu: Copy paste error :)

src/web_interface/components/database_routes.py
maringuu: Good find. Somehow it does not work for me. It gives some internal error from ply. So I think it would be better to do this somewhen in the future.

PR #652, opened 2021-09-21
pre-commit: Add double-quote-string-fixer

maringuu:
Maybe it is just me but out of habit I sometimes still use double quotes instead of single quotes.
This enforces single quotes.


jstucke (approved):
Nice find! This should be pretty useful and save us some time

PR #655, opened 2021-09-28
Refactor linter plugin, use eslint

### LINKED ISSUE #639: Suggestion: replace jshint with eslint
The linter plugin uses jshint as a javascript linter. While there is nothing inherently wrong with that, eslint is the default for many modern web projects. For example, JSHint cannot parse reacts .jsx files, and most modern web frameworks like react, vue or angular are only really well supported with rules for eslint.

So, I think the quality of the linter output could be improved by switching linter.

maringuu:
When we depend on newer Ubuntu versions we might include the [eslint security plugin](https://github.com/nodesecurity/eslint-plugin-security).

Closes #639


src/plugins/analysis/linter/internal/js_linter.py
```diff
@@ -1,46 +1,35 @@
...
-    Wrapper for jshint javascript linter
+    Wrapper for eslint javascript linter
     '''
+
     def do_analysis(self, file_path):
-        linter_output = execute_shell_command('jshint --config={} --verbose {}'.format(CONFIG_FILE_PATH, file_path))
-        return self._parse_linter_output(linter_output)
+        # The linter will have nonzero returncode when a rule matches
+        # pylint: disable=subprocess-run-check
+        output_raw = subprocess.run(
```
maringuu, line 17: Am I mistaken or is `helperFunctions.docker.run_docker_container` only capable of mounting one volume?

src/plugins/analysis/linter/code/source_code_analysis.py
```diff
@@ -33,59 +29,51 @@ class AnalysisPlugin(AnalysisBasePlugin):
     DEPENDENCIES = ['file_type']
     VERSION = '0.5'
```
jstucke, line 30: We should not forget to update the version (otherwise old analyses will not get updated)
```suggestion
    VERSION = '0.6'
```

src/plugins/analysis/linter/internal/js_linter.py
maringuu: Tbh I think building a whole new docker container just for having files in the right place is a bit too much.

The helper functions does not provide that much benefit, does it? The only benefit I see is that it differentiates between DockerExceptions and timeouts.
`subprocess.run` has timeouts as well. What is does not have is DockerExcpetions.

If you would be fine with it, I'd perfer leaving it as it is right now.

PR #660, opened 2021-10-15
scheduler/Analysis: Analyses aren't up to date when their dependencie…

maringuu:
…s updated

### DIFF
--- a/src/test/unit/scheduler/test_analysis.py
+++ b/src/test/unit/scheduler/test_analysis.py
@@ -428,9 +428,8 @@ class TestAnalysisShouldReanalyse:
         (20, 10, True)
     ])
     def test_analysis_is_up_to_date(self, plugin_root_date, plugin_dep_date, is_up_to_date, monkeypatch):
-        def _analysis_get_date_mock(plugin_name, uid, backend_db_interface):
-            _ = backend_db_interface
-            _ = uid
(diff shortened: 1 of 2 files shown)


src/test/unit/scheduler/test_analysis.py
```diff
@@ -401,3 +405,44 @@ def test_is_forced_update(self):
...
+        cls.init_patch.stop()
+
+    @pytest.mark.parametrize('plugin_root_date, plugin_dep_date, is_up_to_date', [
+        (10, 20, False),
+        (20, 10, True)
+    ])
+    def test_analysis_is_up_to_date(self, plugin_root_date, plugin_dep_date, is_up_to_date, monkeypatch):
+        def _analysis_get_date_mock(plugin_name, uid, backend_db_interface):
+            _ = backend_db_interface
+            _ = uid
```
jstucke, line 433: Did you do this to shut pylint up? :smile: 
Since the parameters are never used as keyword arguments, it should be fine to ignore them
```suggestion
        def _analysis_get_date_mock(plugin_name, *_):
```
or to use a `#pylint: disable=`

jstucke (commented):
Seems to work fine already. Only small stuff:

maringuu:
Should be good to go now.

jstucke (approved):

PR #686, opened 2021-11-08
fixed error when there are no tags set in rest firmware put

### LINKED ISSUE #684: not handling optional field 'tags' properly through rest api
'tags' is an optional attribute of the firmware model, but if the `tags` attribute is not provided, the backend crashes as there's no default value for `tags`.

The frontend bails here because `tag_string` is None.
https://github.com/fkie-cad/FACT_core/blob/0ab195bfe23c50f2d0a88ad75085d67a467d6f88/src/helperFunctions/mongo_task_conversion.py#L97

This happens through the API `rest/firmware` endpoint.

jstucke:
- fixes a bug when a firmware is uploaded via REST without tags
- pylint fixes

resolves #684

### DIFF
--- a/src/helperFunctions/mongo_task_conversion.py
+++ b/src/helperFunctions/mongo_task_conversion.py
@@ -2,7 +2,7 @@ import logging
 from configparser import ConfigParser
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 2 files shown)


rhelmke (approved):
Approved

PR #689, opened 2021-11-09
adding the Tilera/TILE-Gx architecture for detection

wideglide:
Adding TILE* family of architectures.  Not sure if you have a preference on the arch name since `Tilera` is the company name?


dorpvom (approved):
We'll run CI just in case, but this looks fine.

dorpvom:
Can you provide an example binary? I've not come across this architecture yet.

wideglide:
[libxml.so.gz](https://github.com/fkie-cad/FACT_core/files/7513282/libxml.so.gz)

Here's the libxml library compiled for TILE-Gx.  You can find firmware and more examples on the MicroTik [download](https://mikrotik.com/download) page.

Issue #691, opened 2021-11-12
cwe_checker causes backend offline. 


AlexPeiran:
Hello, Fact is a very convenient automated firmware analysis platform, thanks for your work.
Last month, the plugin cwe_checker worked well and help me a lot.To make sure the plugin is up-to-date, I re-install Fact regularly. Recently I found out that cwe_checker does not work properly and causes backend offline. 

The log shows:
[2021-11-11 11:45:04][cwe_checker][ERROR]: cwe_checker execution failed: Execution of Ghidra plugin failed: Process was terminated.
INFO  REPORT: Save succeeded for file: /input (HeadlessAnalyzer)  
INFO  REPORT: Post-analysis succeeded for file: /input (HeadlessAnalyzer)  
………………

UID: cd5eabea0aa4c0aef0c6cb571233b91f5e6e2f73e126ffa57776ce28518454e8_215240
[2021-11-11 12:58:37][cwe_checker][ERROR]: Timeout or error during cwe_checker execution.
UID: 95d80c685435bcc979c3ef04252fc7c471eee442c305520eede47ee882e8d5ad_2840360
[2021-11-11 13:01:55][docker][WARNING]: [Docker]: encountered process error while processing
[2021-11-11 13:02:06][cwe_checker][ERROR]: Timeout or error during cwe_checker execution.
………………

Should I try to install an older version of cwe_checker?

Enkelmann:
That the cwe_checker plugin throws errors (like those in your log) for some files is normal, but that should not cause problems with the whole backend of FACT. Could you describe what you mean with it causing "backend offline"?

Regarding the version of the cwe_checker: As far as I remember, the standard version of the cwe_checker installed with FACT is its latest stable version (v0.5). At the moment I recommend sticking to that version, since newer dev-versions of the cwe_checker currently have some runtime issues, which causes the cwe_checker-FACT-plugin to run into timeouts quite often.

AlexPeiran:
In the web interface , I can see the status of the "backend" is "offline", just in the system dashboard. And the analysis process stopped. I can see only 1 process of cwe-checker is running(sometimes None) after a few time since the analysis started.

When I use the command "docker image list", I can see that cwe-checker has been updated a few days ago（the old one's tag is "<None>", the newer one's tag is “Latest”）. Since then, analysis has been difficult to start.

[...]

PR #692, opened 2021-11-15
switch to the stable cwe_checker image instead of the latest one

Enkelmann:
Switch the Docker image of the cwe_checker plugin to the stable image instead of the latest one. This should hopefully prevent issues like in #691. The downside is that bugfixes to the cwe_checker need longer until they also land in FACT.


jstucke (approved):
Looks good to me

PR #693, opened 2021-11-16
Device tree

Mic27M:
new plugin to use the device-tree-compiler in fact


src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,39 @@
...
+
+from analysis.PluginBase import AnalysisBasePlugin
+
+
+class AnalysisPlugin(AnalysisBasePlugin):
+    '''
+    Device Tree Plug-in
+    '''
+    NAME = 'device_tree'
+    DESCRIPTION = 'get the device tree in text from from the device tree blob'
```
dorpvom, line 11: typo

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,39 @@
...
+        file_object.processed_analysis[self.NAME]['device_tree'] = device_tree
+
+        file_object.processed_analysis[self.NAME]['summary'] = ['device tree found']
+
+        return file_object
+
+    def execute_device_tree_compiler(self, file_path):
+        dtc_result, return_code = execute_shell_command_get_return_code(f'dtc -I dtb -O dts {file_path}')
+        if return_code != 0:
+            raise ValueError(f'The Device Tree Compiler exited with non-zero return code {return_code} after working on {file_path}')
```
jstucke, line 38: It would be better to only log a warning (and maybe store a result in `file_object.processed_analysis[self.NAME]['warning']`) in case the conversion failed instead of letting the worker process crash

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,39 @@
...
+class AnalysisPlugin(AnalysisBasePlugin):
+    '''
+    Device Tree Plug-in
+    '''
+    NAME = 'device_tree'
+    DESCRIPTION = 'get the device tree in text from from the device tree blob'
+    DEPENDENCIES = []
+    VERSION = '0.1'
+    MIME_WHITELIST = ['linux/device-tree']
+    # TODO installation
```
jstucke, line 15: This comment is not needed anymore

src/plugins/analysis/device_tree/test/test_device_tree.py
```diff
@@ -0,0 +1,37 @@
...
+        test_object.file_path = str(self.test_file)
+        result = self.analysis_plugin.process_object(test_object)
+
+        assert result.processed_analysis[self.PLUGIN_NAME]['summary'] == ['device tree found']
+
+    def test_execute_device_tree_compiler(self):
+        test_object = self.test_file
+        result = self.analysis_plugin.execute_device_tree_compiler(test_object)
+
+        assert '#address-cells = <0x01>;' in result
```
jstucke, line 37: This test fails in the CI. Also, the results also contain a lot of structural warnings. Maybe we can parse the output and store them separately. Also, adding a test for this case would make sense.

src/plugins/analysis/device_tree/test/test_device_tree.py
```diff
@@ -0,0 +1,37 @@
...
+
+from objects.file import FileObject
+from test.unit.analysis.analysis_plugin_test_class import AnalysisPluginTest
+
+from ..code.device_tree import AnalysisPlugin
+
+TEST_DATA = Path(__file__).parent.parent / 'test/data'
+
+
+class test_device_tree(AnalysisPluginTest):
```
jstucke, line 11: class names should be CamelCase©

jstucke (changes requested):

src/plugins/analysis/device_tree/test/test_device_tree.py
```diff
@@ -0,0 +1,36 @@
...
+    def setUp(self):
+        super().setUp()
+        config = self.init_basic_config()
+
+        self.analysis_plugin = AnalysisPlugin(self, config=config)
+        self.test_file = TEST_DATA / 'device_tree.dtb'
+
+    def tearDown(self):
+        super().tearDown()
+        # additional tearDown can go here
```
jstucke, line 23: This function can be removed, since it does nothing more than call the function of the parent class

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,44 @@
...
+        self.config = config
+
+        super().__init__(plugin_administrator, config=config,
+                         recursive=recursive, plugin_path=__file__)
+
+    def process_object(self, file_object):
+
+        device_tree = self.execute_device_tree_compiler(file_object.file_path)
+
+    # store the results
```
jstucke, line 28: The comment should be on the same indentation level as the rest of the function. It can probably be removed, though, since it is pretty obvious that the results are stored below

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,44 @@
...
+
+class AnalysisPlugin(AnalysisBasePlugin):
+    '''
+    Device Tree Plug-in
+    '''
+    NAME = 'device_tree'
+    DESCRIPTION = 'get the device tree in text from the device tree blob'
+    DEPENDENCIES = []
+    VERSION = '0.1'
+    # MIME_WHITELIST = ['linux/device-tree']
```
jstucke, line 16: Why is this commented out?

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -0,0 +1,44 @@
...
+from analysis.PluginBase import AnalysisBasePlugin
+
+
+class AnalysisPlugin(AnalysisBasePlugin):
+    '''
+    Device Tree Plug-in
+    '''
+    NAME = 'device_tree'
+    DESCRIPTION = 'get the device tree in text from the device tree blob'
+    DEPENDENCIES = []
```
jstucke, line 14: If we use a whitelist, the file_type plugin should be added to the dependencies
```suggestion
    DEPENDENCIES = ['file_type']
```

src/plugins/analysis/device_tree/code/device_tree.py
Mic27M: I commented it out for testings, but forgot  to undo it...

jstucke (approved):
LGTM

PR #698, opened 2021-11-27
Update INSTALL.md with WSL specific instructions for enabling dockerd

timkoers:
Added instructions specific for WSL (since systemctl is not available on WSL), that make sure that the docker daemon is actually started


jstucke:
Hi and thank you for your contribution. Does this mean FACT runs on WSL2 apart from this problem? I always wanted to test it but somehow never got around to actually do it.

jstucke (approved):
Looks good to me

PR #699, opened 2021-11-28
Improve Fedora support

juju4:
This should fix most issues on Fedora

* nginx: fix selinux context for etc and log files
* mongodb: use current 4.2
* remove pip-installed setuptools as seems to break selinux tools...

### DIFF
--- a/src/install.py
+++ b/src/install.py
@@ -164,7 +164,7 @@ def install():
 def install_fact_components(args, distribution, none_chosen, skip_docker):
     common(distribution)
     if args.frontend or none_chosen:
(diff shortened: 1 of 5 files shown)


jstucke:
Hi and thank you for your contribution! The problem is that we can't really switch to a newer version of MongoDB because of its problematic license. Since the last version which still uses the AGPL license has reached its end of life, we will probably move away from MongoDB altogether in the long term.

juju4:
no problem. reverted mongodb part

jstucke (approved):
Looks good to me and the CI build passed.

juju4:
Thanks!

PR #711, opened 2021-12-02
Extend device tree plugin

jstucke:
extended device tree plugin with functionality to dump device trees from arbitrary files


src/plugins/mime_blacklists.py
```diff
@@ -0,0 +1,5 @@
+MIME_BLACKLIST_NON_EXECUTABLE = ['audio', 'filesystem', 'image', 'video']
+MIME_BLACKLIST_COMPRESSED = [
+    'application/gzip', 'application/gzip', 'application/x-7z-compressed', 'application/x-tar',
+    'application/x-xz', 'application/zip'
+]
```
rhelmke, line 5: good idea :-)

rhelmke (approved):

PR #716, opened 2021-12-13
updated base image for FSR

jstucke:
- updated the base image of the format string resolver (software_components plugin)
    - the previous base image may be susceptible to the vulnerability in log4j


Enkelmann (approved):
LGTM

PR #720, opened 2021-12-20
ELF Analysis Plugin: Add x-pie-executable

rhelmke:
* Add `x-pie-executable` mimetype to plugin whitelist
* Added corresponding test case
* Robustness: during analysis, a ValueError may occur when unpacking the result of `_analyze_elf`. We didn't catch that
* Bump version to `3.3.0`


jstucke (approved):
Looks good to me

PR #721, opened 2021-12-20
fixed binwalk installation

jstucke:
- binwalk installation with `setup.py` seems to be broken on some systems
    - changed installation to use `pip`
- added error handling to binwalk plugin


rhelmke (approved):
I still don't know why this error appeared out of the blue, but I'm glad you found and fixed it :-)

PR #724, opened 2021-12-21
added test for oms plugin

jstucke:


jstucke:
It seems like this PR will be superseded by #732

jstucke:
superseded by #732

Issue #712, opened 2021-12-03
Make Dependency Graph Searchable


frakman1:
The Dependency graph is nice but of limited use for large products. If I can see the module/library/executable, then I can click on it and follow the links. However with hundreds of circles, there's no way of finding the module I am looking for.

![image](https://user-images.githubusercontent.com/5826484/144525003-5fb33462-e5eb-4c8c-aa97-656ee87b8c5a.png)

Suggestion is to either allow CTL+F style text searching, or add a search widget that will highlight the circles that match the string being searched.

dorpvom:
Cue:

![](https://boyter.org/static/books/a0f278b167ec23c35fe091f38096571a.jpg)

But for real: We'll look into this, but on first glance it seems to be more complicated than adding a few lines of code. But we might be able to add a search form. As non of us is a JS pro it might be a minute until we come around to do this though.

rhelmke:
Hey,

so I am no web-pro, but I made an effort to re-implement the whole dependency graph functionality. You can find the code in pull request #725. It won't be merged before next January as most of us are already in the holidays. But feel free to check it out once it is merged into master.

Cheers

PR #725, opened 2021-12-22
Reimplement dependency graph with tools for navigation

rhelmke:
depgraph_rework
 
Issue #712 made clear that the current dependency graph implementation is of limited use: Usage in larger environments fails due to missing tools for proper navigation. As the client-side code was rather messy and unmaintained, I took this as a chance to completely re-implement the feature with added functionality.

This PR includes following changes:

* Enabled `vis.js` physics using the rather fast [forceAtlas2Based](https://visjs.github.io/vis-network/docs/network/physics.html) solver
* Nodes are no longer draggable (the physics solver will take care of clustering, if even possible in complex dependecy graphs)
* Physics simulation will stop after 60 seconds to save power and CPU cycles (should be sufficient to kind of stabilize the graph)
* Each element in the sidebar is collapsible
* Moved right click modal with node information to a sidebar element
* Added `Connected Nodes` list as sidebar element
* `Connected Nodes` list is dynamically filterable/searchable using regex input
* Nodes in `Connected Nodes` can be used for navigation and further searching within the graph
* Reworked legend, made it a pure html text element in the sidebar
* Added support for `x-pie-executable` ELF files
* Linearly scalable colors for nodes and edges
* Deterministic colors: Colors were switched up between graphs in the previous implementation
* Added camera movements and generic zoom on node selection to visually aid graph navigation

_For a screenshot of the new dependency graph see above_

### DIFF
--- a/src/test/unit/web_interface/test_dependency_graph.py
+++ b/src/test/unit/web_interface/test_dependency_graph.py
@@ -67,8 +67,8 @@ GRAPH_RES_SYMLINK = {'nodes':
                        'full_file_type': 'test text'},
                       {'label': 'file three', 'id': '0987654', 'group': 'inode/symlink',
                        'full_file_type': 'symbolic link to \'file two\''}],
-                      'edges': [{'from': '0987654', 'to': '7654321', 'id': 0},
-                                {'from': '7654321', 'to': '1234567', 'id': 1}],
+                     'edges': [{'from': '0987654', 'to': '7654321', 'id': 0},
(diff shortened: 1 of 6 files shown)


jstucke (approved):
I think this feature is much more useful now. Seems to work great. I did not review ALL the JavaScript code, though.

While testing this I noticed one thing that could be optimized: The symlinks that are not connected to ELF files (or maybe only the `.properties` links) should probably be filtered out. This isn't really related to the PR, though. 

Sometimes most symlinks are not rendered at all. This is a bug also not related to this PR and #726 addresses this problem.

PR #726, opened 2022-01-03
File object update parents bugfix

jstucke:
`parents` field was missing from `FileObject` update method.
This results in the `parents` being overwritten each time the object is updated.
This PR fixes this problem by merging the old parents and new parents instead (similar to the "parent firmware UIDs").
Added a test to make sure this works as intended.

### DIFF
--- a/src/helperFunctions/merge_generators.py
+++ b/src/helperFunctions/merge_generators.py
@@ -1,9 +1,10 @@
 import itertools
 from copy import deepcopy
 from random import sample, seed
-from typing import Sequence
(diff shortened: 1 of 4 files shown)


rhelmke (changes requested):
So after further troubleshooting this bugfix in combination with the new dependency graph, this looks fine except for ambiguity tagged below.

However, this PR does actually NOT fix the issue of missing symlinks in the dependency graph. As the culprit of all evil appears to lay elsewhere. We're going to address this in a separate PR

rhelmke (approved):
Looks good to me, thanks!

PR #729, opened 2022-01-26
Init systems single file analysis bugfix

jstucke:
Starting a single-file analysis with plugin "init_systems" led to an error, because `file_object.root_uid` was missing in this context.

- fixed missing `root_uid` bug
- pylint fixes
- added test


rhelmke (approved):
Yup, works for me :-) Thanks!

Once all CI jobs succeed, we can surely merge

PR #730, opened 2022-01-28
Remove common_helper_process

maringuu:
The module does not provide much benefit over just using the subprocess module.


src/compile_yara_signatures.py
```diff
@@ -18,12 +18,12 @@
 '''
 
 import os
+import subprocess
 import sys
-from subprocess import CalledProcessError
+from subprocess import PIPE, STDOUT, CalledProcessError
```
jstucke, line 23: Was importing `PIPE, STDOUT, CalledProcessError` directly but calling run as `subprocess.run` intentional?

src/helperFunctions/install.py
```diff
@@ -179,8 +176,8 @@ def install_github_project(project_path: str, commands: List[str]):
 
 def _checkout_github_project(github_path: str, folder_name: str):
     clone_url = 'https://www.github.com/{}'.format(github_path)
-    _, return_code = execute_shell_command_get_return_code('git clone {}'.format(clone_url))
-    if return_code != 0:
+    git_p = subprocess.run('git clone {}'.format(clone_url), shell=True, stdout=DEVNULL, stderr=DEVNULL, text=True)
+    if git_p != 0:
```
jstucke, line 180: It seems `.returncode` is missing:
```suggestion
    if git_p.returncode != 0:
```

src/helperFunctions/install.py
```diff
@@ -70,12 +69,12 @@ def log_current_packages(packages: Tuple[str], install: bool = True):
 
 
 def _run_shell_command_raise_on_return_code(command: str, error: str, add_output_on_error=False) -> str:  # pylint: disable=invalid-name
-    output, return_code = execute_shell_command_get_return_code(command)
-    if return_code != 0:
+    cmd_p = subprocess.run(command, shell=True, stdout=PIPE, stderr=STDOUT, text=True)
```
jstucke, line 72: I would prefer better readable variable names like simply "process" instead of "cmd_p" (I know what that stands for but someone else who looks at the file at some point in the future maybe doesn't)

src/compile_yara_signatures.py
```diff
@@ -48,7 +48,7 @@ def _create_compiled_signature_file(directory, tmp_file):
     target_path = os.path.join(SIGNATURE_DIR, '{}.yc'.format(_get_plugin_name(directory)))
     try:
         command = 'yarac -d test_flag=false {} {}'.format(tmp_file.name, target_path)
-        execute_shell_command(command, check=True)
+        subprocess.run(command, stdout=PIPE, stderr=STDOUT, shell=True, check=True)
```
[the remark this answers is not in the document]
jstucke: It could also make sense to replace some other calls where we a primarily interested in the output with `subprocess.check_output`, because it is usually shorter.

src/compile_yara_signatures.py
maringuu: Yes, I think its more readable to use `subprocess.run` instead of `run` only.

src/compile_yara_signatures.py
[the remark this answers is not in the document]
maringuu: `subprocess.{check_call,check_output}` are listed as `Older high-level API` in the docs.
Should we still use them?

maringuu:
The CI should work now.
The tests in `src/plugins/analysis/device_tree/test/test_device_tree.py` also fail on master for me.

src/compile_yara_signatures.py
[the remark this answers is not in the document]
jstucke: I did not notice that :sweat_smile: 
If that is the case, then it makes sense to only use `run` (though I doubt that the functions will be deprecated anytime soon)

jstucke:
It seems tests in `test_shell_linter.py` also failed because the mock method is missing arguments (i.e. `**_`)
The device_tree test seems to be failing because the method `dump_device_tree` now returns a byte string instead of a string (because `subprocess.run().stdout` unlike `execute_shell_command_get_return_code()` has type `bytes` and not `str`)

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -112,9 +112,16 @@ def crack_hash(passwd_entry: bytes, result_entry: dict, format_term: str = '') -
...
+        john_process = subprocess.run(
+            f'{JOHN_PATH} --wordlist={WORDLIST_PATH} {fp.name} {format_term}',
+            shell=True,
+            stdout=PIPE,
+            stderr=STDOUT,
+            universal_newlines=True,
+        )
+        result_entry['log'] = john_process.stdout
+        john_process = subprocess.run(f'{JOHN_PATH} {fp.name} --show {format_term}', shell=True, stdout=PIPE, stderr=STDOUT, universal_newlines=True)
+        output = john_process.stdout.split('\n')
```
jstucke, line 124: again I'm not 100% sure this will not trigger a `UnicodeDecodeError` and it would be safer to do the `decode()` manually with `errors='replace'`

src/plugins/analysis/users_and_passwords/install.py
```diff
@@ -26,11 +26,11 @@ class UsersAndPasswordsInstaller(AbstractPluginInstaller):
...
         super().install_system_packages()
 
-        lshw_p = subprocess.run('lshw -c display'.split(), stdout=PIPE, stderr=PIPE, check=True)
+        lshw_process = subprocess.run('lshw -c display'.split(), stdout=PIPE, stderr=PIPE, check=True)
         opencl_pkgs = []
-        if lshw_p.stdout == 'NVIDIA':
+        if lshw_process.stdout == 'NVIDIA':
             opencl_pkgs = ['nvidia-opencl-dev']
-        elif lshw_p.stdout == 'AMD':
+        elif lshw_process.stdout == 'AMD':
```
jstucke, line 33: It seems this did not work in the past because it should either also use a `universal_newlines=True` or compare the `stdout` with byte strings

src/helperFunctions/yara_binary_search.py
```diff
@@ -37,7 +37,8 @@ def _execute_yara_search(self, rule_file_path: str, target_path: Optional[str] =
         '''
         compiled_flag = '-C' if Path(rule_file_path).read_bytes().startswith(b'YARA') else ''
         command = f'yara -r {compiled_flag} {rule_file_path} {target_path or self.db_path}'
-        return execute_shell_command(command)
+        yara_process = subprocess.run(command, shell=True, stdout=PIPE, stderr=STDOUT, universal_newlines=True)
+        return yara_process.stdout
```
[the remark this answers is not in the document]
jstucke: I looked at the code again and since YARA is not called with `-s` this should be fine

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
jstucke: I reconsidered this and since the passwords that are in the word list probably won't contain undecodable password strings, this also should not be an issue,

jstucke (approved):
The merge conflict needs to be resolved

PR #731, opened 2022-01-31
temporary fix for outdated python in bionic

jstucke:
This is a temporary fix for the pip installation aimed solely at Ubuntu bionic (which is seemingly stuck with Python 3.6 until it reaches End of Life next year).


rhelmke (approved):
Great! Thank you very much :-) Merging..

PR #733, opened 2022-02-01
Threaded qemu exec plugin

jstucke:
- replaced `multiprocessing.pool` with more lightweight `ThreadPoolExecutor`
- fixed missing fallback method from `BinaryService`
- added timeout to occasionally hanging tests

### DIFF
--- a/src/install/requirements_common.txt
+++ b/src/install/requirements_common.txt
@@ -11,6 +11,7 @@ psutil
 pylint
 pytest
 pytest-cov
+pytest-timeout
 python-magic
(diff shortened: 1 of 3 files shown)


rhelmke (approved):
1. Using timeouts for the touched tests is reasonable.
2. The previous runs with the broken focal worker succeeded and the memory footprint was way lower.

This looks promising, thank you very much! I started the build process for all other workers. Once they finish successfully, we can merge.

However, we should ask @maringuu to synchronize his dev branches with the master.

PR #739, opened 2022-02-08
added back loading gif that was erroneously removed

jstucke:
- call to function `showImg()` was still present in three places but did no longer exist
- PR removes `showImg()` from "binary search" (has its own loading page) and restores it for upload and basic search


rhelmke:
So the CI workers failed due to the missing fix of PR #740. Since that one just hit master, I re-synced with that branch and am restarting the CI workers now. Lets see :-)

rhelmke (changes requested):
Looks good so far :-) Maybe we can add that change suggestion in `loading.js`?

PR #742, opened 2022-02-18
fixed error in show_analysis caused by missing root_uid

jstucke:
Fixed bug in on "show analysis" page caused by missing root UID (can happen e.g. when coming from a DB search)

```python
{% block body %}{% endblock %}
File "<template>", line 67, in block 'body'
TypeError: can only concatenate str (not "NoneType") to str
```


rhelmke (approved):
Sorry :see_no_evil:

PR #743, opened 2022-02-23
Device tree filter

jstucke:
- added new filter for hiding large binary blobs inside device trees from the analysis view
- added view template for device tree plugin
- refactoring


rhelmke (approved):
Looks way better to me now. Thanks!

PR #744, opened 2022-02-25
updated device tree plugin to support multiple device trees

jstucke:
- used "device tree scanner" internals to replace the tool `fdtdump` for dumping device trees (the tool described itself as "not stable" and could not extract multiple device trees)
- changed result format to support an arbitrary number of device trees in the result (sometimes multiple device trees are contained in a file, sometimes even inside other device trees)
- added metadata to result and view
- added code highlighting and made the results view collapsible


src/plugins/analysis/device_tree/view/device_tree.html
```diff
@@ -1,22 +1,83 @@
...
-            </td>
-        </tr>
-    {% endif %}
+{# Don't break on legacy result #}
+{% if "device_tree" in firmware.processed_analysis[selected_analysis] %}
+    <tr class="table-warning">
+        <td> Deprecation Warning </td>
+        <td> Analysis version is deprecated, please update the analysis </td>
+    </tr>
+    <tr>
```
rhelmke, line 50: I like that :-)

src/plugins/analysis/device_tree/view/device_tree.html
```diff
@@ -1,22 +1,83 @@
...
+                    <tr>
+                        {{ device_tree_table_cell(dt_dict.device_tree, "3") }}
+                    </tr>
+                </table>
+            </div>
+        </td>
+    </tr>
 {% endfor %}
 
 {% endblock %}
```
rhelmke, line 83: That looks really cool! And it appears to work as expected. Thank you very much.

I understand that the summary table is part of the base plugin.. but could we somehow switch it off? I mean it is kind of obsolete because of the collapsible. Is that easily possible? What do you think? Maybe we can discuss about that next monday :-)

rhelmke (changes requested):

src/plugins/analysis/device_tree/view/device_tree.html
rhelmke: For the record: we just talked about that one. The summary does only exist on Firmware-Level. This was basically my fault :-)

rhelmke (approved):

PR #746, opened 2022-03-03
Add tag searching to basic search

lwilms:


src/web_interface/templates/database/database_search.html
```diff
@@ -2,43 +2,50 @@
...
-                ("File Name", "file_name"),
-                ("Device Name", "device_name"),
-                ("Firmware Version", "version"),
-                ("Release Date", "release_date"),
-                ("Hash", "hash_value")
+            ("File Name", "file_name"),
+            ("Device Name", "device_name"),
+            ("Firmware Version", "version"),
+            ("Release Date", "release_date"),
+            ("Hash", "hash_value")
```
jstucke, line 48: Why did you unindent this part?

jstucke (changes requested):
Code looks fine. Web interface looks nice and seems to work but there is a problem with the queries that are generated. The tests still need to be fixed, though.

src/web_interface/components/database_routes.py
```diff
@@ -152,18 +151,17 @@ def _build_search_query(self):
...
             self._add_tag_query(query)
-        print(query)
         return json.dumps(query)
 
     @staticmethod
     def _add_multiple_choice(query, key):
-        new_query = [{key: name} for name in dict(request.form.lists())[key]]
-        query.setdefault('$or', []).extend(new_query)
+        for name in dict(request.form.lists())[key]:
+            query[key] = {"$in": [name]}
```
jstucke, line 159: This would replace the key in each iteration of the for loop.
Instead, it would have to be something like this:
```suggestion
        query[key] = {"$in": [name for name in dict(request.form.lists())[key]]}
```

src/web_interface/components/database_routes.py
```diff
@@ -152,18 +151,17 @@ def _build_search_query(self):
...
 
     @staticmethod
     def _add_tag_query(query):
-        tag_query = [{f'tags.{tag}': 'secondary' for tag in dict(request.form.lists())['tags']}]
-        query.setdefault('$or', []).extend(tag_query)
+        for tag in dict(request.form.lists())['tags']:
+            query[f'tags.{tag}'] = {"$in": ['secondary']}
```
jstucke, line 164: The "$in" query doesn't really work here because the values ("secondary") are the same but the keys are different. If should work if we simply leave it as it was if it is the only "$or" query, though.

src/web_interface/components/database_routes.py
```diff
@@ -142,15 +141,27 @@ def _search_database(self, query, skip=0, limit=0, only_firmwares=False, inverte
...
         return json.dumps(query)
 
+    @staticmethod
+    def _add_multiple_choice(query, key):
+        query[key] = {"$in": [name for name in dict(request.form.lists())[key]]}
+
+    @staticmethod
+    def _add_tag_query(query):
+        tag_query = [{f'tags.{tag}': 'secondary' for tag in dict(request.form.lists())['tags']}]
+        query.setdefault('$or', []).extend(tag_query)
```
jstucke, line 163: I only just noticed that this will also not do exactly what we want: this matches only firmwares that have all selected tags, but it would intuitively make more sense if it matched any firmware with at least one of the tags. Accomplishing this isn't very complicated, though. This should do it:
```suggestion
        tag_queries = [{f'tags.{tag}': 'secondary'} for tag in dict(request.form.lists())['tags']]
        query.setdefault('$or', []).extend(tag_queries)
```

jstucke (changes requested):
Apart from what happens if you search for multiple tags, this seems to work fine. One test currently fails on the CI, though: `TestAcceptanceNormalSearch.test_search`

src/web_interface/templates/database/database_search.html
```diff
@@ -12,26 +14,26 @@
 <div class="row justify-content-center">
 
     <div class="col-xl-6 col-lg-8 col-md-10 mt-4">
+        <script type="text/javascript"
+                src="{{ url_for('static', filename='web_js/bootstrap-select.min.js') }}`"></script>
```
rhelmke, line 18: This template imports `bootstrap-select.min.js` twice. Is there a reason for that?

src/web_interface/templates/database/database_search.html
```diff
@@ -51,6 +53,16 @@ <h3 class="mb-3">Search Firmware Database</h3>
...
+                    {% for tag in tag_list | sort %}
+                    <option value='{{ tag }}'>{{ tag }}</option>
+                    {% endfor %}
+                </select>
+            </div>
+
+
             <button type="submit" value=submit class="btn btn-primary" id="input_submit" onclick='showImg()'>
                 <i class="fas fa-search"></i> Search
             </button>
```
rhelmke, line 68: There is mixed quote usage in the html tag attributes, e.g.:

```html
<button type="submit" value=submit class="btn btn-primary" id="input_submit" onclick='showImg()'>
```

Please use double quotes here.

rhelmke (changes requested):

src/web_interface/templates/database/database_search.html
jstucke: removed

src/web_interface/templates/database/database_search.html
jstucke: fixed

rhelmke (approved):
Thank you both! Works fine for me.

Cherry on top of the cake would be variadic logical tag composition. Switching between `and` and `or` using a checkbox or button would be cool! But maybe that is more of a feature request addressable by another PR.

rhelmke (approved):

PR #756, opened 2022-03-29
fixed bad requests interface tests

rhelmke:
Flask's behavior changed between 2.0.2 and 2.0.3 with regards to requests containing unexpected Content-Types.
This causes FACT's API to answer with

```json
{
    "message": "The browser (or proxy) sent a request that this server could not understand."
}
```

instead of

```json
{
    "message": "Input payload validation failed"
}
```

However, some of our unit tests check only for the latter. This causes tests to fail when `flask>2.0.2` is installed. I suggest to check for `response.status_code == 400` instead of the response's message content. This is more robust across flask versions (and, btw, fixes our CI ;-)

### DIFF
--- a/src/test/unit/web_interface/rest/test_rest_firmware.py
+++ b/src/test/unit/web_interface/rest/test_rest_firmware.py
@@ -2,6 +2,7 @@ import json
 from base64 import standard_b64encode
 from urllib.parse import quote
 
+import logging
 from test.common_helper import TEST_FW
 from test.unit.web_interface.rest.conftest import decode_response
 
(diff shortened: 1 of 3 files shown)


src/test/unit/web_interface/rest/test_rest_firmware.py
```diff
@@ -68,13 +69,13 @@ def test_successful_uid_request(test_app):
...
-    assert 'Input payload validation failed' in result['message']
+    response = test_app.put('/rest/firmware')
+    assert response.status_code == 400
 
 
-def test_submit_empty_data(test_app):
-    result = decode_response(test_app.put('/rest/firmware', data=json.dumps({})))
-    assert 'Input payload validation failed' in result['message']
+def test_submit_non_json_data(test_app):
+    response = test_app.put('/rest/firmware', data='non-json-string')
```
jstucke, line 77: I think the intention behind this test was to check that the input validation assures that a request that does not contain the required fields fails. This seems to test something different (since the input data isn't even valid JSON). What was the idea behind this change? Maybe we should test both.

rhelmke: Oh okay, I understand! The issue here was that the `data` kwarg sets `Content-Type: application/x-www-form-urlencoded`. The latest versions of flask and flask-restx filter this because `application/json` is expected. In previous versions, this was no issue because the content was passed to our endpoint handler either way.

So in order to test valid json but unfulfilled scheme requirements, we should do something like:

```python
def test_submit_empty_data(test_app):
    result = decode_response(test_app.put('/rest/firmware', json={}))
    assert 'Input payload validation failed' in result['message']
```

Generally, this should also be backward-compatible to older package versions. Yet I didn't test an empty dict here, maybe we need some bogus key-value set in order to actually send content. But that would be fine, right?

jstucke: That looks good to me :+1:

rhelmke: done. starting CI

jstucke (approved):
Looks good to me

PR #761, opened 2022-04-05
fix for incompatibility with new prompt-toolkit version

jstucke:
fixed an incompatibility with the new version of prompt-toolkit


rhelmke (approved):
Thanks for the fix!

PR #764, opened 2022-04-07
hide loading gif when going back in page history

jstucke:
There was a problem with the loading gif covering the basic search and upload form when going back to the page (e.g. by clicking on the "go back" button in the browser). This should fix this problem.


rhelmke (approved):
Works for me in Chromium, Brave, and Firefox. Could not verify Internet Explorer, tho ;-)

PR #767, opened 2022-04-29
Graceful start fail

jstucke:
- fail gracefully when:
    - FACT is not installed
    - FACT was upgraded to v4 but not reinstalled
- added docs for:
    - migration from v3 to v4
    - missing helper functions

### DIFF
--- a/src/fact_base.py
+++ b/src/fact_base.py
@@ -9,9 +9,8 @@ try:
 
     from helperFunctions.program_setup import program_setup
     from statistic.work_load import WorkLoadStatistic
-except ImportError:
-    import sys
-    logging.error(
+except (ImportError, ModuleNotFoundError):
+    logging.exception(
         'Could not load dependencies. Please make sure that you have installed FACT correctly '
         '(see INSTALL.md for more information). If you recently updated FACT, you may want to rerun the installation.'
     )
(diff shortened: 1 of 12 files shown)


src/fact_base.py
```diff
@@ -3,10 +3,24 @@
...
+    logging.error(
+        'Could not load dependencies. Please make sure that you have installed FACT correctly '
+        '(see INSTALL.md for more information). If you recently updated FACT, you may want to rerun the installation.'
+    )
+    logging.warning(
+        'The database of FACT switched from MongoDB to PostgreSQL with the release of FACT 4.0. '
+        'For instructions on how to upgrade FACT and how to migrate your database see '
+        'https://fkie-cad.github.io/FACT_core/migration.html'
+    )
+    sys.exit(1)
```
dorpvom, line 23: You should be able to replace this by a raise of the original error. Then you can keep the syntax in the dependent files and loose the unnecessary `sys.exit` here.

dorpvom (changes requested):

src/start_fact_db.py
```diff
@@ -16,11 +16,14 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 '''
-
+import logging
 import sys
 
+from sqlalchemy.exc import SQLAlchemyError
+
 from fact_base import FactBase
```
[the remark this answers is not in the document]
jstucke: changed it to be consistent in all start scripts

src/fact_base.py
jstucke: added the `raise` and changed the `logging.error` to a `logging.exception` so that we still know what caused the error

dorpvom (approved):

PR #770, opened 2022-05-04
DbInterfaceBackend: Don't crash when updating when FileObject is alre…

maringuu:
Assigning `__class__` is not really nice, but since we the other functions of the db interface are allowed to use `isinstance` this is the only thing that comes to my mind.


jstucke:
I don't think this is really a useful use case. Having a file as a firmware and a file object inside another firmware causes all sorts of additional problems (e.g. if the firmware is deleted as firmware, it actually deletes the file even though it is still contained in the other firmware)

Instead of handling errors like this, we should probably throw an error during upload instead. But then we need to consider the opposite case as well (a new file is uploaded that contains a file that is a firmware already in the DB). But these should be corner cases and (hopefully) not occur during normal use.

PR #777, opened 2022-05-09
fixed bug with rubocop linter installation in older systems

jstucke:
`rubocop` linter installation failed for bionic and buster. This PR should fix this.

- switched to rubocop docker image to fix installation problems on older systems
- pylint fixes


rhelmke (approved):
Looks good to me. Linter Plugin still works as expected.

PR #779, opened 2022-05-13
kernel config plugin import bugfix

jstucke:
- fixed an import error occurring in some cases during analysis with the `kernel_config` plugin


rhelmke (approved):
Weird. Not quite sure why this problem occurs in the first place.

But I'm glad you found & fixed it! Thanks.

PR #782, opened 2022-05-16
installer: Allow pulling docker images on unsupported distribution

maringuu:
In FACT_docker I plan to have a seperate container to execute scripts like `init_postgres.py` and install docker images.
This container has no reason not to be alpine based. So this allows the installer to be executed on unsupported distributions.


src/helperFunctions/install.py
```diff
@@ -242,8 +242,12 @@ def check_distribution():
     if distro.id() == 'fedora':
         logging.debug('Fedora detected')
         return 'fedora'
-    logging.critical('Your Distribution ({} {}) is not supported. FACT Installer requires Ubuntu 18.04, 20.04 or compatible!'.format(distro.id(), distro.version()))
-    sys.exit(1)
+    msg = f'Your Distribution ({distro.id()} {distro.version()}) is not supported. FACT Installer requires Ubuntu 18.04, 20.04 or compatible!'
+    if allow_unsupported:
+        logging.warning(msg + '(Ignoring)')
```
jstucke, line 247: IMHO we can omit the "ignoring" and even convert it to log level "info" so it does not get mixed up with actual warnings or errors that might occur during installation and which we might want to report to the user (also for FACT docker)

maringuu: Done.

jstucke (approved):

PR #787, opened 2022-05-25
Refactoring and test coverage

jstucke:
- added error handling for cases when trying to add an object to the DB when the root object or the parent(s) are missing (can happen when deleting a FW that is currently being analyzed)
- fixed a bug in the file diff where diffs containing HTML elements could cause errors (especially diffs containing `</script>` because this closes the surrounding script tag even when it is inside a JavaScript string)
- refactoring

### DIFF
--- a/src/web_interface/components/compare_routes.py
+++ b/src/web_interface/components/compare_routes.py
@@ -167,7 +167,7 @@ class CompareRoutes(ComponentBase):
             file1.content.splitlines(keepends=True), file2.content.splitlines(keepends=True),
             fromfile=f'{file1.file_name}', tofile=f'{file2.file_name}'
         )
-        return ''.join(diff_list).replace('`', '\\`')
+        return ''.join(diff_list).replace('`', '\\`').replace('<', '\\<\\/')
 
     def _get_data_for_file_diff(self, uid: str, root_uid: Optional[str]) -> FileDiffData:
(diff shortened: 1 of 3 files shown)


src/web_interface/components/compare_routes.py
```diff
@@ -167,7 +167,7 @@ def _get_file_diff(file1: FileDiffData, file2: FileDiffData) -> str:
             file1.content.splitlines(keepends=True), file2.content.splitlines(keepends=True),
             fromfile=f'{file1.file_name}', tofile=f'{file2.file_name}'
         )
-        return ''.join(diff_list).replace('`', '\\`')
+        return ''.join(diff_list).replace('`', '\\`').replace('<', '\\<\\/')
```
rhelmke, line 170: I'm not quite sure why we sanitize here. The corresponding [template marks the resulting diffstr as safe](https://github.com/fkie-cad/FACT_core/blob/a2433cccab38fec9dae51285f0ac2c2ef2d86e3d/src/web_interface/templates/compare/text_files.html#L24), which appears not quite right. Why don't we mark the data as unsafe and let jinja2 sanitize its content? If that doesn't work as required here, maybe we should (by default) replace all special chars as HTML entities in the diffstr.

rhelmke (changes requested):

rhelmke (approved):

PR #791, opened 2022-06-01
updated linux kernel software signature

jstucke:
- the old version does not detect minor versions with 3 digits
- also includes suffixes


rhelmke (approved):
looks good to me!

PR #802, opened 2022-06-14
removed WebFrontEnd from manage_users and replaced it with new app module

jstucke:
manage_users script unnecessarily depended on the full `WebFrontEnd` with all its endpoints
- replaced with empty Flask app
- refactored app creation to own module
- pylint fixes


rhelmke (approved):
Works fine for me. Thanks!

Putting the actual webapp into a separate module makes sense.

This drastically reduces the dependencies of the user management.

PR #808, opened 2022-06-20
run_cmd_with_logging: Also print stdout on error

maringuu:
It turns out stderr it not enough.
E.g. `docker build` writes the output of a failed RUN command to stdout


maringuu:
Closing as master already combines stderr and stdout.

PR #811, opened 2022-06-23
Do not overwrite docker proxy settings when proxy environment variables are set

rhelmke:
So I'm not quite sure which route to go here. To honor environment variables from the `*_PROXY` family, @jstucke initially implemented a function that checks for `*_PROXY` environment variable presence. The idea was to pass proxy settings to docker containers during the build process of some dockerized plugins.

However, in some edge cases these `--build-arg` switch injections make it very hard to select the correct path in a multi-proxy environment. Exemplary scenario: *Use caching proxy **A** for the whole system, but use non-caching proxy **B** for docker containers*.

Now, the current implementation would not only use the system-wide proxy, but would also overwrite explicit [proxy configuration for docker environments](https://docs.docker.com/network/proxy/#configure-the-docker-client).

This draft PR suggests to remove the `--build-arg` functionality from the qemu plugin installation because it provides more flexibility in rather complex network setups.

### DIFF
--- a/src/plugins/analysis/qemu_exec/install.py
+++ b/src/plugins/analysis/qemu_exec/install.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import logging
-import os
 import urllib.request
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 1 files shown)


jstucke (approved):
Looks good to me. It would probably be a good idea if we could add a section to `install.md` with information on running FACT behind a proxy.

rhelmke:
Great, so I just wrote a few paragraphs for proxy configuration in `INSTALL.md`.

PR #813, opened 2022-06-27
Update analysis: Make it working

maringuu:
Now everything can be updated just as when uploading a firmware.
~WIP because I didn't yet adapt the tests.~


src/web_interface/components/analysis_routes.py
```diff
@@ -121,13 +122,20 @@ def get_update_analysis(self, uid, re_do=False, error=None):
...
+            plugin_dict = intercom.get_available_analysis_plugins()
+
+        # Add a temporary preset that holds the plugins of the current analysis
+        current_analysis_preset = old_firmware.uid
+        analysis_presets = [current_analysis_preset] + list(self._config['default-plugins'])
 
         previously_processed_plugins = list(old_firmware.processed_analysis.keys())
-        with ConnectTo(self.intercom, self._config) as intercom:
-            plugin_dict = self._overwrite_default_plugins(intercom.get_available_analysis_plugins(), previously_processed_plugins)
+        # FIXME: why is this even in there?
```
dorpvom, line 133: This snippet replaces the default plugins (default set when uploading a new firmware) with the plugins run previously on the firmware. I'd rather keep it in a function as before. Maybe rename to sth. like `set_selected_plugins_to_previous`.

src/scheduler/analysis.py
```diff
@@ -229,18 +229,19 @@ def get_plugin_dict(self) -> dict:
...
-        default_flag_dict = {}
+        plugin_sets = self._get_plugin_sets_from_config()
         result = {}
         for plugin in plugin_list:
+            current_plugin_plugin_sets = {}
             mandatory_flag = plugin in MANDATORY_PLUGINS
-            for key in default_plugins:
-                default_flag_dict[key] = plugin in default_plugins[key]
+            for plugin_set in plugin_sets:
+                current_plugin_plugin_sets[plugin_set] = plugin in plugin_sets[plugin_set]
```
dorpvom, line 238: plugin_plugin is a typo, correct?

src/scheduler/analysis.py
```diff
@@ -229,18 +229,19 @@ def get_plugin_dict(self) -> dict:
...
             blacklist, whitelist = self._get_blacklist_and_whitelist_from_plugin(plugin)
+            # TODO this should not be a tuple but rather a dictionary/class
```
dorpvom, line 240: There was a reason for this once. But I forgot it ...

src/test/unit/helperFunctions/test_task_conversion.py
```diff
@@ -62,3 +65,4 @@ def test_convert_analysis_task_to_firmware_object(self):
         self.assertEqual(len(fw_obj.scheduled_analysis), 2)
         self.assertIn('dummy', fw_obj.scheduled_analysis)
         self.assertIsInstance(fw_obj.tags, dict, 'tag type not correct')
+        assert list(fw_obj.tags.keys()) == ['a', 'b'], 'tags not correct'
```
dorpvom, line 68: Where does `'a'` and `'b'` come from?

dorpvom (changes requested):

src/scheduler/analysis.py
maringuu: It was intended to be read as plugin_sets of current_plugin.
The structure of the dict that is returned contains a list of plugin sets for each plugin.
So while the name is a bit wired I think the name better conveys this information.

src/test/unit/helperFunctions/test_task_conversion.py
maringuu: They are assigned by the `convert_analysis_task_to_fw_obj` call.

src/scheduler/analysis.py
maringuu: Then I'd prefer to have the comment.

src/web_interface/components/analysis_routes.py
maringuu: I extracted the logic in a function.

src/test/unit/web_interface/test_re_analyze.py
```diff
@@ -1,11 +0,0 @@
-from test.common_helper import CommonIntercomMock
```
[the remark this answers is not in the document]
maringuu: The function `overwrite_default_plugins` was removed. I don't think the new function is complex enough to add a new test.

dorpvom (approved):

PR #814, opened 2022-06-27
Priority queue poc

jstucke:
added custom multiprocessing priority queue class to enable priority scheduling for unpacking and analysis


maringuu:
I didn't have a close look but what about using [queue.PriorityQueue](https://docs.python.org/3/library/queue.html#queue.PriorityQueue) as a queue implementation?

jstucke:
Testing this revealed that, though it is working correctly, it is way too slow for larger files.

PR #816, opened 2022-07-08
Fix session recycling

jstucke:
When multiple worker threads are configured in the frontend, the same DB interface instances and with that the same sessions may be used by multiple threads which caused errors. This PR moves the DB engine to its own class so that the DB interface can be instanced for each worker thread in the frontend.

### DIFF
--- a/src/helperFunctions/database.py
+++ b/src/helperFunctions/database.py
@@ -1,9 +1,8 @@
 from configparser import ConfigParser
 from contextlib import contextmanager
-from typing import Generic, Type, TypeVar
+from typing import ContextManager, Generic, Type, TypeVar
 
 DatabaseInterface = TypeVar('DatabaseInterface')
-DbInterface = TypeVar('DbInterface')
 
 
 class ConnectTo(Generic[DatabaseInterface]):
@@ -34,6 +33,6 @@ class ConnectTo(Generic[DatabaseInterface]):
 
 
 @contextmanager
-def get_shared_session(database: DbInterface) -> DbInterface:
+def get_shared_session(database: DatabaseInterface) -> ContextManager[DatabaseInterface]:
     with database.get_read_only_session():
         yield database
--- a/src/helperFunctions/database.py
+++ b/src/helperFunctions/database.py
@@ -1,7 +1,9 @@
 from configparser import ConfigParser
+from contextlib import contextmanager
 from typing import Generic, Type, TypeVar
 
 DatabaseInterface = TypeVar('DatabaseInterface')
+DbInterface = TypeVar('DbInterface')
 
 
 class ConnectTo(Generic[DatabaseInterface]):
@@ -29,3 +31,9 @@ class ConnectTo(Generic[DatabaseInterface]):
 
     def __exit__(self, *args):
         pass
+
+
+@contextmanager
+def get_shared_session(database: DbInterface) -> DbInterface:
+    with database.get_read_only_session():
+        yield database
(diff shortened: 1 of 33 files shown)


rhelmke (changes requested):

src/helperFunctions/database.py
```diff
@@ -1,7 +1,9 @@
 from configparser import ConfigParser
+from contextlib import contextmanager
 from typing import Generic, Type, TypeVar
 
 DatabaseInterface = TypeVar('DatabaseInterface')
+DbInterface = TypeVar('DbInterface')
```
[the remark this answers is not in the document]
jstucke: fixed. Also added `ContextManager[]` so that PyCharm is able to correctly determine the returned type

src/web_interface/components/ajax_routes.py
```diff
@@ -37,22 +37,23 @@ def _get_exclusive_files(self, compare_id, root_uid):
...
 
     @roles_accepted(*PRIVILEGES['view_analysis'])
     @AppRoute('/ajax_root/<uid>/<root_uid>', GET)
     def ajax_get_tree_root(self, uid, root_uid):
         root = []
-        for node in self.db.frontend.generate_file_tree_level(uid, root_uid):  # only a single item in this 'iterable'
-            root = [convert_to_jstree_node(node)]
+        with get_shared_session(self.db.frontend) as db:
+            for node in db.generate_file_tree_level(uid, root_uid):  # only a single item in this 'iterable'
+                root = [convert_to_jstree_node(node)]
```
rhelmke, line 56:
```suggestion
        with get_shared_session(self.db.frontend) as frontend_db:
            for node in frontend_db.generate_file_tree_level(uid, root_uid):  # only a single item in this 'iterable'
                root = [convert_to_jstree_node(node)]
```

src/web_interface/rest/rest_compare.py
```diff
@@ -86,8 +87,9 @@ def get(self, compare_id):
             )
 
         result = None
-        if self.db.comparison.comparison_exists(compare_id):
-            result = self.db.comparison.get_comparison_result(compare_id)
+        with get_shared_session(self.db.comparison) as db:
+            if db.comparison_exists(compare_id):
+                result = db.get_comparison_result(compare_id)
```
rhelmke, line 92:
```suggestion
        with get_shared_session(self.db.comparison) as comparison_db:
            if comparison_db.comparison_exists(compare_id):
                result = comparison_db.get_comparison_result(compare_id)
```

src/web_interface/rest/rest_compare.py
```diff
@@ -33,21 +33,22 @@ def put(self):
...
-            return error_message(
-                'Compare already exists. Use "redo" to force re-compare.',
-                self.URL, request_data=request.json, return_code=200
-            )
+        with get_shared_session(self.db.comparison) as db:
+            if db.comparison_exists(compare_id) and not data['redo']:
+                return error_message(
+                    'Compare already exists. Use "redo" to force re-compare.',
+                    self.URL, request_data=request.json, return_code=200
+                )
```
rhelmke, line 41:
```suggestion
        with get_shared_session(self.db.comparison) as comparison_db:
            if comparison_db.comparison_exists(compare_id) and not data['redo']:
                return error_message(
                    'Compare already exists. Use "redo" to force re-compare.',
                    self.URL, request_data=request.json, return_code=200
                )
```

rhelmke (changes requested):
Found some more generic usage of `with get_shared_session(...) as db`

PR #823, opened 2022-07-26
reduce false negative matches in information leaks plugin path matching

rhelmke:
The current regular expressions in the `information_leaks` path matching fail when the path hierarchy ends after level two, e.g. `/root/.ssh`. Depending on the follow-up characters, the expressions then match either nothing or include non-printable characters. The first edge case is a false-negative. The second edge case parse a corrupted path including non-printable bytes. Since the postgres database doesn't accept such data on `string`-type columns, the backend crashes once analysis results are written to the database. As our old DBMS `mongodb` just swallowed those values, nobody noticed.


jstucke (approved):
Looks good to me

PR #786, opened 2022-05-23
Introduce and use analysis_plugin fixture for plugin tests

maringuu:
This is a cherrypick of two commits of #763. Depends on #749.
The fixture is intendet to reduce complexity of the testing framework.
It should be easier to write plugin tests now.
Also it removes duplication.
```sh
$ gds --shortstat config_rework
 38 files changed, 1072 insertions(+), 1284 deletions(-)
```

PR #763, opened 2022-04-06
Rework testing to use fixtures

maringuu:
Depends on #749 (Note for reviewing: I'll probably rebase on the config_rework branch when its review is finished)

See individual commits.
The tests are mosly unchanged. All new code is contained in the the several `conftest.py` files.

PR #825, opened 2022-07-29
Do not merge: Rework testing new

maringuu:
This is a reincarnation of #763 
The commits are independent of each other and review should be easier.
Depends on #749 and includes #786.

## Changes
Some more fixtures are added.
Mainly:
- use_database: Indicate that the actual postgres database shall be used, creates test tables and removes them when the fixture goes out of scope
- mock_database: The inverse to use_database. All other fixtures respect use_database/mock_database.
- web_frontend: An instance of WebFrontEnd
- analysis_scheduler: An instance of AnalysisScheduler

Also the use of unittest is removed. We now just use pytest and don't have several ways to perform patching but just one.

## Rationale
Previously there was much code duplication, especially the instanciation of WebFrontEnd and AnalysisScheduler.
Now they are only instanciated in one place, this also makes it easier to understand test failures. Adding new tests is also easier because you can just use these fixtures and exepct everything to work.

Removing unittest unifies the code base and also the module is not needed when pytest fixtures are used.

# Status
This is still a draft, as many changes of #763 are not included yet.


maringuu:
This PR is now just an archive to be cherrypicked from and will be closed once I'm happy with the code that we cherrypicked.

maringuu:
Closing as I don't need that much code from here and cherrypicked what I wanted.

PR #828, opened 2022-08-02
assure that pinned werkzeug version is not overwritten

jstucke:
assure that pinned werkzeug version is not overwritten during installation of other dependencies (flask, flask-login, flask-restx) by putting it right at the bottom of the list

### DIFF
--- a/src/install/requirements_frontend.txt
+++ b/src/install/requirements_frontend.txt
@@ -1,5 +1,3 @@
-werkzeug==2.1.2
-
 bcrypt==3.2.2
 email-validator==1.2.1
 flask-login==0.6.1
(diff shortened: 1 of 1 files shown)


src/install/requirements_frontend.txt
```diff
@@ -16,6 +14,7 @@ python-dateutil==2.8.2
 si-prefix==1.2.2
 uwsgi==2.0.20
 virtualenv
+werkzeug==2.1.2
```
maringuu, line 17: Maybe add a comment above this line to prevent anyone from reordering this list?

jstucke: Good idea

PR #829, opened 2022-08-02
updated all flask dependencies for flask 2.2

jstucke:
supersedes #828 if it works


jstucke:
flask-restx doesn't work yet with Flask 2.2. As soon as it is fixed (maybe with https://github.com/python-restx/flask-restx/pull/463) we should pick this up again

maringuu:
Superseeded by #897

PR #834, opened 2022-08-08
added summaries to analysis page for file objects

jstucke:
adds intermediate summaries for all file objects with "included files"


dorpvom:
I'm thinking of rejecting solely based on all the refactoring ^^

src/test/integration/storage/test_db_interface_common.py
```diff
@@ -218,16 +218,6 @@ def test_get_summary(db):
     assert fw.uid not in result_sum['file exclusive sum b'], 'parent as origin but should not be'
 
 
-def test_collect_summary(db):
```
dorpvom, line 221: Seeing you deleted a test and none is renamed or adapted one has to wonder if this existing functionality and the new adapted one is tested anywhere. Is it?

src/web_interface/templates/show_analysis.html
```diff
@@ -303,7 +303,7 @@ <h5 class="modal-title">Add analysis to file</h5>
                 {% endblock %}
 
                 {# summary of included files #}
-                {%- if firmware.vendor and selected_analysis -%}
+                {%- if selected_analysis -%}
```
dorpvom, line 306: As discussed we want to check if this can be lazy loaded with a click-of-a-button, so no unnecessary data is transferred while perusing the file tree.

dorpvom (changes requested):

src/web_interface/templates/show_analysis.html
jstucke: What do you think would be the best (most intuitive) text for the button? "Load Summary"? "Generate Summary"? "Show Summary"? Or "Load/Generate/Show Analysis Summary" maybe?

src/test/integration/storage/test_db_interface_common.py
jstucke: `_collect_summary_for_uid_list` may not get tested but `get_summary` (which calls the function) does. I added a new test case for this function to explicitly also test summary generation for `FileObject`s

src/web_interface/templates/show_analysis.html
dorpvom: I changed to `Load Summary for Included Files` as a short enough and understandable. Changed naming in other places to be consistent.

dorpvom (approved):

PR #836, opened 2022-08-10
changed rest/fo endpoint to return all VFPs

### LINKED ISSUE #833: missing virtual file paths in rest/file_object
the REST endpoint `/rest/file_object/<uid>` returns the file object data including its "virtual file paths" (VFPs). But only the VFPs for one root UID are returned and it is not possible to specify the root UID. Therefore, the VFP data for all other firmwares (root UIDs) cannot be queried via this endpoint. 

Possible fixes:
- returning the VFPs for all root UIDs
- adding the parameter `root_uid` to the endpoint

jstucke:
changed `rest/file_object/<uid>` endpoint to return all virtual file paths instead of paths for one random Firmware
resolves #833


rhelmke (approved):
Looks good. Thanks!

PR #838, opened 2022-08-12
add sensible defaults for anaylsis db entries

jstucke:
the DB schema allows `null` values for `AnalysisEntry.result`
While there shouldn't be a Null entries at the moment, such entries would cause errors for stats updates. This is because the stats generation functions use PostgreSQL functions which generally don't work on Null values.
This modification adds a reasonable default value, which should prevent such problems in the future.


rhelmke (approved):
Thanks! Works for me.

PR #839, opened 2022-08-16
Firmware: Remove dead code

### LINKED ISSUE #812: User defined tags color can't be set
The color of user defined tags always defaults to gray.
In the code there is support for setting a custom color, but it can't be set through the user interface.
We could either support setting custom colors or adapt the code to make clear that we don't support multiple colors.

Relevant function: `Firmware.set_tag`

maringuu:
Closes #812


jstucke (approved):
Looks good to me

PR #840, opened 2022-08-16
Run pyupgrade

maringuu:
For future reference:
```sh
pyupgrade --py37-plus $(git ls-tree -r --name-only master| rg "\.py$")
```
Four files need manual fixing as pyupgrade does not insert blank lines sometimes when needed.

I'm not sure if 3.7+ is the right python version. What version are we currently supporting?


src/plugins/analysis/known_vulnerabilities/docker/scripts/detect_CVE-2021-45608.py
```diff
@@ -19,7 +19,7 @@ def find_kmalloc_calls_in_function(func):
         if call.getName() == '__kmalloc':
             for ref in getReferencesTo(call.getEntryPoint()):  # noqa
                 if func.getBody().contains(ref.getFromAddress()):
-                    print('found __kmalloc call @ {}'.format(ref.getFromAddress()))
```
jstucke, line 22: this file is a Ghidra / jython script and cannot be upgraded in this way (which is also the reason the tests fail in the CI)

jstucke (changes requested):
I didn't know `pyupgrade` until now and it seems to be really useful.

> I'm not sure if 3.7+ is the right python version. What version are we currently supporting?

Generally all stable versions (so 3.7 to 3.10 at the moment and also 3.11 soon)

> Four files need manual fixing as pyupgrade does not insert blank lines sometimes when needed.

Do they still need fixing? Which files?

maringuu:
I reran everything ontop of master.

> Do they still need fixing? Which files?

Now this pr is split in two commits. The second fixes the files.

> This is also a ghidra script and cannot be upgraded this way

Oh of course.

jstucke (approved):

PR #842, opened 2022-08-17
Proposal: Autoformat code

maringuu:
I went for yapf since it allows for some customisation and I prefer its style over the non-configurable style of black.
The advantage of yapf over autopep8 is that the code is more readable but still pep8 compatible.

As the FACT codebase has many really long lines many lines are split by the autoformatter.
I personally perfer this style over 200 column long lines.


jstucke:
Since we probably want to use black, I will close this PR.

PR #844, opened 2022-08-23
added target, ref to hyperlink in cve view (thanks Logan @lcarp17)

dorpvom:


jstucke (approved):
Looks good to me and seems to work as intended

PR #845, opened 2022-08-23
linux kernel yara rule: exchange [1-5] limit on major release to \d

rhelmke:
Major release 6 is around the corner - we better future-proof our YARA rule :-)


jstucke (approved):
Looks good to me. I don't know why it wasn't written like this in the first place.

rhelmke:
PR passed CI checks, lets merge then!

PR #850, opened 2022-09-08
Clean up program_setup and dont magically change logfile names

maringuu:
See individual commits.

Depends on #749 because of the tests.


maringuu:
@jstucke this needs the CI script to be updated because of the removal of `--debug` and `--silent`

maringuu:
This is now rebased against current master and the CI should pass.

maringuu:
@jstucke Ci passes now, should be good to go.

src/helperFunctions/program_setup.py
```diff
@@ -64,40 +39,59 @@ def _setup_argparser(name, description, command_line_options, version=__VERSION_
...
-    log_level = getattr(logging, cfg.logging.loglevel, None)
+def get_logging_config(args, component):
+    """
+    Returns a tuple of (logfile, loglevel) read from args and the config file.
+    The loglevel is returned as an integer.
+    Assumes that `config.load` was called beforehand.
+    """
+    console_loglevel = logging.getLevelName(args.log_level)
+    if isinstance(console_loglevel, str):
+        raise Exception(f'Invalid log level: {console_loglevel}')
```
jstucke, line 64: I don't think this is even necessary as the possible values for this are limited by the "choices" of the argument in argparse (meaning this should always result in a valid log level).

src/install/db.py
```diff
@@ -63,7 +63,7 @@ def main():
 
     with OperateInDirectory('../../'):
         with suppress(FileNotFoundError):
-            Path('start_fact_db').unlink()
-        Path('start_fact_db').symlink_to('src/start_fact_db.py')
+            Path('start_fact_database').unlink()
+        Path('start_fact_database').symlink_to('src/start_fact_database.py')
```
jstucke, line 67: renaming this could mean that you have two files `start_fact_db` and `start_fact_database` after you run this in an existing FACT instance (which could be confusing and could even lead to errors if the user keeps using the old one).

src/start_fact.py
```diff
@@ -56,13 +53,17 @@ def _evaluate_optional_args(args: argparse.Namespace):
 
 def _start_component(component: str, args: argparse.Namespace) -> Popen | None:
     script_path = Path(get_src_dir()) / f'../start_fact_{component}'
-    if not script_path.exists():
-        logging.debug(f'{component} not installed')
-        return None
```
jstucke, line 61: This script is used to start all *installed* components (hence the name `start_all_installed_fact_components` of the symlink). If we remove this, it will no longer work for distributed setups where not all components are installed.

src/start_fact.py
```diff
@@ -56,13 +53,17 @@ def _evaluate_optional_args(args: argparse.Namespace):
...
-    logging.info(f'starting {component}')
```
jstucke, line 62: Why did you remove this?

src/start_fact.py
```diff
@@ -71,60 +72,37 @@ def _terminate_process(process: Popen):
...
 
 
 signal.signal(signal.SIGINT, shutdown)
 signal.signal(signal.SIGTERM, shutdown)
 
 
-class FactStarter:
-    run = False
+if __name__ == '__main__':
+    process_list = []
```
jstucke, line 90: This seems to be unused

src/test/unit/helperFunctions/test_program_setup.py
```diff
@@ -2,47 +2,43 @@
...
+    assert cfg.logging.logfile_frontend == logfile
 
-def test_setup_logging():
-    with TemporaryDirectory() as tmp_dir:
-        args = ArgumentMock(tmp_dir)
-        set_logging_cfg_from_args(args)
-        setup_logging(args)
-        logger = logging.getLogger('')
-        assert logger.getEffectiveLevel() == logging.DEBUG
+    _ = get_logging_config(ArgumentMock, 'non-default')
```
jstucke, line 35: What does this part do? Is there a part missing from the test?

src/update_statistic.py
```diff
@@ -19,17 +19,19 @@
...
 
 
 def main(command_line_options=None):
-    if command_line_options is None:
-        command_line_options = sys.argv
-    program_setup(PROGRAM_NAME, PROGRAM_DESCRIPTION, command_line_options=command_line_options)
+    args = setup_argparser(PROGRAM_NAME, PROGRAM_DESCRIPTION, command_line_options=command_line_options or sys.argv)
+    config.load(args.config_file)
+    logfile, file_loglevel, console_loglevel = get_logging_config(args, 'statistic')
+    setup_logging(logfile, file_loglevel, console_loglevel)
```
jstucke, line 34: see above

jstucke (commented):
I think we need to discuss some of the changes

src/start_fact.py
```diff
@@ -56,13 +53,17 @@ def _evaluate_optional_args(args: argparse.Namespace):
...
-        logging.debug(f'{component} not installed')
-        return None
-    logging.info(f'starting {component}')
     optional_args = _evaluate_optional_args(args)
-    command = f'{script_path} -l {cfg.logging.logfile} -L {cfg.logging.loglevel} -C {args.config_file} {optional_args}'
-    return Popen(split(command))
+    command = f'''{script_path} \
+        -l {getattr(cfg.logging, f"logfile_{component}")} \
+        -L {cfg.logging.loglevel} \
+        '''
```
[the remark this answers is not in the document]
maringuu: What about double slashes?
I see that there is room to argue, but shell commands become way more readable if you split the arguments to multiple lines.

src/helperFunctions/program_setup.py
maringuu: Good catch.

src/install/db.py
maringuu: I don't see a problem here.
Mark this as a breaking change in the changelog and users are good to go.
Also if someone uses FACT they probably should be competent enough to realise that the symlink is invalid and then recognise that there is a new file which has almost the same name.

src/start_fact.py
maringuu: Hmm I get that. But what is the usecase for this script if you have a distributed setup. If I'd deploy FACT I would definitely not use this script but rather start the compoents directly via their start scripts.
That also is what we do in FACT_docker.

But yeah, I can remove this change if you disagree.

src/start_fact.py
maringuu: I don't see any benefit in logging that FACT started. Isn't that kinda obvious?
If we printed some information about FACT (e.g. version) like postgres and redis do I'd see a reason to print this.
But I will revert this change if you disagree.

Redis: ` # Redis version=6.0.16, bits=64, commit=00000000, modified=0, pid=677410, just started`

postgresql: `starting PostgreSQL 14.6 (Ubuntu 14.6-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0, 64-bit`

jstucke: Since FACT heavily relies on multiprocessing, I find it somewhat reassuring to see that every component was started successfully. I would say that it makes sense for logging level "info". I can also understand why you think that it does not actually convey any useful information, though. IMHO we should at least move it to logging.debug instead of removing it. Has @dorpvom an opinion on this?

I agree that we should log some version info. That could e.g. be useful for error logs in the new issue template :sweat_smile:

src/start_fact.py
jstucke: I agree that now that the use of `start_fact_db` is at least questionable, the cases where you would use the script in this way are very limited. If we decide to remove it, we should rename `start_all_installed_fact_components`, though. Otherwise, the name would be misleading.

src/start_fact.py
```diff
@@ -71,60 +72,37 @@ def _terminate_process(process: Popen):
...
-
-def _process_is_running(process: Popen) -> bool:
-    try:
-        os.kill(process.pid, 0)
-        if process.poll() is not None:
-            return False
-        return True
-    except ProcessLookupError:
-        return False
+    global run
```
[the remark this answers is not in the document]
jstucke: This may very well be what the global statement was originally intended for. I would still argue (for the reasons mentioned above) that if we *can* avoid using it, we should do so. We should probably convert it to an instance variable and move the `shutdown()` function into the class, though.

src/install/db.py
jstucke: > Mark this as a breaking change in the changelog and users are good to go.

So you expect that all users actually read that? :laughing: I just want to avoid reasons for new issues if possible.

We could turn it into a "non-breaking" (or at least "less breaking"?) change if we also try to remove `start_fact_db` in addition to the new file. This would avoid any inconsistencies upfront. I'm not opposed to renaming the file.

jstucke (approved):

PR #851, opened 2022-09-08
Remove fact_env.sh

maringuu:
According to the comment this was done due to something related to mongodb which was removed.


jstucke (approved):
Looks good to me. I wonder what other unused artifacts are hidden somewhere in the code.

Issue #820, opened 2022-07-13
why are there missing files


Oren-i:
For some firmware images, the analysis never finishes, and there are missing files in the Admin/Find Missing Analysis tab.

Is this expected? In case no, how can I see what it went wrong to fix it? There are no messages on the Logs tab, and I am using FACT_docker.

jstucke:
Hi,

> For some firmware images, the analysis never finishes

Do you mean that entries in "Currently analyzed firmware" (http://localhost:5000/system_health) never complete? That could happen if there are errors during analysis or unpacking and the file gets lost during scheduling (but it should obviously not happen). Since there don't seem to be any helpful log messages, it could be complicated to debug the problem. Did you maybe see any errors or stack traces in the terminal output (`docker logs` could help here since you are using FACT_docker)? There could be unexpected errors that don't result in log messages.

Oren-i:
> Do you mean that entries in "Currently analyzed firmware" (/system_health) never complete? 

Correct. 

I rerun the test and reproduced the error, but I now see that there is a time out exception that seems to be not handled correctly, maybe because when handling an exception other exceptions were raised. Attached is the exception log message.
[fact_log.txt](https://github.com/fkie-cad/FACT_core/files/9122024/fact_log.txt)

The extractor that seems to take a long time before the exception is binwalk as seen by ps.

jstucke:
The error is indeed not handled correctly. Nevertheless, it is also not clear what caused the error in the first place. Was it a particularly large or in some other way unusual file? 
You could also try to run the extractor manually on the file as documented [here](https://github.com/fkie-cad/fact_extractor#command-line-usage) to maybe see what causes the error.

Oren-i:
The issue here I think is that binwalk does indeed take too much time for some files, and that FACT_core does not correctly handle timeouts in FACT_extractor. In some cases binwalk extracts bogus data and as FACT_extractor is called in a recursive manner, a very large file can be sent to binwalk for further extraction. 

Feel free to assign this to me.

jstucke:
We are always happy to receive external contributions and will try to support you, so feel free to try to improve this. Some things to note:
- binwalk is only used in two cases for extraction:
    1. when the file format is not known (e.g. the file is a binary blob without headers)
    2. when the extraction for the file's format with the designated plugin fails (as a fall-back option)
- the output of binwalk is already (partly) filtered: we try to sort out bogus archives by verifying if the output is really an archive of the type
[...]

PR #852, opened 2022-09-08
Handle timeout errors when calling fact_extractor, related to #820

Oren-i:
Timeouts are handled by catching all requests exceptions because requests raise ConnectionError on a timeout.

### DIFF
--- a/src/unpacker/unpack.py
+++ b/src/unpacker/unpack.py
@@ -36,6 +36,10 @@ class Unpacker(UnpackBase):
         with TemporaryDirectory(prefix='fact_unpack_', dir=self.config['data-storage']['docker-mount-base-dir']) as tmp_dir:
             file_path = self._generate_local_file_path(current_fo)
             extracted_files = self.extract_files_from_file(file_path, tmp_dir)
+            if extracted_files is None:
+                self._store_unpacking_error_skip_info(current_fo)
+                return []
+
             extracted_file_objects = self.generate_and_store_file_objects(extracted_files, Path(tmp_dir) / 'files', current_fo)
             extracted_file_objects = self.remove_duplicates(extracted_file_objects, current_fo)
             self.add_included_files_to_object(extracted_file_objects, current_fo)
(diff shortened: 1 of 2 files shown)


src/unpacker/unpack.py
```diff
@@ -44,6 +48,15 @@ def unpack(self, current_fo: FileObject):
...
         return extracted_file_objects
 
+    @staticmethod
+    def _store_unpacking_error_skip_info(file_object: FileObject):
+        file_object.processed_analysis['unpacker'] = {
+            'plugin_used': 'None', 'number_of_unpacked_files': 0, 'plugin_version': '0.0', 'analysis_date': time(),
+            'info': 'Unpacking stopped because extractor raised a exception (possible timeout)',
+        }
+        tag_dict = {'unpacker': {'extractor error': {'value': 'possible extractor timeout', 'color': TagColor.ORANGE, 'propagate': False}}}
+        file_object.analysis_tags.update(tag_dict)
```
jstucke, line 58: The tags don't seem to be saved correctly in the database. But something like this should work:

```suggestion
        file_object.processed_analysis['unpacker'] = {
            'plugin_used': 'None', 'number_of_unpacked_files': 0, 'plugin_version': '0.0', 'analysis_date': time(),
            'info': 'Unpacking stopped because extractor raised a exception (possible timeout)',
            'tags': {'extractor error': {'value': 'possible extractor timeout', 'color': TagColor.ORANGE, 'propagate': False}},
        }
```

This also applies to the `_store_unpacking_depth_skip_info()` method below

jstucke (commented):
Looks good to me and seems to work as intended with one exception: the tags don't seem to be rendered correctly

jstucke:
What is more, it would probably be a good idea to add `, 'tags'` to the list in l. 30 of `src/web_interface/templates/analysis_plugins/unpacker.html` so that tags are no longer shown with the rest of the "analysis" data on the analysis results page of the unpacker

src/unpacker/unpack.py
Oren-i: Converting to draft as the changes are going do cause unneeded timeouts when using fact_extractor in standalone mode. I think the correct way to implement this should be by getting the timeout value from fact_core as a parameter.

jstucke: > Converting to draft as the changes are going do cause unneeded timeouts when using fact_extractor in standalone mode.

I'm not sure I can follow. The changes in this PR should not affect the extractor. Did you maybe mean https://github.com/fkie-cad/fact_extractor/pull/94?

I just tested the changes again and everything seems to work fine (with one exception: analysis tags are not always rendered correctly for firmware containers but that is something that is out of scope for this PR).

jstucke (approved):
Looks good to me. Thanks again for your contribution!

src/unpacker/unpack.py
Oren-i: You are correct, I meant to mark the other pull request as draft. This pull request is fine (I can not remove the draft status myself).

PR #870, opened 2022-10-07
added .svn artifacts to information leaks plugin

jstucke:
I stumbled upon a .svn folder inside a D-Link firmware and had to discover that the "information_leaks" plugin did not detect it yet.

- added `.svn` artifacts to information leaks plugin
- updated some descriptions
- some minor refactoring

### DIFF
--- a/src/plugins/analysis/information_leaks/code/information_leaks.py
+++ b/src/plugins/analysis/information_leaks/code/information_leaks.py
@@ -1,60 +1,60 @@
+from __future__ import annotations
+
 import re
 from itertools import chain
 
 from analysis.PluginBase import AnalysisBasePlugin
+from helperFunctions.virtual_file_path import get_top_of_virtual_path
 from objects.file import FileObject
 
 PATH_REGEX = {
     'user_paths': re.compile(rb'/home/[^%\n \x00]+'),
     'root_path': re.compile(rb'/root/[^%\n \x00]+'),
-    'www_path': re.compile(rb'/var/www/[^\n \x00]+')
+    'www_path': re.compile(rb'/var/www/[^\n \x00]+'),
 }
 
 FILES_REGEX = {
-    'any_history': re.compile(rb'.+_history')
+    'any_history': re.compile(rb'.+_history'),
 }
 
 PATH_ARTIFACT_DICT = {
     '.git/config': 'git_config',
+    '.svn/entries': 'svn_entries',
     '.conda/environments.txt': 'conda_environment',
     'default.conf': 'possible_code_blocks_config',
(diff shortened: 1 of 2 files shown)


src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -63,55 +61,68 @@ class AnalysisPlugin(AnalysisBasePlugin):
...
+                file_object.processed_analysis[self.NAME].update(
+                    _check_file_path(get_top_of_virtual_path(virtual_path))
+                )
+
+
+def _check_file_path(file_path: str) -> dict[str, list[str]]:
+    for search_function in (_check_for_files, _check_for_directories, _find_files):
+        results = search_function(file_path)
+        if results:
+            return results
```
maringuu, line 95: Is it intendet to ignore the other `search_function`'s once the first one matches?
If it isn't intendet, probably a testcase should be added.

src/plugins/analysis/information_leaks/code/information_leaks.py
```diff
@@ -1,60 +1,58 @@
...
     'user_paths': re.compile(rb'/home/[^%\n \x00]+'),
     'root_path': re.compile(rb'/root/[^%\n \x00]+'),
-    'www_path': re.compile(rb'/var/www/[^\n \x00]+')
+    'www_path': re.compile(rb'/var/www/[^\n \x00]+'),
 }
 
-FILES_REGEX = {
-    'any_history': re.compile(rb'.+_history')
-}
+FILES_REGEX = {'any_history': re.compile(rb'.+_history')}
```
maringuu, line 16: nit: Formatting. For consistency with the others this should not be a single line.

maringuu (changes requested):

maringuu:
Except the comments this LGTM. Still I have another remark.
This PR is a good example for our (internally) discussed way of handeling existing formatting and pylint issues in modified files.
IMO the history should look like this:

> \* src/plugins/analysis/information_leaks/code/information_leaks.py: Run black
> \* added .svn artifacts to information leaks plugin

jstucke:
> IMO the history should look like this:

I did not intentionally run black (it ran because of the pre-commit hook) but I agree. I will try to separate the changes from now on to make it easier to review them.

src/plugins/analysis/information_leaks/code/information_leaks.py
jstucke: Yes, that is the original intention (I only refactored the code to be more "Clean Code compliant")
I added a test for the function to make sure it works as intended

maringuu:
LGTM!

maringuu (approved):

PR #872, opened 2022-10-11
fix multiple newly discovered postgres DB migration issues

jstucke:
fixed some postgres migration issues:
- wrong "sanitize-db" name and respective gridfs exception handling
- `KeyError` in analysis result migration
- handled error that may occur when trying to migrate comparisons of deleted firmware entries

### DIFF
--- a/src/config/migration.cfg
+++ b/src/config/migration.cfg
@@ -3,6 +3,7 @@
 mongo-server = localhost
 mongo-port = 27018
 main-database = fact_main
+sanitize-database = faf_sanitize
 
 # Authentication
 db-admin-user = fact_admin
--- a/src/migrate_db_to_postgresql.py
+++ b/src/migrate_db_to_postgresql.py
@@ -63,7 +63,6 @@ class MongoInterface:
 
 
 class MigrationMongoInterface(MongoInterface):
-
     def _setup_database_mapping(self):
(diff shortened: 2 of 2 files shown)


maringuu:
I'm not that deep into mongo/gridfs/postgres so this LGTM but I didn't have a deeper look.
Also I can't test this since I don't have a mongodb setup anymore.

jstucke:
> I'm not that deep into mongo/gridfs/postgres so this LGTM but I didn't have a deeper look. Also I can't test this since I don't have a mongodb setup anymore.

Well we did kinda test this in the lab while migrating a FACT instance there. So you could say we tested it there and it worked (eventually). :sweat_smile:

jstucke:
Maybe we could use a migration test setup but that could also be a lot of work (and I don't know if it's worth the effort)

maringuu:
> Maybe we could use a migration test setup but that could also be a lot of work (and I don't know if it's worth the effort)

Hmm, good idea/question. I don't think its necessary as probably not that many people use FACT in production, do they?

jstucke:
> So this can be merged?

Probably I guess? The migration on the respective PC was successful after the changes. I don't think anyone of our team still has a MongoDB instance and could also try it out.

maringuu (approved):

PR #867, opened 2022-10-06
 config-dev branch

maringuu:
This branch is intended to transition between the old way of using the config (passing it around) to the new way (importing it).

PR #873, opened 2022-10-13
Cherrypick testing code for config.py

maringuu:
Cherrypicked from #867 

This code should be mergable as it is, as it was already reviewed in #749.
Having this in master allows us to test code using the  `config.py` module (E.g. #850 ).
Since the `config.py` module is already merged in master it makes no sense to only have these two commits in the `config-dev` branch.

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -1,0 +1,143 @@
+from __future__ import annotations
+
+import shutil
+import tempfile
+from configparser import ConfigParser
+from pathlib import Path
+
+import pytest
+
+import config
+from config import Config
+from test.common_helper import create_docker_mount_base_dir
+
+
+def _get_test_config_tuple(defaults: dict | None = None) -> tuple[Config, ConfigParser]:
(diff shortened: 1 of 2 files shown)


src/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+
+    return cfg, configparser_cfg
+
+
+@pytest.fixture
+def cfg_tuple(request):
+    """Returns a `config.Config` and a `configparser.ConfigParser` with testing defaults.
+    Defaults can be overwritten with the `cfg_defaults` pytest mark.
+    """
+    # TODO Use iter_markers to be able to overwrite the config.
```
maringuu, line 114: This is a future TODO that I'll probably implement once needed.
The usecase is that one could theoretically have testing classes that inherit some other class which already sets this marker. When the child class also sets this marker both markers should be merged.

I'm not quite sure if we will (and should) have inheritance in testing classes that use this module. So maybe this TODO can be removed in the future.

jstucke: Maybe we can cross this bridge when we get to it :sweat_smile:

PR #878, opened 2022-10-18
removed non-functional advanced search "variety" and angular

jstucke:
- removed non-functional advanced search "varietyjs" (doesn't work since FACT 4.0)
- removed EoL angular dependency (which also seems to have vulnerabilities)


src/web_interface/templates/database/database_advanced_search.html
```diff
@@ -12,6 +12,19 @@
     </script>
 {% endblock %}
 
+{% block styles %}
+    <style>
+        .example-json {
```
maringuu, line 17: Is `example` intendet or a copy-pase artifact?

src/web_interface/templates/database/database_advanced_search.html
```diff
@@ -112,42 +125,6 @@ <h4>Structure of file objects in database:</h4>
 - tags                  # User defined tags, e.g. Partial, Crawled
 - vendor                # e.g. EvilCorp
 - version               # e.g 1.0.0.312</pre>
-
-            <hr>
-
-            <h4>Structure of processed_analysis field:</h4>
-
-            {% if database_structure is mapping %}
```
maringuu, line 120: Is the `database_structure` variable used anywhere else? If it isn't is there maybe some python code that can be cleaned up too?

maringuu (changes requested):
If I didn't miss something some more things can be removed (See comments).
What about `src/bin/variety.js`?

src/web_interface/templates/database/database_advanced_search.html
jstucke: actually it was intentional since it is used for the _advanced search_ examples :sweat_smile:

src/web_interface/templates/database/database_advanced_search.html
jstucke: You are absolutely right. Removed unused `database_structure` from code and more.

jstucke: Also removed varietyjs installation and config entries. I hope I didn't miss anything this time :sweat_smile:

maringuu:
LGTM!

maringuu:
CI needs fixing

maringuu (approved):

PR #879, opened 2022-10-18
Run black

maringuu:
This PR is the result of running `black .` in `src`.
Sure, some files could be formattet better (e.g. by adding a trailing comma).
When you stumble accross sth. that could be formatted better fix this in a separate commit.

Still i think merging this PR reduces noise of PRs that edit files that do currently not conform to black's formatting.


jstucke:
Maybe we should also "fix" the isort config here so that isort doesn't clash with blacks formatting:
```patch
diff --git a/.isort.cfg b/.isort.cfg
index a6c417ac..bd7247b0 100644
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -4,4 +4,4 @@ default_section = THIRDPARTY
 known_first_party=analysis,compare,helperFunctions,install,intercom,objects,plugins,
     scheduler,statistic,storage,test,unpacker,version,web_interface
 known_third_party=docker
-multi_line_output=6
+profile=black
```
or even better: put it in the `pyproject.toml`

maringuu:
Removed `.isort.cfg` and put the options in `pyproject.toml`.

jstucke (approved):

PR #880, opened 2022-10-18
added SIGTERM signal handler

jstucke:
Added SIGTERM signal handler to support graceful shutdown of FACT running in a container (i.e. FACT_docker). This should solve the problems with the radare GUI in FACT_docker and resolve https://github.com/fkie-cad/FACT_docker/issues/32.

### DIFF
--- a/src/fact_base.py
+++ b/src/fact_base.py
@@ -36,9 +36,11 @@ class FactBase:
         if started_by_start_fact_py:
             signal.signal(signal.SIGUSR1, self.shutdown_listener)
             signal.signal(signal.SIGINT, lambda *_: None)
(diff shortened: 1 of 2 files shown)


maringuu:
Could you reformulate the commit message?
It would be great to explicitly mention that docker-compose sends SIGTERM.
Beside this LGTM.

Also terminating on SIGTERM is something we should have done before :sweat_smile:

jstucke:
> Could you reformulate the commit message?
It would be great to explicitly mention that docker-compose sends SIGTERM.

Happy now? :see_no_evil:

maringuu:
LGTM!

maringuu (approved):

PR #886, opened 2022-10-21
added analysis live stats

jstucke:
added analysis duration live stats to plugin cards on the system health page


dorpvom:
Things we discussed for this:
- It's a lot of numbers. I'd say we can remove variance at least. Mean and median should give enough information
- We should set a limit of calculations for the stat keeping. E.g. 2000 analyses per plugin. This would cap the memory needed for this. Counting can be done independently.

jstucke:
> It's a lot of numbers. I'd say we can remove variance at least. Mean and median should give enough information
I agree that the variance is not that useful. The standard deviation on the other hand is interesting IMHO since it shows the average deviation from the mean value. This could be interesting for finding plugins that vary a lot in their analysis times.

> We should set a limit of calculations for the stat keeping. E.g. 2000 analyses per plugin. This would cap the memory needed for this. Counting can be done independently.
I implemented a limit of 1000 and switched to faster and more efficient `multiprocessing.Array`s

src/statistic/analysis_stats.py
```diff
@@ -0,0 +1,30 @@
...
+        stats_array = np.array(plugin.analysis_stats.get_obj(), ctypes.c_float)
+        if stats_count < plugin.ANALYSIS_STATS_LIMIT:
+            stats_array = stats_array[:stats_count]
+        return dict(
+            min=_format_float(stats_array.min()),
+            max=_format_float(stats_array.max()),
+            mean=_format_float(stats_array.mean()),
+            median=_format_float(np.median(stats_array)),
+            std_dev=_format_float(stats_array.std()),
+            count=str(len(stats_array)),
```
dorpvom, line 22: Might be nit-picky but is that not `stats_count`?

dorpvom (approved):

PR #887, opened 2022-10-21
removed `register_plugin` construct

jstucke:
Removed `register_plugin` and `plugin_administrator` constructs where the scheduler class is passed to the plugin so that the plugin can write itself into the list of plugins inside the scheduler, which is both unnecessarily complicated and generally unnecessary.
Instead, this can be done directly inside the scheduler (where the plugins are initialized anyway).

### DIFF
--- a/src/scheduler/analysis.py
+++ b/src/scheduler/analysis.py
@@ -184,18 +184,7 @@ class AnalysisScheduler:  # pylint: disable=too-many-instance-attributes
                 # be missing dependencies. So if anything goes wrong we want to inform the user about it
                 logging.error(f'Could not import plugin {plugin_name} due to exception', exc_info=True)
             else:
-                plugin.AnalysisPlugin(self, config=self.config)
-
-    def register_plugin(self, name: str, plugin_instance: AnalysisBasePlugin):
-        '''
-        This function is used by analysis plugins to register themselves with this scheduler. During initialization the
-        plugins will call this functions giving their name and a reference to their object to allow the scheduler to
-        address them for running analyses.
-
-        :param name: The plugin name for addressing in runner and collector
-        :param plugin_instance: A reference to the plugin object
-        '''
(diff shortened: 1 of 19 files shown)


src/config/main.cfg
```diff
@@ -79,6 +79,9 @@ threads = 2
 [crypto_material]
 threads = 2
 
+[cve_lookup]
+threads = 3
```
maringuu, line 83: Were the changes in this file included by accident?

If not please put them in a seperate commit.

src/scheduler/analysis.py
```diff
@@ -418,7 +407,7 @@ def _start_result_collector(self):
         self.result_collector_process = ExceptionSafeProcess(target=self._result_collector)
         self.result_collector_process.start()
 
-    def _result_collector(self):  # pylint: disable=too-complex
```
maringuu, line 421: Nit: This change should also be a separate commit.

maringuu (changes requested):
Good to finally have this thing removed.
Just the two comments and this should be good to go.

src/config/main.cfg
jstucke: That was not intentional -> removed

src/scheduler/analysis.py
jstucke: I removed it because it caused this:

> src/scheduler/analysis.py:410: [bad-option-value:] Bad option value 'too-complex'

I'm not even sure why. Was this option removed in a recent pylint version?

maringuu: Hmm I dont get this error.

```
$ pylint --version
pylint 2.15.3
astroid 2.12.10
Python 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0]
```

maringuu (approved):

PR #889, opened 2022-10-25
text file comparison from url instead of session

jstucke:
- the text file diff page did not support reloading/bookmarking/etc. because the comparison UIDs were taken from the session instead of the URL
- this PR splits the initiation of a text file diff from showing the actual page to fix those problems

### DIFF
--- a/src/web_interface/components/compare_routes.py
+++ b/src/web_interface/components/compare_routes.py
@@ -1,3 +1,7 @@
+# pylint: disable=no-self-use
+
+from __future__ import annotations
+
 import difflib
 import logging
 from contextlib import suppress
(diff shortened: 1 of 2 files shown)


src/web_interface/components/compare_routes.py
```diff
@@ -128,7 +132,7 @@ def browse_comparisons(self):
     @roles_accepted(*PRIVILEGES['submit_analysis'])
     @AppRoute('/comparison/add/<uid>', GET)
     @AppRoute('/comparison/add/<uid>/<root_uid>', GET)
-    def add_to_compare_basket(self, uid, root_uid=None):  # pylint: disable=no-self-use
```
maringuu, line 131: Nit: pylint fixes/changes should be a separate commit.

maringuu (changes requested):
Could you describe the way it works now in the commit message?
I was confused for a minute and I think more documentation does not hurt.

maringuu:
LGTM! Once the CI ran this can be merged.

maringuu (approved):

PR #897, opened 2022-11-08
Update pip dependencies

jstucke:
Since [flask-restx finally released the new version] which is compatible with Flask 2.2.x we can update our frontend dependencies.
I also updated all the other dependencies where possible and fixed resulting problems (mostly tests).


src/plugins/analysis/cve_lookup/test/test_setup_repository.py
```diff
@@ -1047,7 +1047,7 @@
 ]
 
 
-@pytest.fixture(scope='session', autouse=True)
```
maringuu, line 1050: If you don't want to explicit you can omit `scope=function`. If you want to be explicit a comment might be useful to explain why it is important to have this be function scoped.

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
```diff
@@ -1107,30 +1107,28 @@ def test_exists(monkeypatch):
...
-        sr.extract_relevant_feeds(from_table='new', where_table='outdated')
-    )
+    assert sorted(sr.extract_relevant_feeds(from_table='new', where_table='outdated')) == [
+        ('CVE-2018-0002', 2018),
+        ('CVE-2018-0003', 2018),
+    ]
 
 
 def test_delete_outdated_feeds():
+    sr.DATABASE = sr.DatabaseInterface(PATH_TO_TEST + 'test_update.db')
```
maringuu, line 1117: Shouldnt monkeypatch be used for this kind of stuff?

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
```diff
@@ -1047,7 +1047,7 @@
 ]
 
 
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture(scope='function', autouse=True)
```
maringuu, line 1050: Not your code so not your fault but shouldn't this function also use monkeypatch for `sr.QUERIES.update(TEST_QUERIES)`?

src/test/integration/web_interface/test_filter.py
```diff
@@ -60,9 +60,3 @@ def test_render_analysis_tags_fix(frontend):
         output = render_analysis_tags(tags).replace('\n', '').replace('    ', ' ')
     assert 'badge-primary' in output
     assert '> wow<' in output
-
-
-def test_render_analysis_tags_bad_type():
```
maringuu, line 65: What made this test useless?

src/plugins/analysis/binwalk/requirements.txt
```diff
@@ -1,3 +1,2 @@
 capstone==4.0.2
 cstruct==4.0
-matplotlib==3.6.2
```
maringuu: I don't like removing the dependency from the plugin dependency file.

If we removed the matplotlib dependency from FACT the plugin would stop working.

src/install/requirements_common.txt
```diff
@@ -7,6 +7,7 @@ setuptools
 appdirs==1.4.4
 flaky==3.7.0
 lief==0.12.3
+matplotlib==3.5.3
```
maringuu, line 10: Grepping (`rg matplotlib`) through the code reveals that matplotlib is only ever used in `src/helperFunctions/web_interface.py` which would only be frontend.

I don't know why it is listed as backend dependency, but I think it should not be.

maringuu (changes requested):
The following modules did increase their major version:
- crpytography
- docker
- pytest-cov
- bcrypt
- flask_restx
- flask-security-too
- flask_sqlalchemy
- more_itertools
- cstruct

Its alot but did you read the changelogs?
Theoretically backwards incompatible changes that dont have test coverage by our tests could have been made.

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
jstucke: Probably. The whole module shouldn't be structured like it is but I saw it as out of scope to fix it in this PR

src/test/integration/web_interface/test_filter.py
jstucke: It aimed at provoking a ValueError by submitting an Int as key but it didn't cause an error (maybe because of the updated version). Testing this doesn't even make sense, since the tested function doesn't even handle such errors.

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
maringuu: As discussed lets leave it as it is.

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
maringuu: > shouldn't this function also use monkeypatch?

It should but we won't change this in this pr.

src/plugins/analysis/cve_lookup/test/test_setup_repository.py
jstucke: removed

src/plugins/analysis/binwalk/requirements.txt
jstucke: Yes you are probably right
restored the matplotlib dependency in binwalk and the frontend requirements file

src/install/requirements_common.txt
jstucke: I don't know but possibly because of the plugin

maringuu (approved):

PR #899, opened 2022-11-09
renamed session cookie to fix session problems

jstucke:
with other flask apps running on the same system

### DIFF
--- a/src/web_interface/app.py
+++ b/src/web_interface/app.py
@@ -20,6 +20,9 @@ def _add_config_to_app(app, config):
     app.config['SECURITY_UNAUTHORIZED_VIEW'] = '/login'
     app.config['LOGIN_DISABLED'] = not config.getboolean('expert-settings', 'authentication')
 
+    # rename session cookie to fix user session problems with other flask apps running on the same system
+    app.config['SESSION_COOKIE_NAME'] = 'FACT_session_cookie'
+
     # As we want to use ONLY usernames and no emails but email is hardcoded in
     # flask-security we change the validation mapper of 'email'.
     # Note that from the perspective of flask-security we still use emails.


rhelmke (approved):

maringuu (changes requested):
Hmm I'm not sure if this is the right way to solve the problem.
As far as I understand the problem is that multiple flask apps running on localhost all 
get the same cookie.
To fix this one could also add sth. like `127.0.0.1 fact.dev` in `/etc/hosts`
for every flask app, can't you?

The documentation of `SERVER_NAME` suggests the same solution. 

> SERVER_NAME
>
>     Inform the application what host and port it is bound to. Required for subdomain route matching support.
>
>     If set, will be used for the session cookie domain if SESSION_COOKIE_DOMAIN is not set. Modern web browsers will not allow setting cookies for domains without a dot. To use a domain locally, add any names that should route to the app to your hosts file.
>
>     127.0.0.1 localhost.dev
>
>     If set, url_for can generate external URLs with only an application context instead of a request context.
>
>     Default: None

rhelmke:
Yeah I mean the auth cookie still gets leaked using the present fix because RFC6265 states that [cookies do not provide host isolation by port](https://www.rfc-editor.org/rfc/rfc6265#section-8.5).

However, I'm against the `/etc/hosts` proposal. Someone who accesses FACT via `localhost` is also able to access **any** cookie on that machine. There is little benefit in pulling off another invasive stunt that would be our installer modifying `/etc/hosts` files.

From a security point of view, consider the following: If users can not trust other flask apps on the same host leaking browser cookies to each other, I'd rather ask why they installed them in the first place.

However, the fix by @jstucke has a functional benefit: FACT gets its custom named cookie for authentication on the same origin. Logging in to other flask apps does not interfere with the user authentication state in FACT (and vice versa).

maringuu:
> There is little benefit in pulling off another invasive stunt that would be our installer modifying /etc/hosts files.

My suggestion is not to add this to the installer.
Imo it is the responsibility of the user/sysadmin to ensure that different applications on the host get different hostnames.
Theoretically another flask app could name the cookie the same way as we do (`FACT_session_cookie`).

jstucke:
I think @maringuu is right that in a production setup if you wanted to run multiple flask apps on the same host, then you should use subdomains. That doesn't mean that this is mutually exclusive with renaming the session cookie, though. The latter can be helpful to avoid collisions for users that don't want to mess with editing their hosts file just to be able to run FACT and another app on their system at the same time (as it was the case for us recently). FACT is not only run by sys admins :sweat_smile:

rhelmke:
> My suggestion is **not** to add this to the installer.
> Imo it is the responsibility of the user/sysadmin to ensure that different applications on the host get different hostnames.
> Theoretically another flask app could name the cookie the same way as we do (`FACT_session_cookie`).

Oh now I understand. @jstucke and me just had a talk which is basically summed up by his previous comment in this PR. We both agree that renaming the cookie is basically a measure to avoid session collisions in the first place. As not every user is also a developer and/or power user, we chose to still merge this PR.

I hope you're fine with that.

maringuu:
> FACT is not only run by sys admins

True. Maybe we should prioritize fleshing out FACT_docker and point "normal" users to that.

PR #900, opened 2022-11-09
added github action for ci

jstucke:


rhelmke (approved):
*slaps on roof of CI server*

This bad boy can fit in so many actions.

*kernel panics*

PR #908, opened 2022-11-18
adjusted PW regex for saltless unix PWs and extended john wordlist

jstucke:
- adjusted the Unix passwd regex so that passwords with empty salt are found
- extended the wordlist of john which was missing ssh and telnet default passwords

### DIFF
--- a/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
+++ b/src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
@@ -21,7 +21,7 @@ WORDLIST_PATH = Path(get_src_dir()) / 'bin' / 'passwords.txt'
 USER_NAME_REGEX = br'[a-zA-Z][a-zA-Z0-9_-]{2,15}'
 UNIX_REGEXES = [
     USER_NAME_REGEX + br':[^:]?:\d+:\d*:[^:]*:[^:]*:[^\n ]*',
-    USER_NAME_REGEX + br':\$[1256][ay]?\$[a-zA-Z0-9\./+]+\$[a-zA-Z0-9\./+]{16,128}={0,2}',  # MD5 / Blowfish / SHA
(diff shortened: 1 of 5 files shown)


src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -21,7 +21,7 @@
 USER_NAME_REGEX = br'[a-zA-Z][a-zA-Z0-9_-]{2,15}'
 UNIX_REGEXES = [
     USER_NAME_REGEX + br':[^:]?:\d+:\d*:[^:]*:[^:]*:[^\n ]*',
-    USER_NAME_REGEX + br':\$[1256][ay]?\$[a-zA-Z0-9\./+]+\$[a-zA-Z0-9\./+]{16,128}={0,2}',  # MD5 / Blowfish / SHA
+    USER_NAME_REGEX + br':\$[1256][ay]?\$[a-zA-Z0-9\./+]*\$[a-zA-Z0-9\./+]{16,128}={0,2}',  # MD5 / Blowfish / SHA
```
maringuu, line 24: These regexes deserve some documentation. Despite the commit message it is not clear to me why this change makes it possible to leave out the salt.
Maybe we should at least document that they are used to parse entrys to `/etc/passwd`.

jstucke: > Maybe we should at least document that they are used to parse entrys to /etc/passwd.

that is what `UNIX_REGEXES` should imply but we could also rename it to `UNIX_PASSWD_REGEXES` to make it more clear

maringuu: >  rename it to UNIX_PASSWD_REGEXES to make it more clear

What about `ETC_PASSWD_REGEXES`?
But I dont have a strong opinion on that. Go ahead and merge if you think this is good to go.

maringuu (approved):

PR #909, opened 2022-11-21
applied new file tree theme to comparison page

jstucke:
applied the new file tree theme from the analysis page to the comparison page


maringuu:
LGTM! Just one question before merging: Would is somehow be possible to do this in a way that we only have to change one thing to affect both places?
This prevents oversights like this in the future.

maringuu:
What about adding a comment to both places reminding us to not forget the other one?

jstucke:
I looked at the code again and found enough common parts to justify a common module. I think the code also is much clearer now.

maringuu (approved):
LGTM. CI failure is unrelated. I probably still will wait for it to be fixed before merging.

PR #910, opened 2022-11-25
added no_text_file to software_components signatures where missing

### LINKED ISSUE #404: False positive software results
Some software signatures that target binary files miss the `no_text_file` flag (e.g. threadx, mOS, fireOS). This leads to false positives from license files a.s.o.

jstucke:
- added no_text_file to software_components signatures where it was missing
- fixed errors in miniigd (wrong regex) and sqlite3 (wrong logic) signatures

resolves #404


maringuu:
Maybe its obvious but I think we should put the motivation behind ignoring text files to the commit message.

maringuu (changes requested):
See comment above.

jstucke:
Honestly i don't think that is necessary. We are looking for binary executables (and libraries in some cases) and text files should always be false positives. I think the only exception to this is jQuery.

maringuu (approved):
LGTM!

PR #912, opened 2022-11-25
file diff optimization and bugfixes

jstucke:
- moved file_diff functionality from the frontend to the backend (where the files are located)
    - this way only one intercom task is needed instead of two and also less data is transmitted (the diff string instead of the contents of the two files)
    - added intercom listener for that
-  adjusted intercom response poll delay to stop wasting lots of time
- file_diff load times should be a lot faster (at least twice as fast)
- fixed and added tests
- added escape for three elements that caused errors in the file_diff template:
    - str format variables (`${var}`)
    - backslashes
    - backticks

### DIFF
--- a/src/web_interface/components/compare_routes.py
+++ b/src/web_interface/components/compare_routes.py
@@ -194,7 +194,6 @@ class CompareRoutes(ComponentBase):
             diff_str = intercom.get_file_diff((uid_1, uid_2))
         if diff_str is None:
             return render_template('compare/error.html', error='File(s) not found.')
-        diff_str = diff_str.replace('\\', '\\\\').replace('`', '\\`').replace('${', '\\$\\{')
 
         return render_template(
             'compare/text_files.html', diffstr=diff_str, hid0=diff_files[0].fw_hid, hid1=diff_files[1].fw_hid
--- a/src/web_interface/components/compare_routes.py
+++ b/src/web_interface/components/compare_routes.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import difflib
 import logging
 from contextlib import suppress
 from typing import NamedTuple, Optional
@@ -24,8 +23,6 @@ from web_interface.security.privileges import PRIVILEGES
 
 class FileDiffData(NamedTuple):
     uid: str
-    content: str
-    file_name: str
     mime: str
     fw_hid: str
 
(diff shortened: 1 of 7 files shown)


src/web_interface/components/compare_routes.py
```diff
@@ -192,34 +190,24 @@ def compare_text_files(self, uid_1: str, uid_2: str, root_uid_1: str | None = No
                 error=f'Can\'t compare non-text mimetypes. ({diff_files[0].mime} vs {diff_files[1].mime})',
             )
 
-        diffstr = self._get_file_diff(*diff_files)
+        with ConnectTo(self.intercom, self._config) as intercom:
+            diff_str = intercom.get_file_diff((uid_1, uid_2))
+        if diff_str is None:
+            return render_template('compare/error.html', error='File(s) not found.')
+        diff_str = diff_str.replace('\\', '\\\\').replace('`', '\\`').replace('${', '\\$\\{')
```
maringuu, line 197: Isn't there another method to do the replacements?
Is this maintainable to replace everything ourselfes?

I had a look at the javascript which apparently I wrote some time ago and you patched a bug in it.
How does this line work? 
``` $('<textarea />').html(`{{ diffstr }}`).text(); ```
I fail to find the documentation.

Also it seems that I used backticks for enclosing the `{{ diffstr }}`. I'm unsure why. Maybe to avoid having to replace quotes in the diffstring.
I really have little experience in html/js programming but isn't there a way to do this?
Should we maybe have a hidden html element that contains the diffstring? This way we would not have to do any escaping ourselfs but jinja will do it for us.

**Review by maringuu — COMMENTED** [review]
Code overall lgtm. We should really use a better escaping mechanism for the diff.
I hope my idea in the comment is suitable.

File: src/web_interface/components/compare_routes.py (hunk 1 above)
Comment (jstucke) [reply src/web_interface/components/compare_routes.py:197]: Actually I think I have found a robust (and maintainable) way to do this: the jinja filter "tojson" also escapes characters so passing this as `{{ diffstr | tojson }}` to the javascript code seems to work every time (at least everything that I tried that was problematic before). The `$('<textarea />')` construct was used for escaping html but it seems we also don't need that anymore.

File: src/web_interface/templates/compare/text_files.html [hunk 2]
```diff
@@ -21,7 +21,7 @@
 <script>
 document.addEventListener('DOMContentLoaded', () => {
     // This string is made safe by the python code rendering this template
```
maringuu, line 23: This comment should be removed.

maringuu (changes requested):

src/web_interface/components/compare_routes.py
maringuu: Good find!

src/web_interface/templates/compare/text_files.html
jstucke: Removed. That is the problem with having many comments: they also need maintenance :sweat_smile:

maringuu (approved):

PR #921, opened 2022-12-08
cve_lookup: bugfix for InvalidVersion in new packaging version

jstucke:
- fixed a bug where a new version of packaging could lead to an uncaught exception in the cve_lookup plugin
- also fixes a test


rhelmke (approved):
LGTM!

PR #941, opened 2022-12-29
style: Use python 3.10 typehints

maringuu:
Automatically done with the following shell commands:

```sh
pyupgrade --py310-plus $(git ls-tree -r --name-only master| rg "\.py$")

for file in $(git diff --name-only); do
    printf '%s\n%s\n' "from __future__ import annotations" "$(cat $file)" > $file
done

autoflake --remove-all-unused-imports -i $(git diff --name-only --staged)

python3 -m isort (git diff --name-only --staged)

black $((git diff --name-only --staged))
```

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -9,7 +9,7 @@ from werkzeug.local import LocalProxy
 
 # pylint: disable=invalid-name
 _cfg = None
-cfg: 'Config' = LocalProxy(lambda: _cfg)
+cfg: Config = LocalProxy(lambda: _cfg)
 
 _configparser_cfg = None
 # For transitioning between using ConfigParser and this module.
--- a/src/plugins/analysis/qemu_exec/docker/start_binary.py
+++ b/src/plugins/analysis/qemu_exec/docker/start_binary.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
+from __future__ import annotations
+
 import subprocess
 import sys
 from base64 import b64encode
 from json import dumps
-from typing import Tuple
 
 FIRMWARE_ROOT = '/opt/firmware_root/'
 ARCH = sys.argv[1]
@@ -11,7 +12,7 @@ FILE_PATH = FIRMWARE_ROOT + sys.argv[2]
 TIMEOUT_ERROR_EXIT_CODES = [124, 128 + 9]
 
 
-def get_output_error_and_return_code(command: str) -> Tuple[bytes, bytes, int]:
+def get_output_error_and_return_code(command: str) -> tuple[bytes, bytes, int]:
     process = subprocess.run(command, capture_output=True, shell=True)
     return process.stdout, process.stderr, process.returncode
 
--- a/src/plugins/installer.py
+++ b/src/plugins/installer.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
+
 import os
 from pathlib import Path
-from typing import Optional
 
 from helperFunctions.install import (
     check_distribution,
@@ -29,7 +30,7 @@ class AbstractPluginInstaller:
     #: Must be overwritten by a class variable of a child class
     base_path = None
 
-    def __init__(self, distribution: Optional[str] = None, skip_docker: bool = _skip_docker_env):
+    def __init__(self, distribution: str | None = None, skip_docker: bool = _skip_docker_env):
         self.distribution = distribution or check_distribution()
         self.build_path = self.base_path / 'build'
         self.skip_docker = skip_docker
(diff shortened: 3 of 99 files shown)


maringuu:
Tbh I don't understand why either the jammy ci nor the buster ci succeeded.
I'm leaving this open if anyone else wants to investigate.

jstucke:
> Tbh I don't understand why either the jammy ci nor the buster ci succeeded.
I'm leaving this open if anyone else wants to investigate.

As described in [PEP 563](https://peps.python.org/pep-0563/), `from __future__ import annotations` will postpone the evaluation of type hints. This means that you can use Python 3.9 or 3.10 syntax in the type hints even if you are using Python 3.7. If you look at the errors in the CI build you will notice that they did not occur inside a type hint, but rather inside an expression:

```python3
     Stats = list[tuple[str, int]]
E   TypeError: 'type' object is not subscriptable
```

This means that statements like `Stats = list[tuple[str, int]]` still need to use valid Python 3.7 syntax, i.e. `Stats = List[Tuple[str, int]]`.
We could also replace `Stats` (and possible other occurrences) with the type hint but IMHO it improved readability in the module.

src/config.py
```diff
@@ -9,7 +9,7 @@
 
 # pylint: disable=invalid-name
 _cfg = None
-cfg: 'Config' = LocalProxy(lambda: _cfg)
+cfg: Config = LocalProxy(lambda: _cfg)
```
jstucke, line 12: There was a reason that this was a string -- does this cause any errors?

src/plugins/analysis/qemu_exec/docker/start_binary.py
```diff
@@ -1,17 +1,18 @@
+from __future__ import annotations
```
jstucke, line 1: Flake8 complains quite rightly that the shebang should be at the top of the file

src/storage/db_interface_stats.py
```diff
@@ -10,8 +13,8 @@
 from storage.db_interface_base import ReadOnlyDbInterface, ReadWriteDbInterface
 from storage.schema import AnalysisEntry, FileObjectEntry, FirmwareEntry, StatsEntry
 
-Stats = List[Tuple[str, int]]
-RelativeStats = List[Tuple[str, int, float]]  # stats with relative share as third element
+Stats = list[tuple[str, int]]
+RelativeStats = list[tuple[str, int, float]]  # stats with relative share as third element
```
jstucke, line 17: This needs to be fixed to make it Python 3.7/3.8 compatible

src/web_interface/frontend_database.py
```diff
@@ -12,13 +12,13 @@
...
-        template: Optional[Type[ViewReader]] = None,
-        stats_viewer: Optional[Type[StatsDbViewer]] = None,
-        stats_updater: Optional[Type[StatsUpdateDbInterface]] = None,
+        frontend: type[FrontEndDbInterface] | None = None,
+        editing: type[FrontendEditingDbInterface] | None = None,
+        admin: type[AdminDbInterface] | None = None,
+        comparison: type[ComparisonDbInterface] | None = None,
+        template: type[ViewReader] | None = None,
+        stats_viewer: type[StatsDbViewer] | None = None,
+        stats_updater: type[StatsUpdateDbInterface] | None = None,
```
jstucke, line 21: Nice I didn't know you could do that!

jstucke (changes requested):
Pretty cool that pyupgrade can do that. Only a couple of places that need to be fixed manually

src/web_interface/frontend_database.py
maringuu: Me neither, but pyupgrade does :)

maringuu:
> As described in [PEP 563](https://peps.python.org/pep-0563/), from __future__ import annotations will postpone the evaluation of type hints

Oh that makes sense. I didn't think of the failing things to be expressions rather than typehints.
Thanks for the explanation!

src/config.py
maringuu: I think putting the typehint in quotes kind of does what PEP 563 does. Previously it failed because the class was declared after the typehint iirc.
Lets see how the ci does.

maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/3844813134

maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/3845218736

jstucke (commented):
the CI still fails:

> src/test/integration/conftest.py:101: 
           NameError: name 'MockDataStorage' is not defined

Does this have something to do with this?
> FIXME remove this. The ci fails otherwise

Also, there seems to be a merge conflict

maringuu:
The fixme is removed, the ci passes and I rebased it on master.

src/plugins/installer.py
```diff
@@ -29,7 +29,7 @@ class AbstractPluginInstaller:
     #: Must be overwritten by a class variable of a child class
     base_path = None
 
-    def __init__(self, distribution: Optional[str] = None, skip_docker: bool = _skip_docker_env):
+    def __init__(self, distribution: str | None = None, skip_docker: bool = _skip_docker_env):
```
jstucke, line 32: > 'typing.Optional' imported but unused (F401)

`from typing import Optional` seems to be unused now

jstucke (approved):

PR #942, opened 2023-01-02
installer: Documentation AbstractPluginInstaller

maringuu:
The assumption about the cwd was not clear.
Before this commit running `./src/install.py --backend-docker-images` would not succeed because of the relative path in the software_components plugin.

### DIFF
--- a/src/plugins/installer.py
+++ b/src/plugins/installer.py
@@ -11,21 +11,31 @@ from helperFunctions.install import (
 
 
 class AbstractPluginInstaller:
+    """A class that is used to handle plugin installation.
+    Any class subclassing it may overwrite any public method except :py:func:`install`.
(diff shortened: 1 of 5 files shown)


src/plugins/installer.py
```diff
@@ -11,6 +11,12 @@
 
 
 class AbstractPluginInstaller:
+    """A class that is used to handle plugin installation.
+    Any class subclassing it may overwrite any public method except :py:func:`install`.
```
jstucke, line 15: The reference to the method is probably not that useful if we don't add this file to the sphinx docs

jstucke (approved):

maringuu:
Added to the sphinx docs and added some small cleanup commits.

jstucke (approved):
Looks good to me

PR #943, opened 2023-01-02
plugins/analysis/input_vectors: Fix test name

maringuu:
The Tests were not actually run before as pytest only auto-discovers tests in classes prefixed with `Test`. [1]

[1] https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#test-discovery


jstucke (approved):

jstucke:
Wow the tests even work (even though they apparently never ran on the CI?)

PR #945, opened 2023-01-02
Ghidra base image update

jstucke:
- updated the ghidra headless base image to be in sync with the cwe_checker container (so that only one version of the base image gets installed) (see also https://github.com/fkie-cad/docker_ghidra_headless_base/pull/8)
- [x] merging should be postponed until the new stable version of the CWE checker is released (see https://github.com/fkie-cad/cwe_checker/issues/367)
- added the Mint 20.3 codename that was missing from the installation


maringuu:
https://github.com/fkie-cad/cwe_checker/releases/tag/v0.7

maringuu (approved):
lgtm!

PR #949, opened 2023-01-05
File tree mime icons

jstucke:
- improved file tree icons with support for many more MIME types
- ~~needs `elementary-xfce-icon-theme` package, which is available from Ubuntu 20.04 and Debian "buster" onwards~~
    - ~~switched to obsidian icon theme (from GitHub) for universal support~~
    - switched again, this time to papirus-icon-theme after intensive user study (and because it is available as deb package on all supported systems)
      - it would probably be a good idea to squash the commits when merging this one :see_no_evil:

### DIFF
--- a/src/web_interface/file_tree/file_tree.py
+++ b/src/web_interface/file_tree/file_tree.py
@@ -8,62 +8,61 @@ from web_interface.file_tree.file_tree_node import FileTreeNode
 
 WEB_BASE_PATH = Path(__file__).parent.parent
 ICON_PATH = WEB_BASE_PATH / 'static/file_icons/mimetypes'
-TYPE_TO_PATH = {p.stem: f'/{p.relative_to(WEB_BASE_PATH)}' for p in ICON_PATH.iterdir()}
-CRYPTO = 'application-certificate'
-CONFIG = 'text-x-makefile'
-TYPE_TO_PATH.update(
+MIME_TO_ICON_PATH = {p.stem: f'/{p.relative_to(WEB_BASE_PATH)}' for p in ICON_PATH.iterdir()}
+MIME_TO_ICON_PATH.update(
     {
-        # MIME types
-        'application-x-pie-executable': TYPE_TO_PATH['application-x-executable'],
-        'application-x-dosexec': TYPE_TO_PATH['application-x-ms-dos-executable'],
+        'application-x-pie-executable': MIME_TO_ICON_PATH['application-x-executable'],
+        'application-x-dosexec': MIME_TO_ICON_PATH['application-x-ms-dos-executable'],
         'linux-device-tree': '/static/file_icons/firmware.svg',
-        # file suffixes
-        'c': TYPE_TO_PATH['text-x-csrc'],
-        'cert': TYPE_TO_PATH[CRYPTO],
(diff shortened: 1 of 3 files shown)


src/install/frontend.py
```diff
@@ -120,6 +122,18 @@ def _install_docker_images(radare):
         raise InstallationError(f'Failed to pull pdf report container:\n{docker_process.stdout}')
 
 
+def _copy_mime_icons():
```
maringuu, line 125: Could we maybe symlink the directory rather than copying it?

src/web_interface/file_tree/file_tree.py
```diff
@@ -1,64 +1,101 @@
...
 from typing import Dict, Iterable, List, NamedTuple, Optional, Set
 
 from web_interface.file_tree.file_tree_node import FileTreeNode
 
+WEB_BASE_PATH = Path(__file__).parent.parent
+ICON_PATH = WEB_BASE_PATH / 'static/file_icons/mimetypes'
+TYPE_TO_PATH = {p.stem: f'/{p.relative_to(WEB_BASE_PATH)}' for p in ICON_PATH.iterdir()}
+CRYPTO = 'application-certificate'
+CONFIG = 'text-x-makefile'
+TYPE_TO_PATH.update(
```
maringuu, line 14: I'm confused why exactly we need this.
Shouldn't a mime type map to one icon?

src/web_interface/file_tree/file_tree.py
```diff
@@ -71,23 +108,42 @@ class FileTreeData(NamedTuple):
...
+def get_icon_for_file(mime_type: Optional[str], file_name: str) -> str:
     '''
-    Retrieve the path to appropriate icon for a given mime type. The icons are located in the static folder of the
-    web interface and the paths therefore start with "/static". Archive types all receive the same icon.
+    Retrieve the path to the appropriate icon for a given mime type and file name. The icons are located in the static
+    folder of the web interface and the paths therefore start with "/static". Archive types all receive the same icon.
 
-    :param mime_type: The MIME type of a file (in the file tree).
+    :param mime_type: The MIME type of the file (in the file tree).
+    :param file_name: The file name.
```
maringuu, line 117: Like the comment above I'm confused why we need the filename.
From my understanding it should be like this:
If we don't know the mimetype figure it out by the filename.
Then use `get_icon(mimetype)` to get the icon?

Is there a reason we need the mimetype AND the filename?

src/web_interface/file_tree/file_tree.py
jstucke: see below

src/install/frontend.py
jstucke: I don't think that this would be a good idea (for safety reasons) even if it works (I'm not that proficient with path traversal attacks though so I can't say for sure)

maringuu (commented):
> it would probably be a good idea to squash the commits when merging this one

Yeah, please suqash in commits that make sense to you.

Aside this lgtm.

jstucke:
> Yeah, please suqash in commits that make sense to you.

My idea was to simply use the "Squash and merge" button if that is OK :sweat_smile:

maringuu (approved):
LGTM!

PR #952, opened 2023-01-09
ipc plugin summary bugfix

jstucke:
- bugfix for empty results in ipc plugin summary
- changed summary to use result entry types as items


maringuu (approved):
LGTM!

PR #953, opened 2023-01-09
Optimized shutdown

jstucke:
- replaced killing of the whole process group with only stopping remaining subprocesses
- handled shutdown of `multiprocessing.Manager` correctly
- only log reception of shutdown signal in main process (to reduce logging spam during shutdown)

### DIFF
--- a/src/fact_base.py
+++ b/src/fact_base.py
@@ -29,10 +29,14 @@ class FactBase:
 
     def __init__(self):
         self.run = True
+        self.args = program_setup(self.PROGRAM_NAME, self.PROGRAM_DESCRIPTION, self.COMPONENT)
+        self._register_signal_handlers()
+        self.work_load_stat = WorkLoadStatistic(component=self.COMPONENT)
+
+    def _register_signal_handlers(self):
         # Check whether the process was started by start_fact.py
         parent = ' '.join(psutil.Process(os.getppid()).cmdline())
         started_by_start_fact_py = 'start_fact.py' in parent or 'start_all_installed_fact_components' in parent
-
         if started_by_start_fact_py:
             signal.signal(signal.SIGUSR1, self.shutdown_listener)
             signal.signal(signal.SIGINT, lambda *_: None)
@@ -42,11 +46,9 @@ class FactBase:
             signal.signal(signal.SIGINT, self.shutdown_listener)
             signal.signal(signal.SIGTERM, self.shutdown_listener)
 
-        self.args = program_setup(self.PROGRAM_NAME, self.PROGRAM_DESCRIPTION, self.COMPONENT)
-        self.work_load_stat = WorkLoadStatistic(component=self.COMPONENT)
-
     def shutdown_listener(self, signum, _):
-        logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
+        if self._is_main_process():
+            logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
         self.run = False
 
     def shutdown(self):
@@ -61,3 +63,7 @@ class FactBase:
             if self.args.testing:
                 break
         self.shutdown()
+
+    @staticmethod
+    def _is_main_process() -> bool:
+        return os.getpid() == os.getpgrp()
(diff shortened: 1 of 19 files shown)


src/fact_base.py
```diff
@@ -42,11 +46,9 @@ def __init__(self):
...
             signal.signal(signal.SIGTERM, self.shutdown_listener)
 
-        self.args = program_setup(self.PROGRAM_NAME, self.PROGRAM_DESCRIPTION, self.COMPONENT)
-        self.work_load_stat = WorkLoadStatistic(component=self.COMPONENT)
-
     def shutdown_listener(self, signum, _):
-        logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
+        if self._is_main_process():
+            logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
         self.run = False
```
maringuu, line 52: If I'm not mistaken setting `self.run` for any other process than the main process does not do anything.

My suggestion is to change the listener to
```python
def shutdown_listener(self, signum, _):
    if not self._is_main_process():
        # Comment about why we ignore the signal, but still have the handler registered.
        return
    logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
    self.run = False
```

src/helperFunctions/process.py
```diff
@@ -21,8 +23,35 @@ def complete_shutdown(message: Optional[str] = None) -> None:
...
+            future.result()  # call result to make sure all threads are finished and there are no exceptions
+
+
+def _stop_process_by_pid(pid: int):
+    try:
+        proc = psutil.Process(pid)
+        try:
+            proc.wait(5)
+        except psutil.TimeoutExpired as err:
+            logging.warning(f'Timeout while waiting for process to shut down: {err} -> kill')
```
maringuu, line 50: Maybe add some more information to this log entry.
I'm not sure what exactly would be helpful, but it would be great to somehow be able to determine what process exactly was killed.

src/helperFunctions/process.py
```diff
@@ -155,12 +184,22 @@ def new_worker_was_started(new_process: ExceptionSafeProcess, old_process: Excep
...
-        if process.is_alive():
-            process.terminate()
+        thread = Thread(target=stop_process, args=(process, timeout))
+        thread.start()
+        thread_list.append(thread)
+    for thread in thread_list:
+        thread.join()
+
+
+def stop_process(process: Process, timeout: float = 10.0):
```
maringuu, line 201: Should we maybe reconsider the timeout?
Does it really take that long for a process to exit sucessfully?
Are there processes that do that much cleanup?

src/test/acceptance/base.py
```diff
@@ -58,30 +58,31 @@ def tearDown(self):
         gc.collect()
 
     def _stop_backend(self):
-        with ThreadPoolExecutor(max_workers=4) as pool:
+        with ThreadPoolExecutor(max_workers=5) as pool:
```
maringuu, line 61: I'm really not a fan of using a ThreadPoolExecutor to just call multiple methods at once.
Imo the shutdown methods should not be blocking.

But there is nothing to change here, I just wanted to say that I don't like this solution.

maringuu (changes requested):

src/test/acceptance/base.py
jstucke: > Imo the shutdown methods should not be blocking.

The shutdown method needs to block to give the worker processes (which run as independent subprocesses) time to do a clean shutdown. I don't see how we could avoid this (except for simply killing the processes which we don't want).

src/helperFunctions/process.py
jstucke: We could print the pid but I don't think that this would really be useful after we killed it :sweat_smile: 
Sadly there seems to be no easy way to find out what the process was actually doing when we try to stop it

src/helperFunctions/process.py
maringuu: > We could also make it configurable.

I don't think that is necessary. Why should the user know when to kill the processes?

> `max(cfg.expert_settings.intercom_poll_delay, cfg.expert_settings.block_delay) + 1`

Sounds good. If you like, you could add a comment about why we chose this timeout since it isn't obvious imo.

maringuu (approved):
LGTM!

PR #955, opened 2023-01-13
yara plugin system version bugfix

jstucke:
the system version in `YaraBasePlugin` is no valid version according to [PEP 440](https://peps.python.org/pep-0440/) because it contains an underscore. This in turn causes an `InvalidVersion` exception in newer versions of packaging. This PR replaces the underscore with a dash (which makes it valid). Also fixes this in the analysis scheduler where old versions from the DB are checked (which may still contain underscores).

### DIFF
--- a/src/scheduler/analysis.py
+++ b/src/scheduler/analysis.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Queue, Value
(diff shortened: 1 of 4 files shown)


src/scheduler/analysis.py
```diff
@@ -487,3 +493,9 @@ def check_exceptions(self) -> bool:
             if plugin.check_exceptions():
                 return True
         return check_worker_exceptions([self.schedule_process, self.result_collector_process], 'Scheduler')
+
+
+def _fix_system_version(system_version: str | None) -> str:
```
maringuu, line 498: If I understand correctly this function is supposed to account for outdated database entrys that use the underscore instead of the hypen.

Shouldn't this be fixed in the database?

maringuu (changes requested):

maringuu (approved):

PR #956, opened 2023-01-13
fix slow pytest test discovery by ignoring more paths

jstucke:
a fully populated installation directory with (compose-env, john, node_modules, etc.) can make pytest test discovery slow -> ignore more paths

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,3 +40,16 @@ known_first_party = ["analysis", "compare", "helperFunctions", "install", "inter
     "statistic", "storage", "test", "unpacker", "version", "web_interface"]
 known_third_party = "docker"
 profile = "black"
+
+[tool.pytest.ini_options]
+addopts = "-v"
+testpaths = [
+    "src/test",
+    "src/plugins/**/test",
+]
+markers = [
+    "AnalysisPluginClass: Set the Class of the analysis_plugin fixture",
+    "plugin_init_kwargs: Keyword arguments to pass to AnalysisPluginClass",
+    "plugin_start_worker: Whether or not to start actual workers",
+    "cfg_defaults: Overwrite defaults for the testing config",
+]
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,9 +1,0 @@
-[pytest]
-addopts = -v
-norecursedirs = */cwe_checker/internal/*
-
-markers =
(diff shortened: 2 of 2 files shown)


maringuu:
What about setting explicitly what directorys shall be scanned?

```patch
diff --git a/pytest.ini b/pytest.ini
index 669137e2..cea27da3 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,8 @@
 [pytest]
 addopts = -v
-norecursedirs = */cwe_checker/internal/*
+testpaths =
+    src/test
+    src/plugins
 
 markers =
     AnalysisPluginClass: Set the Class of the analysis_plugin fixture

```


Here are the results of `hyperfine --warmup 1 'pytest --collect-only'`

### On master
```
Benchmark 1: pytest --collect-only
  Time (mean ± σ):      4.976 s ±  0.368 s    [User: 4.925 s, System: 1.117 s]
  Range (min … max):    4.637 s …  5.847 s    10 runs

```

### This PR
```
Benchmark 1: pytest --collect-only
  Time (mean ± σ):      3.558 s ±  0.133 s    [User: 3.566 s, System: 1.051 s]
  Range (min … max):    3.291 s …  3.723 s    10 runs

```

### My approach
```
Benchmark 1: pytest --collect-only
  Time (mean ± σ):      2.995 s ±  0.255 s    [User: 3.042 s, System: 1.007 s]
  Range (min … max):    2.681 s …  3.451 s    10 runs
```

maringuu (changes requested):
See above.

jstucke:
Good catch!
I forgot that we moved the John installation into the container because it was still there in my old FACT installation :sweat_smile: 
The only downside is that `testpaths` will only work
> [when executing from the root directory](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-testpaths)

:facepalm:

maringuu:
>  The only downside is that testpaths will only work [when executing from the root directory](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-testpaths)

Well that's unfortunate but I don't think that matters too much.
If you are not in the root directory you probably are already in one of the paths listed in testpaths.

maringuu:
Please rebase on master & suqash the commits to make the ci run again.

maringuu (approved):

PR #959, opened 2023-01-15
storage/db_connection: Fix engine url

### LINKED ISSUE #958: Postgres url fails when password contains `@`
I was running the DB migration, and it kept failing. After tracking down the problem, it led me to my password, `testest@r1`(not my real password but a comparative example). The error output i got was `swiper@factor:/FACT_core$ sudo python3 src/migrate_db_to_postgresql.py
ERROR:root:Database error when trying to read from the database: (psycopg2.OperationalError) could not translate host name "r1@localhost" to address: Name or service not known`. I believe the issue is in the `DbConnection` class in `src/storage/db_connection.py`. The line responsible for this is `engine_url = f'postgresql://{user}:{password}@{address}:{port}/{database}'`. My workaround was putting `password = password.replace("@","%40")`. I know that's not a great fix but i just needed it to work tbh.

maringuu:
We did not properly escape special characters.
This is exactly what `sqlalchemy.URL` is for.

Closes #958

### DIFF
--- a/src/storage/db_connection.py
+++ b/src/storage/db_connection.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from sqlalchemy import create_engine
+from sqlalchemy.engine import URL
 from sqlalchemy.orm import sessionmaker
 
 from config import cfg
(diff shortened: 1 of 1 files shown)


jstucke (commented):
Looks good but it seems we need to fix the CI/installation first to test this

maringuu:
> Looks good but it seems we need to fix the CI/installation first to test this

My bad. I was looking at the sqlalchemy 2 docs.
The import path was wrong. Fixed.

jstucke (approved):

PR #962, opened 2023-01-18
web_interface: Fix analysis presets

maringuu:
pydantic populates the class `config.data_storage`. Iterating over it will yield tuples rather than the dictionary keys as before.


jstucke (approved):
LGTM. I tested it and it seems to work again!

PR #963, opened 2023-01-18
missing files in file tree w/o root_uid fix

jstucke:
There can still be files missing from the file tree on the analysis page under certain circumstances:
- if no root_uid is provided (e.g. when coming to the analysis page from a file search)
- and if the file in the file tree is also contained in another FW (i.e. if there are multiple keys in the virtual file path dict)
- and if any of the other FWs does not contain the parent file in the file tree we are looking at
because the root_uid is selected at random if it is missing. 

This PR adds an "educated guess" to make sure that we select a root_uid from a FW that contains the parent file from the file tree we are currently looking at.

### DIFF
--- a/src/web_interface/file_tree/file_tree.py
+++ b/src/web_interface/file_tree/file_tree.py
@@ -148,11 +148,22 @@ class VirtualPathFileTree:
 
     def __init__(self, root_uid: str, parent_uid: str, fo_data: FileTreeData, whitelist: list[str] | None = None):
         self.uid = fo_data.uid
-        self.root_uid = root_uid if root_uid else list(fo_data.virtual_file_path)[0]
+        self.root_uid = root_uid if root_uid else self._find_root_uid(parent_uid, fo_data)
         self.parent_uid = parent_uid
         self.fo_data: FileTreeData = fo_data
         self.whitelist = whitelist
         self.virtual_file_paths = self._get_virtual_file_paths()
+
+    @staticmethod
+    def _find_root_uid(parent_uid: str, fo_data: FileTreeData) -> str:
+        '''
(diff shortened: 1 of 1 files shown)


maringuu (approved):
I don't really get the problem here tbh, I'm not that deep into file_tree.py.
But since its just very few loc I think its fine if you think the code is good.

src/web_interface/file_tree/file_tree.py
```diff
@@ -148,12 +148,23 @@ class VirtualPathFileTree:
...
+    @staticmethod
+    def _find_root_uid(parent_uid: str, fo_data: FileTreeData) -> str:
+        '''
+        If we don't have a rood_uid, we must find a root_uid that contains the parent_uid (we can't just take a
+        random one because then the files could be missing from the file tree).
+        '''
+        for root_uid, vfp_list in fo_data.virtual_file_path.items():
+            if any(parent_uid in vfp for vfp in vfp_list):
+                return root_uid
+        return list(fo_data.virtual_file_path)[0]  # safety fallback: this should not occur under normal circumstances
```
maringuu, line 166: If it "should not occur under normal circumstances" can't we just throw an exception?

What are unnormal circumstances where this could happen?

jstucke: I don't think we should let it crash at this point but maybe logging an error or warning would be appropriate (but this could also be a new source for spam in the logs so I'm hesitant). Also I'm not sure when this could happen. Maybe if you look at a firmware while deleting it or something like that :sweat_smile:

jstucke: If you look at the requirements that I described that are needed for this to happen, it was a rather rare case already.

PR #965, opened 2023-01-20
fix for older versions of the icon-theme not providing a inode-symlink icon

jstucke:
older versions (<jammy) of the papirus-icon-theme package don't include a dedicated inode-symlink icon, which leads to symlinks being displayed as empty files. This PR fixes this and uses the emblem link icon instead (which is included in all versions from [bionic](https://packages.ubuntu.com/bionic/all/papirus-icon-theme/filelist) to [lunar](https://packages.ubuntu.com/lunar/all/papirus-icon-theme/filelist))

If the icons are already installed, the `src/web_interface/static/file_icons/mimetypes` folder needs to be deleted for the installation to replace the files.

Also removes an unused function that slipped through the review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ docsrc/_build/
 docs/.buildinfo/
 src/bin
 src/web_interface/static/node_modules
+src/web_interface/static/file_icons
 
 # pytest
 .pytest_cache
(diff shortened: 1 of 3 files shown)


maringuu:
`src/web_interface/static/file_icons` is not in  `.gitignore` can we also fix this in this PR?

jstucke:
> `src/web_interface/static/file_icons` is not in `.gitignore` can we also fix this in this PR?

Good idea! I will add it

maringuu (approved):

PR #970, opened 2023-01-24
improved cve_lookup version comparison

jstucke:
Improved cve_lookup version comparison which was somewhat broken after the update to the 'packaging' module (which removed the `LegacyVersion` and forced versions to be PEP 440 compliant).

Also reduced log spam if the version cannot be parsed (which looked kind of confusing because the errors were logged with `logging.exception()` which prints traces even though the exception was actually caught).

### DIFF
--- a/src/plugins/analysis/cve_lookup/code/cve_lookup.py
+++ b/src/plugins/analysis/cve_lookup/code/cve_lookup.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import logging
 import operator
+import re
 import sys
 from collections import namedtuple
 from collections.abc import Callable
+from distutils.version import Version
 from itertools import combinations
 from pathlib import Path
-from re import match
 from typing import NamedTuple
 
 from packaging.version import InvalidVersion
@@ -31,6 +32,8 @@ MAX_TERM_SPREAD = (
     3  # a range in which the product term is allowed to come after the vendor term for it not to be a false positive
 )
 MAX_LEVENSHTEIN_DISTANCE = 0
+DOTTED_VERSION_REGEX = re.compile(r'^[a-zA-Z0-9\-]+(\\\.[a-zA-Z0-9\-]+)+$')
+VALID_VERSION_REGEX = re.compile(r'v?(\d+!)?\d+(\.\d+)*([.-]?(a(lpha)?|b(eta)?|c|dev|post|pre(view)?|r|rc)?\d+)?')
 
 
 class Product(NamedTuple):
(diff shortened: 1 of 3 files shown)


src/plugins/analysis/cve_lookup/code/cve_lookup.py
```diff
@@ -230,11 +232,32 @@ def versions_match(cpe_version: str, cve_entry: CveDbEntry) -> bool:
...
+    except InvalidVersion:
+        # try to convert other conventions (e.g. debian policy) to PEP 440
+        fixed_version = version.lower().replace('~', '-').replace(':', '!', 1).replace('_', '-')
+    try:
+        return parse_version(fixed_version)
+    except InvalidVersion:
+        match = VALID_VERSION_REGEX.match(fixed_version)
+        if match:
+            valid_version = match.group()
+            rest = re.sub('[^a-zA-Z0-9._-]', '', fixed_version[len(valid_version) :]).lstrip('._-')
```
rhelmke, line 250: `\w=[A-Za-z0-9_]` --> Maybe a little simpler here

```suggestion
            rest = re.sub('[^\w.-]', '', fixed_version[len(valid_version) :]).lstrip('._-')
```

src/plugins/analysis/cve_lookup/code/cve_lookup.py
```diff
@@ -230,11 +232,32 @@ def versions_match(cpe_version: str, cve_entry: CveDbEntry) -> bool:
...
+            return parse_version(f'{valid_version}+{rest}')
+        # try to throw away revisions and other stuff at the end as a final measure
+        return parse_version(re.split('[^v.0-9]', fixed_version)[0])
```
rhelmke, line 253: Use `\d`
```suggestion
        return parse_version(re.split('[^v.\d]', fixed_version)[0])
```

src/plugins/analysis/cve_lookup/code/cve_lookup.py
```diff
@@ -142,11 +144,11 @@ def generate_search_terms(product_name: str) -> list[str]:
 
 def find_matching_cpe_product(cpe_matches: list[Product], requested_version: str) -> Product:
```
rhelmke, line 145: Am I missing something or don't we convert the `requested_version`, too? If not, shouldn't we do so in order to establish a somewhat consistent comparability between requested version and the `next_closest_version`?

rhelmke (changes requested):
So I'm no

src/plugins/analysis/cve_lookup/code/cve_lookup.py
jstucke: I replaced `parse_version` with the new function here now, too

src/plugins/analysis/cve_lookup/code/cve_lookup.py
jstucke: fixed (with a raw string though see [re](https://docs.python.org/3/library/re.html))

src/plugins/analysis/cve_lookup/code/cve_lookup.py
jstucke: fixed (with a raw string though see [re](https://docs.python.org/3/library/re.html))

rhelmke (approved):

PR #971, opened 2023-01-26
imphash: replaced lief with readelf

jstucke:
lief is used for imp(ort) hash calculation but can crash with a segfault on broken ELF files which is ... unfortunate during an analysis of the mandatory "hash" plugin and shouldn't happen. This PR replaces lief with the more robust `readelf` which is included in the `binutils` package (which is already installed for the backend). The output should be identical in most cases (so that new hashes should match old ones).

Also removed the now unnecessary `_suppress_stdout` function.

### DIFF
--- a/src/helperFunctions/hash.py
+++ b/src/helperFunctions/hash.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
+
 import contextlib
 import logging
 import sys
 from hashlib import md5, new
+from typing import TYPE_CHECKING
 
 import lief
 import ssdeep
@@ -9,7 +12,15 @@ import tlsh
 
 from helperFunctions.data_conversion import make_bytes
 
-ELF_MIME_TYPES = ['application/x-executable', 'application/x-object', 'application/x-sharedlib']
+if TYPE_CHECKING:
+    from objects.file import FileObject
+
+ELF_MIME_TYPES = [
+    'application/x-executable',
+    'application/x-object',
+    'application/x-pie-executable',
+    'application/x-sharedlib',
+]
 
 
 def get_hash(hash_function, binary):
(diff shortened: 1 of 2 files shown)


dorpvom (changes requested):
I can tell without looking to closely, that for text-parsing like that I'll always want a low level unit test on example output 😉

maringuu:
I'm not a fan of parsing text output.
Could https://github.com/eliben/pyelftools probably be used to to what we need?

Also is there an issue that reports the segfault for lief?

maringuu:
Also lief should be removed from the requirements.txt

jstucke:
> Also is there an issue that reports the segfault for lief?

Do you mean in the lief GitHub repo? Not yet. There were multiple issues reporting segfaults and I'm not sure if this is actually a new one.

maringuu:
> Do you mean in the lief GitHub repo? Not yet. 

Then lets create one, they'll probably fix it and we can use lief again in the future.

Did you consider https://github.com/eliben/pyelftools?

dorpvom:
I observed that pyelftools seems to be ~ 500 % slower than readelf. This seems unnecessary for my taste.

romainthomas:
Let me know if there is a segfault in the LIEF's parsers, I'll be happy to fix it.

jstucke:
Hi @romainthomas and thank you for the offer. I can send you the file that caused the segfault tomorrow when I'm back in the office

src/helperFunctions/hash.py
```diff
@@ -51,27 +62,25 @@ def get_tlsh_comparison(first, second):
...
     :param file_object: The FileObject of which the imphash shall be computed
     '''
     if _is_elf_file(file_object):
         try:
             with _suppress_stdout():
-                # pylint: disable=c-extension-no-member
-                functions = normalize_lief_items(lief.parse(file_object.file_path).imported_functions)
-            return md5(','.join(sorted(functions)).encode()).hexdigest()
-        except Exception:
+                functions = [f.name for f in lief.ELF.parse(file_object.file_path).imported_functions]
```
maringuu, line 75: How exactly does this fix the error that we got before?

jstucke: The problematic binaries that caused the segfaults were [android OAT binaries](https://www.droidwiki.org/wiki/OAT) and lief has a separate parser for these files. When calling `lief.parse()` the parser is chosen automatically by lief. The segfault was caused by lief's OAT parser. Using `lief.ELF.parse()` forces the use of the ELF parser (which does not crash).

dorpvom (approved):

PR #981, opened 2023-02-17
Consistently store only category of information leak in summary

dorpvom:
For most things (IDE / SDK configs etc.) leaks are already grouped. File paths are stored independently though, making the summary cluttered. These are now also grouped by their categories (user paths, root paths, www paths).


jstucke (approved):
LGTM!

PR #983, opened 2023-02-21
added optional authentication for Redis

### LINKED ISSUE #982: Backend AuthentificationError
### FACT version

3364b6b19c084b2d9c185fd12b591f63c8b3da2e

### Environment

Ubuntu jammy 22.04.2 LTS

### Steps to reproduce

./start_fact_backend

same problem with ./start_all_installes_fact_components

### Expeced Behavior

Backend starts normally, so FACT can be used

### Installation logs

_No response_

### Backend logs

Unbenannt

### Frontend logs

_No response_

### Other information

I already tried to reinstall and update Backend --> ./src/installer.py -B

installation complete with no errors or warnings

jstucke:
resolves #982

### DIFF
--- a/src/storage/redis_interface.py
+++ b/src/storage/redis_interface.py
@@ -20,8 +20,9 @@ class RedisInterface:
         redis_db = cfg.data_storage.redis_fact_db
         redis_host = cfg.data_storage.redis_host
         redis_port = cfg.data_storage.redis_port
+        redis_pw = cfg.data_storage.redis_pw
 
-        self.redis = Redis(host=redis_host, port=redis_port, db=redis_db)
+        self.redis = Redis(host=redis_host, port=redis_port, db=redis_db, password=redis_pw)
 
     def set(self, key: str, value: Any):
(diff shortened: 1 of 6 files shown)


src/storage/redis_interface.py
```diff
@@ -20,8 +20,9 @@ def __init__(self, chunk_size=REDIS_MAX_VALUE_SIZE):
         redis_db = cfg.data_storage.redis_fact_db
         redis_host = cfg.data_storage.redis_host
         redis_port = cfg.data_storage.redis_port
+        redis_pw = cfg.data_storage.redis_pw or None  # empty str -> None
```
maringuu, line 23: This is not neccessary as `_parse_dict` from `src/config.py` does remove entrys that are only an empty string.
This in turn results to pydantic setting it to `None`.

maringuu (changes requested):

src/storage/redis_interface.py
jstucke: That is good to know. I will remove it then.

maringuu (approved):

PR #984, opened 2023-02-23
Added indicator to signify that unpacking is halted

dorpvom:
- Unpacking is throttle for larger analysis queues. That was not transparent before
- Fixed pre-commit path for flake8 as sideeffect

### DIFF
--- a/src/web_interface/static/js/system_health.js
+++ b/src/web_interface/static/js/system_health.js
@@ -32,8 +32,19 @@ async function updateSystemHealth() {
             const queueElement = document.getElementById("backend-unpacking-queue");
             if (entry.unpacking.unpacking_queue > 500) {
                 queueElement.classList.add("text-warning");
+            } else {
+                queueElement.classList.remove("text-warning");
             }
             queueElement.innerText = entry.unpacking.unpacking_queue.toString();
+
+            const throttleElement = document.getElementById("backend-unpacking-throttle-indicator");
+            if (entry.unpacking.is_throttled) {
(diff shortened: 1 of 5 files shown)


src/web_interface/static/js/system_health.js
```diff
@@ -32,8 +32,19 @@ async function updateSystemHealth() {
...
             if (entry.unpacking.unpacking_queue > 500) {
                 queueElement.classList.add("text-warning");
+            } else {
+                queueElement.classList.remove("text-warning");
             }
             queueElement.innerText = entry.unpacking.unpacking_queue.toString();
+
+            const throttleElement = document.getElementById("backend-unpacking-throttle-indicator");
+            if (entry.unpacking.is_throttled) {
+                throttleElement.innerText = '<i class="fa-regular fa-circle-pause"></i>';
```
jstucke, line 42: this should be `innerHTML` and "fa-regular fa-circle-pause" won't render for me (do we have an old version of font awesome?)
```suggestion
                throttleElement.innerHTML = '<i class="far fa-pause-circle"></i>';
```

src/web_interface/static/js/system_health.js
```diff
@@ -32,8 +32,19 @@ async function updateSystemHealth() {
...
+            }
+            else {
+                throttleElement.innerHTML = '';
+             }
```
jstucke, line 46:
```suggestion
            } else {
                throttleElement.innerHTML = '';
            }
```

jstucke (changes requested):

jstucke (approved):

PR #992, opened 2023-03-09
scheduler/unpacking: Don't call post_unpack when extraction failed

maringuu:
This lead to confusion when unpacking failed but seemingly succeeded because post_unpack was called.

### DIFF
--- a/src/scheduler/unpacking_scheduler.py
+++ b/src/scheduler/unpacking_scheduler.py
@@ -160,6 +160,7 @@ class UnpackingScheduler:  # pylint: disable=too-many-instance-attributes
                 docker_logs = self._fetch_logs(container)
                 logging.warning(f'Exception happened during extraction of {task.uid}.{docker_logs}')
                 container.set_exception()
+                return
 
             sleep(cfg.expert_settings.unpacking_delay)  # unpacking may be too fast for the FS to keep up
 


jstucke:
Do we really want this? I'm not so sure: Even if unpacking fails, it should still make sense to analyze the file (that's where "post_analysis" leads). And why wouldn't it? The analysis does not depend on whether or not the file could be unpacked successfully. What we actually don't want is scheduling extracted files (since there obviously aren't any) but this is already handled.

Maybe we could make it clearer in the "analysis result" of the "unpacker" that the extraction failed (i.e. add an entry in `fo.processed_analysis["unpacker"]` with key `failed` analog to `_get_failed_analysis_result()`).

maringuu:
> What we actually don't want is scheduling extracted files (since there obviously aren't any) but this is already handled.

Good point. 

This PR was motivated by debugging why [using only one unpack worker in the tests](https://github.com/fkie-cad/FACT_core/pull/976/files#diff-4134b50126b9d890bfba5d0cdd70815f4529606889b86a1c152255b1ba333389R105-R111) leads to errors.

> Maybe we could make it clearer in the "analysis result" of the "unpacker" that the extraction failed 

Yeah I like this idea. Otherwise the `post_unpack` function has no way of knowing if the unpacking succeeded.
Tests like `src/test/integration/scheduler/test_unpack_only.py` assume that post_unpack is only called when unpacking succeeded, which is wrong. If the callback would know it could actually count the number of extracted files.

PR #996, opened 2023-03-20
System health "stage 2"

jstucke:
- Adds per plugin details to "current analyses" on page "system health"
- Includes a checkbox to make showing details optional
- adds a status info for "out queues" of plugins


src/scheduler/analysis_status.py
```diff
@@ -52,13 +54,23 @@ def _update_current_analysis(self, fw_object: FileObject):
...
             updated_dict['files_to_unpack'] = list(set(updated_dict['files_to_unpack']).union(new_files))
             if fw_object.uid in updated_dict['files_to_unpack']:
                 updated_dict['files_to_unpack'].remove(fw_object.uid)
                 updated_dict['files_to_analyze'].append(fw_object.uid)
                 updated_dict['unpacked_files_count'] += 1
             self.currently_running[parent] = updated_dict
 
+    def update_post_analysis(self, fw_object: FileObject, plugin: str):
+        for parent in self._find_currently_analyzed_parents(fw_object):
+            self.currently_running_lock.acquire()
```
dorpvom, line 67: `with self._get_lock():`

dorpvom: This should then be acquired outside the loop. I don't see why it should be release in the interim.

dorpvom (changes requested):

src/scheduler/analysis_status.py
jstucke: changed

dorpvom (approved):

PR #997, opened 2023-03-21
Concurrent analysis scheduling

jstucke:
- added concurrency to the analysis scheduler so that scheduling better scales in systems with many cores/threads
    - adds new config variables for configuring count of the scheduler (`scheduling_worker_count`) and result collector processes (`collector-worker-count`)
- added a check that prevents the same file to be scheduled multiple times (this happen with multiple scheduler processes and can lead to errors)
- some refactoring


dorpvom (changes requested):

dorpvom (approved):
Nevermind

PR #1000, opened 2023-03-27
Delete fw speedup

jstucke:
Deletion of firmware is really slow if the firmware has thousands of included files (also used during "redo analysis"). This PR tries to speed up the process by optimizing the deletion operations in the database.
- improved SQLAlchemy deletion event listener
- improved delete_firmware method in DB interface
- schema update (used "on cascade delete" in all foreign keys) needed for bulk delete command compatibility
- added alembic script to automate DB migration

In a test with a firmware with ~1.5k files this reduced deletion time by roughly 80%


src/storage/migration/util.py
```diff
@@ -0,0 +1,17 @@
...
+
+from alembic import config, script
+from alembic.runtime import migration
+
+from storage.db_connection import AdminConnection
+
+alembic_cfg_path = Path(__file__).parent.parent.parent / 'alembic.ini'
+
+
+def db_needs_migration():
```
dorpvom, line 11: I assume this module is where we need to fix the path-problem regarding starting fact from project root or from `src/` directory.

src/storage/migration/versions/221cfef47173_added_oncascade_delte_for_foreign_keys.py
```diff
@@ -0,0 +1,85 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '221cfef47173'
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
dorpvom, line 18: Sure

dorpvom (changes requested):

src/storage/migration/util.py
jstucke: should work now

src/storage/migration/versions/221cfef47173_added_oncascade_delte_for_foreign_keys.py
jstucke: the script was automatically generated but I can remove those things if you like

dorpvom (approved):

PR #1002, opened 2023-03-28
add Mint 21.1 codename to install

jstucke:


maringuu:
CI fail is because of #1003 I guess?

maringuu (approved):

PR #1003, opened 2023-03-28
pinned YARA version to 4.2.3

jstucke:
because output of v4.3 has changed which causes a parsing error

ToDo: upgrade YARA to 4.3 and fix parsing errors

### DIFF
--- a/src/install/backend.py
+++ b/src/install/backend.py
@@ -7,8 +7,6 @@ from contextlib import suppress
 from pathlib import Path
 from subprocess import PIPE, STDOUT
 
-import requests
-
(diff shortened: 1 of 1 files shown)


maringuu:
> ToDo: upgrade YARA to 4.3 and fix parsing errors

I'd say we rather use `yara-python` than parsing `yara` output. Is there any benefit in parsing cli output?

maringuu (approved):
LGTM!

maringuu:
>  a lot slower than using `yara -r`

Hmm this seems wierd as `yara-python` is only a wrapper to the C library.
Before upgrading the parsing I'd like to investigate what causes the slowness.

PR #1005, opened 2023-03-30
helperFunctions/plugin: Update `sys.modules`

maringuu:
The deprecated
[load_module](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) did this.
Since we are using `exec_module` which apparently does not update `sys.modules` we have to do it ourselves.

The documentation is not that clear imo. This fix might not be the correct way to do this, but it works fine.


jstucke (approved):
LGTM!

PR #1006, opened 2023-03-31
bugfix for same vendor/class search on analysis page causing an error

jstucke:


maringuu:
Could you add the reason for the error and why the fix works in the commit message?

jstucke:
> Could you add the reason for the error and why the fix works in the commit message?

Do you think anyone will actually read the message at some point? :sweat_smile: 
Be that as it may, I extended the message (I hope it makes more sense now).

dorpvom (approved):

PR #1010, opened 2023-04-04
replaced advanced search input with text area and added validation

jstucke:
- replaced advanced search input with text area 
    - because the input was too small for complex queries
- added query input validation
    - the validation only checks for valid JSON at the moment
    - but it fixes trailing commas
    - validation before posting has the advantage that you don't lose the query that you were typing
- also removed old unused angular code


maringuu:
LGTM!

maringuu (approved):

PR #1019, opened 2023-04-26
Added unicode error handling in data sanitization

dorpvom:
- Fixes a bug where files with illegal names could not be added to database, as vfp still contains illegal symbols (file name is already sanitized)


jstucke (approved):
LGTM!

PR #1021, opened 2023-04-26
Refactoring and test coverage

jstucke:
- fixed the logging output of the pip package installation (switched to `stdout` because `stderr` is usually empty)
- added an advanced search test for searching in objects (dicts) inside arrays (lists) in `JSONB` columns
  - the syntax is not very intuitive and therefore the test also demonstrates how such a search works
    - which could be interesting for a structural change of the analysis results of some plugins in the future


maringuu:
LGTM!

maringuu (approved):

PR #1025, opened 2023-05-02
sped up general stats with optimized functions

jstucke:


dorpvom:
Zero Division Errors in Tests :wink: 
Otherwise I tried this and it does seem to speed up page load by a factor of 60 or so...

dorpvom (approved):

PR #1026, opened 2023-05-04
users & passwords: use better wordlist

jstucke:
The use of a better wordlist should significantly improve the password cracking abilities of the "users & passwords" plugin (while increasing the average runtime a bit)


maringuu (approved):
LGTM!

PR #1028, opened 2023-05-04
device_tree: Use `PluginV1`

maringuu:
Simelar to #1022 

The alembic commit is probably not needed and superseded by #1027

### DIFF
--- a/src/plugins/analysis/device_tree/internal/schema.py
+++ b/src/plugins/analysis/device_tree/internal/schema.py
@@ -103,9 +103,7 @@ class DeviceTree(pydantic.BaseModel):
             'https://devicetree-specification.readthedocs.io/en/latest/chapter2-devicetree-basics.html?highlight=model#model'
         ),
     )
-    description: Optional[str] = Field(
-        description='',
-    )
+    description: Optional[str] = Field()
 
     @classmethod
     def from_binary(cls, binary: bytes, offset: int = 0):


maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/5543784794

maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/5574650248

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -1,38 +1,70 @@
...
+        self,
+        file_handle: io.FileIO,
+        virtual_file_path: dict,
+        analyses: Dict[str, dict],
+    ) -> Optional[Schema]:
+        del virtual_file_path, analyses
+        device_trees = dump_device_trees(file_handle.readall())
+        if len(device_trees) == 0:
+            return None
+        return AnalysisPlugin.Schema(device_trees=device_trees)
```
jstucke, line 60: why `AnalysisPlugin.Schema` and not `self.Schema` as this is not a class method?

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -1,38 +1,70 @@
...
+
+    def get_tags(self, result, summary):
```
jstucke, line 62: IMHO we should try to type hint new code if possible (this should help later when we introduce mypy)
```suggestion
    def get_tags(self, result: Schema, summary: list[str]) -> list[Tag]:
```

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -1,38 +1,70 @@
...
+            version='1.1.0',
+            system_version=None,
+            mime_blacklist=[*MIME_BLACKLIST_COMPRESSED, 'audio', 'image', 'video'],
+            timeout=10,
+            Schema=AnalysisPlugin.Schema,
+        )
+        super().__init__(metadata=metadata)
 
-        return file_object
+    def summarize(self, result: Schema):
```
jstucke, line 37:
```suggestion
    def summarize(self, result: Schema) -> list[str]:
```

jstucke:
I tried running it and there seem to be problems with the "architecture_detection" plugin (which depends on the device_tree plugin) not expecting `None` as result:

```python
Traceback (most recent call last):
  File "/usr/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "src/helperFunctions/process.py", line 91, in run
    raise exception
  File "src/helperFunctions/process.py", line 86, in run
    Process.run(self)
  File "/usr/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "src/analysis/PluginBase.py", line 196, in process_next_object
    finished_task = self.analyze_file(task)
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "src/analysis/PluginBase.py", line 159, in analyze_file
    fo = self.process_object(file_object)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/plugins/analysis/architecture_detection/code/architecture_detection.py", line 37, in process_object
    arch_dict = construct_result(file_object, self._fs_organizer)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/plugins/analysis/architecture_detection/code/architecture_detection.py", line 50, in construct_result
    result.update(dt.construct_result(file_object))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/plugins/analysis/architecture_detection/internal/dt.py", line 76, in construct_result
    for dt_dict in file_object.processed_analysis['device_tree'].get('result', {}).get('device_trees', []):
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -1,38 +1,70 @@
...
+        device_trees: List[DeviceTree]
+
+    def __init__(self):
+        metadata = AnalysisPluginV0.MetaData(
+            name='device_tree',
+            description='get the device tree in text from the device tree blob',
+            version='1.1.0',
+            system_version=None,
+            mime_blacklist=[*MIME_BLACKLIST_COMPRESSED, 'audio', 'image', 'video'],
+            timeout=10,
```
jstucke, line 32: we didn't discuss this explicitly but as far as I can tell 10 seconds should be enough (I tried it with the FW containing multiple device_trees with large binary data chunks and the max runtime value in the stats was 0.09 sec)

src/plugins/analysis/device_tree/code/device_tree.py
```diff
@@ -1,38 +1,73 @@
...
-                color=TagColor.ORANGE,
-                propagate=False,
-            )
+class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
+    class Schema(pydantic.BaseModel):
+        class DeviceTree(pydantic.BaseModel):
+            header: Optional[dict]
+            device_tree: Optional[str]
+            model: Optional[str]
+            description: Optional[str]
```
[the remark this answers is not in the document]
maringuu: Changed.

src/plugins/analysis/device_tree/code/device_tree.py
maringuu: Now the schema lives in its own module schema.py.

maringuu:
> I tried running it and there seem to be problems with the "architecture_detection" plugin (which depends on the device_tree plugin) not expecting None as result:

Fixed.

src/plugins/analysis/device_tree/internal/schema.py
```diff
@@ -0,0 +1,172 @@
...
+                    byteorder='big',
+                ),
+                size_dt_strings=int.from_bytes(
+                    binary[32:36],
+                    byteorder='big',
+                ),
+                size_dt_struct=int.from_bytes(
+                    binary[36:40],
+                    byteorder='big',
+                ),
```
jstucke, line 84: adding a small helper function could improve readability and reduce code duplication here a lot since this only seems to depend on the offset (byteorder and lenght seem to be constant), i.e. 
```python
def _read_int_from_header(header: bytes, offset: int) -> int:
    return int.from_bytes(header[offset:offset + 4], byteorder='big')
```

src/plugins/analysis/device_tree/internal/schema.py
```diff
@@ -0,0 +1,172 @@
...
+        ),
+    )
+    description: Optional[str] = Field(
+        description='',
+    )
+
+    @classmethod
+    def from_binary(cls, binary: bytes, offset: int = 0):
+        """Given a binary and an offset into that binary constructs an instance of DeivceTree.
+        Raises IllegalDeviceTreeError for non sensical devicetrees.
```
jstucke, line 140:
```suggestion
        """Given a binary and an offset into that binary constructs an instance of DeviceTree.
        Raises IllegalDeviceTreeError for nonsensical device trees.
```

jstucke (commented):
Did you try to run it? I always got an Internal Server Error when trying it out:
```python
  File "src/web_interface/filter.py", line 439, in hide_dts_binary_data
    device_tree = re.sub(r'\[[0-9a-f ]{32,}]', '(BINARY DATA ...)', device_tree)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/re/__init__.py", line 185, in sub
    return _compile(pattern, flags).sub(repl, string, count)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'Undefined'
```

src/plugins/analysis/device_tree/internal/schema.py
```diff
@@ -0,0 +1,172 @@
...
+        version: int
+        last_comp_version: int
+        boot_cpuid_phys: int
+        size_dt_strings: int
+        size_dt_struct: int
+
+        @classmethod
+        def from_binary(cls, binary: bytes):
+            """Given the whole device tree binary parses the header and does some sanity checks."""
+            assert len(binary) >= cls.SIZE, f'Given header has size {len(binary)} but it should be at least {cls.SIZE}'
```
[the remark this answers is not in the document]
maringuu: It was an artifact from a previous schema design iteration. Changed it to IllegalHeaderException

src/plugins/analysis/device_tree/internal/schema.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+    model: Optional[str] = Field(
+        description=(
+            'The model as described in the spec.\n'
+            'https://devicetree-specification.readthedocs.io/en/latest/chapter2-devicetree-basics.html?highlight=model#model'
+        ),
+    )
+    description: Optional[str] = Field(
+        description='',
+    )
```
maringuu, line 135: Regarding the docs for this. Where is this defined? I could not find it in the spec.

maringuu:
Everything should work now.

jstucke:
I still get the error message when looking at old analysis results:
```python
  File "<template>", line 6, in template
  File "src/web_interface/filter.py", line 448, in hide_dts_binary_data
    device_tree = re.sub(r'\[[0-9a-f ]{32,}]', '(BINARY DATA ...)', device_tree)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/re/__init__.py", line 185, in sub
    return _compile(pattern, flags).sub(repl, string, count)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Normally this should be intercepted by the version check, but since the version changed from 1.0.1 to 1.1.0, it is treated as compatible. So a "fix" would be to set the version to 2.0.0 (which would be kinda appropriate since there are breaking changes).

Apart from that everything seems to work fine.

src/plugins/analysis/device_tree/internal/schema.py
maringuu: @jstucke Any ideas?

maringuu:
Bumped the version to 2.0.0. Sill the above question is unsolved.

src/plugins/analysis/device_tree/internal/schema.py
jstucke: I don't know if there is a spec for this. All I know is that this field frequently appears in device trees (and also it usually contains some interesting information)

maringuu: Interesting. I'll write that in the documentation.

Interestingly, this field does not show up in any of the device trees shipped with the Linux kernel (`cat $(fd -e dts) | rg description`).

jstucke: I looked through some device trees in my local DB and it seems the "description" field is present if OpenWRT is used (so maybe it's just an OpenWRT thing?)

jstucke (approved):

PR #1031, opened 2023-05-11
extractor: added retry to mitigate connection errors

jstucke:
`ConnectionError`s mostly appear in tests because the docker container is not yet ready for accepting extraction requests

### DIFF
--- a/src/unpacker/extraction_container.py
+++ b/src/unpacker/extraction_container.py
@@ -4,17 +4,21 @@ import logging
 import multiprocessing
 from contextlib import suppress
 from os import getgid, getuid
+from pathlib import Path
 from tempfile import TemporaryDirectory
 
 import docker
+import requests
 from docker.errors import APIError, DockerException
 from docker.models.containers import Container
 from docker.types import Mount
+from requests.adapters import HTTPAdapter, Retry
 
 from config import cfg
 
 DOCKER_CLIENT = docker.from_env()
 EXTRACTOR_DOCKER_IMAGE = 'fkiecad/fact_extractor'
+retries = Retry(total=5, backoff_factor=0.1)
 
 
 class ExtractionContainer:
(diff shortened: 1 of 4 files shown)


src/unpacker/extraction_container.py
```diff
@@ -4,17 +4,21 @@
...
 from docker.errors import APIError, DockerException
 from docker.models.containers import Container
 from docker.types import Mount
+from requests.adapters import HTTPAdapter, Retry
 
 from config import cfg
 
 DOCKER_CLIENT = docker.from_env()
 EXTRACTOR_DOCKER_IMAGE = 'fkiecad/fact_extractor'
+retries = Retry(total=5, backoff_factor=0.1)
```
maringuu, line 21: Why not put this directly in the argument? It is only used once.

src/unpacker/extraction_container.py
```diff
@@ -24,6 +28,8 @@ def __init__(self, id_: int, tmp_dir: TemporaryDirectory, value: multiprocessing
         self.port = cfg.unpack.base_port + id_
         self.container_id = None
         self.exception = value
+        self.session = requests.Session()
```
maringuu, line 31: This should be a private variable.

src/unpacker/extraction_container.py
```diff
@@ -24,6 +28,8 @@ def __init__(self, id_: int, tmp_dir: TemporaryDirectory, value: multiprocessing
...
+        self.session.mount('http://', HTTPAdapter(max_retries=retries))
```
maringuu, line 32: Shouldn't sessions be closed?

maringuu (changes requested):

src/unpacker/extraction_container.py
jstucke: in-lined it

src/unpacker/extraction_container.py
jstucke: switched to usage as context manager

src/unpacker/extraction_container.py
jstucke: switched to usage as context manager

maringuu (approved):
LGTM!

PR #1032, opened 2023-05-11
updated alpine base image

jstucke:
updated alpine base image of users_and_passwords and qemu_exec plugins from alpine 3.14 to 3.17 (because 3.14 has reached its end of life)


jstucke:
> Alpine 3.18 was released two days ago.

Good point, let's switch to that one then :sweat_smile:

maringuu (approved):

PR #1033, opened 2023-05-12
added ruff pre-commit hook and removed flake8 and pylint

jstucke:
pylint has become so slow on modules like sqlalchemy that it is almost unusable. "ruff" is a linter written in Rust that can replace many checks of popular python linters and is lightning fast. The downside: it does not implement *all* checks that pylint does. On the upside it includes more checks from other linters we did not previously use like bugbear. Other not included  pylint checks 
are covered by included flake8 extensions. On the whole, it extends the linting capabilities and adds some new interesting checks.

- added ruff pre-commit hook (incl. auto-fix)
- removed pylint, isort and flake8 hooks (mostly replaced by ruff)
- updated version of black hook (to address warning)
- added toml check hook
- updated python version (to reflect removal of python 3.7 support #1024 )
- added ruff configuration to pyproject.toml
- the pylint configuration is still present in case someone want to manually run pylint

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,8 +47,8 @@ markers = [
 
 [tool.ruff]
 select = ["F", "E", "W", "C90", "N", "UP", "B", "A", "PL", "C4", "EXE", "ISC", "PIE", "T20", "PT", "RET", "SIM", "TCH",
-    "PTH", "PLR", "PLW", "ARG", "ERA"]
-ignore = ["A003"]
+    "PTH", "PLR", "PLW", "ARG", "ERA", "RUF"]
+ignore = ["A003", "RUF001", "RUF002", "RUF003"]
 # Allow autofix for all enabled rules (when `--fix`) is provided.
 fixable = ["F", "W", "I", "C4", "PT", "RET"]
 exclude = [
@@ -57,6 +57,7 @@ exclude = [
     ".venv",
     "bin",
     "node_modules",
+    "scripts",  # ingore plugins/*/*/docker/scripts/* jython ghidra scripts
     "venv",
 ]
 line-length = 120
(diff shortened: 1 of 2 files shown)


maringuu:
We have a lot of `# pytlint:disable=...` comments in our code. Is there an equivalent to in ruff? Does ruff respect the existing comments?

Also ruff can replace isort too, so I thing we should to that.

jstucke:
> Is there an equivalent to in ruff?

Yes: `# noqa: UP006` https://beta.ruff.rs/docs/tutorial/#ignoring-errors

> Also ruff can replace isort too, so I thing we should to that.

ruff also has a `--fix` option. I will look into how it is used together with pre-commit and how to configure the isort features in ruff.

maringuu:
We can also replace our flake8 action by the ruff action. I'm not sure if we want that but I guess so.
https://beta.ruff.rs/docs/usage/#github-action

jstucke:
> We can also replace our flake8 action by the ruff action. I'm not sure if we want that but I guess so. https://beta.ruff.rs/docs/usage/#github-action

The flake8 action should already be replaced (I replaced pylint, isort and flake8)

maringuu:
> The flake8 action should already be replaced

I meant action as in github action. See https://github.com/fkie-cad/FACT_core/blob/ruff-linter/.github/workflows/lint_action.yml

pyproject.toml
```diff
@@ -52,3 +44,33 @@ markers = [
     "cfg_defaults: Overwrite defaults for the testing config",
     "WebInterfaceUnitTestConfig: Configure the web_interface fixture",
 ]
+
+[tool.ruff]
+select = ["F", "E", "W", "C90", "N", "UP", "B", "A", "PL", "C4", "EXE", "ISC", "PIE", "T20", "PT", "RET", "SIM", "TCH",
+    "PTH", "PLR", "PLW", "ARG", "ERA", "RUF"]
+ignore = ["A003", "RUF001", "RUF002", "RUF003"]
+# Allow autofix for all enabled rules (when `--fix`) is provided.
```
maringuu, line 52: Not true, copy paste artifact?

maringuu (changes requested):

maringuu (approved):
LGTM!

PR #1034, opened 2023-05-12
sped up analysis update by using a more efficient DB method

jstucke:
The method `update_analysis_of_object_and_children()` used to fetch every object during an update individually from the database. Switching to `get_objects_by_uid_list()` improves the performance significantly. During a test with a firmware with ~1.7k files, this reduced the time to schedule the objects from ~58 seconds to ~2.2.


maringuu:
I wonder how many trivial performance improvements are hidden in the FACT codebase.
Good find!

maringuu (approved):

PR #1037, opened 2023-05-15
Added pie mime type to cwe_checker whitelist

dorpvom:
- application/x-pie-executable previously omitted due to negligence


jstucke (approved):
We should probably add it to `plugins/mime_blacklists.py` (and rename it to "mime_lists.py" or something, since it is a whitelist and not a blacklist)

PR #1039, opened 2023-05-16
unpacking fs_organizer bugfix

jstucke:
Somehow, we seem to never reach this code anyway (that's probably why we never noticed this bug)


maringuu:
Would you mind to explain why this code could theoretically be reached?
Maybe we can remove it?

maringuu (approved):

PR #1059, opened 2023-06-15
crypto hints: Use plugin.PluginV0

maringuu:
Depends on #987

### DIFF
--- a/src/plugins/analysis/crypto_hints/code/crypto_hints.py
+++ b/src/plugins/analysis/crypto_hints/code/crypto_hints.py
@@ -1,9 +1,32 @@
-from analysis.YaraPluginBase import YaraBasePlugin
+import io
+
+import pydantic
+import typing
+
+from analysis.plugin import addons, compat
+from analysis.plugin import AnalysisPluginV0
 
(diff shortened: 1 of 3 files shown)


maringuu:
Converting to draft as this will definitely break the schema.

maringuu:
Despite the missing database migration this is ready!

maringuu:
I'll leave out the database migration.
The frontend will show a message that the analysis is outdated and we can just rerun it.
@jstucke Agree? This is what we discussed iirc.

maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/5574661620

maringuu:
https://github.com/fkie-cad/FACT_core/actions/runs/5575210844

jstucke (commented):
In some cases the matching strings appear as strings instead of hex strings (which was not the case in the old version). This can lead to some rather odd results in the web interface:
image
I'm guessing this is a problem of the YARA addon or compat class and not the plugin.

Apart from that, everything looks fine

jstucke (approved):

PR #1065, opened 2023-06-28
Static file path resolution test

jstucke:
* [added test that verfies static urls from templates are found](https://github.com/fkie-cad/FACT_core/commit/aff1a0895f2b8d99e1e3dae01a990a0b8378f847)
* [fixed highlight.js link in text file comparison template](https://github.com/fkie-cad/FACT_core/commit/d6f62dd22a46510c3690c09e59fcc92f0baaafa4)

### DIFF
--- a/src/test/unit/web_interface/test_app_static.py
+++ b/src/test/unit/web_interface/test_app_static.py
@@ -1,23 +1,28 @@
+from __future__ import annotations
+
 import re
 from pathlib import Path
+from typing import Iterable
 
 import pytest
 from flask import url_for
 
+from helperFunctions.fileSystem import get_template_dir, get_src_dir
+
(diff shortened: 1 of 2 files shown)


dorpvom:
Please explain this test a bit more. Does this test if, after FACT installation, all local copies of js / css files exist and are accessible through the paths given in the web interface templates?

jstucke:
> Please explain this test a bit more. Does this test if, after FACT installation, all local copies of js / css files exist and are accessible through the paths given in the web interface templates?

yes, exactly: it looks up all `url_for("static", ...)` constructs in the templates (which should be used for static files) and checks that they are available if you request them with flask's test client

dorpvom (approved):

PR #1045, opened 2023-05-18
Make all linters happy.

maringuu:
Now that we switched to ruff we can easily make all linters happy on all files.

While this adds a lot (about 700 of noqa comments) this is arguably better than having the linter fail.
With the current state I developed the bad habit to just do `git commit --no-verify` because:
* pylint is slow (fixed by using ruff)
* The linter would fail anyways (because we had so many linting errors in the codebase)

Now after this PR this bad habit is not needed anymore and the linters will only produce true positives for changes I made.

Closes #1038 #894

PR #1074, opened 2023-07-05
added log messages for RAM usage

jstucke:
- added log messages for RAM usage
- refactored `fact_base` to avoid code duplication

### DIFF
--- a/src/fact_base.py
+++ b/src/fact_base.py
@@ -56,27 +56,34 @@ class FactBase:
             signal.signal(signal.SIGTERM, self.shutdown_listener)
 
     def shutdown_listener(self, signum, _):
-        if not self._is_main_process():
+        if not _is_main_process():
             return  # all subprocesses also inherit this signal handler (which is intentional for a "clean" shutdown)
         logging.info(f'Received signal {signum}. Shutting down {self.PROGRAM_NAME}...')
         self.run = False
+
+    def start(self):
+        pass
+
+    def _update_component_workload(self):
+        self.work_load_stat.update()
 
     def shutdown(self):
         logging.info(f'Shutting down components of {self.PROGRAM_NAME}')
         self.work_load_stat.shutdown()
 
     def main(self):
+        self.start()
         logging.info(f'Successfully started {self.PROGRAM_NAME}')
+        counter = 0
         while self.run:
-            self.work_load_stat.update()
+            self._update_component_workload()
             sleep(5)
             if self.args.testing:
                 break
+            if not counter % 12:  # only check every minute
+                _check_memory_usage()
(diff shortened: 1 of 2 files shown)


src/fact_base.py
```diff
@@ -86,3 +96,13 @@ def do_self_test():
                 'Please run `alembic upgrade head` from `src` to update the schema.'
             )
             raise DbInterfaceError('Schema mismatch')
+
+    @staticmethod
```
maringuu, line 100: I personally don't think private static methods make much sense.
Why not make it a function in in the module?

src/start_fact_backend.py
```diff
@@ -1,5 +1,5 @@
 #! /usr/bin/env python3
-'''
+"""
```
maringuu, line 2: I'd be really happy if we could refrain from doing unrelated ruff fixes in commits.
All those fixes will eventually (when #987 is merged) land in a single commit in #1045

maringuu (changes requested):

maringuu (changes requested):
See above.

src/fact_base.py
```diff
@@ -61,17 +61,27 @@ def shutdown_listener(self, signum, _):
...
+        self.start()
         logging.info(f'Successfully started {self.PROGRAM_NAME}')
+        counter = 0
         while self.run:
-            self.work_load_stat.update()
+            self._update_component_workload()
             sleep(5)
             if self.args.testing:
                 break
+            if not (counter := counter + 1) % 12:  # only check every minute
```
maringuu, line 83: I personally don't like the walrus operator. While I see its usecase in large list comprehensions to improve performance it doesn't improve performance here.
I find this very readable.

maringuu (changes requested):

src/fact_base.py
jstucke: converted the class method to a function

src/start_fact_backend.py
jstucke: merged master after the merge of #1045

src/fact_base.py
jstucke: > I find this very readable.

What is the problem then? :sweat_smile: 
Fun aside, I agree that it's harder to read than without the walrus operator -> changed

jstucke:
> See above.

What line should be referenced here? `_is_main_process()`?
Changed that as well.

maringuu:
> What line should be referenced here? _is_main_process()?

This was just GitHub messing around and not letting me submit a review that requests changes without also leaving a comment.

maringuu (approved):

PR #1079, opened 2023-07-12
reworked current analysis progress

jstucke:
* moved "current analyses" progress management into a separate process to reduce synchronization overhead
  * this fixes a problem with FWs with very many included files (>100k) where this overhead actually obstructs analysis and unpacking scheduling
* fixed tests
* stored progress in redis intstead of DB
  * added new redis status interface for that
* fixed /rest/status


src/scheduler/analysis.py
```diff
@@ -157,26 +161,23 @@ def update_analysis_of_object_and_children(self, fo: FileObject):
...
-        if self.status.file_should_be_analyzed(fo):
-            self.status.add_to_current_analyses(fo)
-            self.task_scheduler.schedule_analysis_tasks(fo, fo.scheduled_analysis, mandatory=True)
-            self._check_further_process_or_complete(fo)
-        else:
-            logging.info(f'Skipping analysis of {fo.uid} (duplicate file)')
+        """
+        self.status.add_object(fo)
+        self.task_scheduler.schedule_analysis_tasks(fo, fo.scheduled_analysis, mandatory=True)
+        self._check_further_process_or_complete(fo)
```
rhelmke, line 172: We no longer check whether files should be analyzed. I understand that the analysis status now uses sets, which appear to make `file_should_be_analyzed` obsolete. But what about `schedule_analysis_tasks`? It has not been modified. Are there any possible side effects of that missing duplicate check condition?

rhelmke (changes requested):
The `Currently Analyzed Firmware` Stat on the Webinterface's landing/home page does not work anymore.

src/test/unit/conftest.py
```diff
@@ -105,6 +116,8 @@ class WebInterfaceUnitTestConfig(BaseModel):
     database_mock_class: Type = CommonDatabaseMock
     #: A class mocking :py:class:`~intercom.front_end_binding.InterComFrontEndBinding`
     intercom_mock_class: Type[CommonIntercomMock] = CommonIntercomMock
+    #: A class mocking :py:class:`~intercom.front_end_binding.InterComFrontEndBinding`
+    redis_mock_class: Type[StatusInterfaceMock] = StatusInterfaceMock
```
maringuu, line 120: I don't like the name `redis_mock_class`. It isn't really mocking redis.
What about  `status_mock_class`?

Also the doc comment is a copy paste artifact and refers to the wrong class.

maringuu (changes requested):

src/scheduler/analysis.py
```diff
@@ -227,7 +228,7 @@ def get_plugin_dict(self) -> dict:
                 current_plugin_plugin_sets[plugin_set] = plugin in plugin_sets[plugin_set].plugins
             blacklist, whitelist = self._get_blacklist_and_whitelist_from_plugin(plugin)
             try:
-                thread_count = getattr(config.backend.plugin[plugin], 'processes')
+                thread_count = config.backend.plugin[plugin].processes
```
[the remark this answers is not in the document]
jstucke: This is an auto change by ruff. What is the difference between `getattr(config.backend.plugin[plugin], 'processes')` and `config.backend.plugin[plugin].processes`? There is none (since we don't define a fallback value).

src/scheduler/analysis.py
```diff
@@ -474,13 +475,11 @@ def get_scheduled_workload(self) -> dict:
             }
 
         :return: Dictionary containing current workload statistics
-        '''
-        self.status.clear_recently_finished()
+        """
+        # FixMe: move system status from DB to Redis
```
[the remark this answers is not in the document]
jstucke: Only in part. The CPU / RAM usage, etc. still uses the DB.  More correctly the comment should read "Move the rest of the system status from DB to Redis.

src/scheduler/analysis_status.py
```diff
@@ -1,160 +1,228 @@
...
+            files_to_unpack=set(included_files),
+            files_to_analyze={uid},
+            total_files_count=1 + len(included_files),
+            hid=hid,
+            analysis_plugins={p: 0 for p in scheduled_analyses or []},
+        )
+
+    def _add_included_file(self, uid: str, root_uid: str, included_files: set[str]):
+        """
+        new file comes from unpacking:
```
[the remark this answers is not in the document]
jstucke: Reworded the comment. Please check if if it's easier to understand now.

src/test/unit/conftest.py
jstucke: fixed
for whatever reason I also accidentally named the class Rest... instead of Redis... -> fixed that also

maringuu (approved):

src/scheduler/analysis.py
jstucke: `file_should_be_analyzed` was already obsolete and should have been removed with #1050

rhelmke (approved):

PR #1082, opened 2023-07-14
Yara rule meta parsing fix

jstucke:
- fixed bug in signature metadata parsing where commas in strings would cause errors
  - used pyyaml for parsing the metadata (this is a bit hacky but seems to work fine)
- added a test that should make sure that the metadata in all plugin signatures is compatible


maringuu:
lgtm!

maringuu (approved):

PR #1094, opened 2023-08-02
fixed a yara meta parsing bug

jstucke:
- fixed a YARA meta parsing bug caused by single quotes in metadata strings
  - YARA apparently inserts backslashes before single quotes in metadata strings (even though the strings themselves use double quotes)
  - the YAML parser does not like that ("unknown escape")
- changed the parsing test so that it replicates the problem


maringuu (approved):
lgtm

PR #1095, opened 2023-08-02
CVE lookup result sort fix

jstucke:
* fixed sorting of cve_lookup results
  * the results were erroneously sorted alphabetically instead of numerically (resulting in score 1<10<2)
  * changed sorting key so that the results are also sorted by CVSS v3 score (instead of v2 score only)
* added test cases
* ruff fixes


maringuu (approved):
lgtm!

PR #1097, opened 2023-08-07
added /rest/analysis endpoint

jstucke:
added `/rest/analysis/<uid>/<plugin>` REST endpoint for requesting the result of a specific analysis plugin of a specific file


maringuu (approved):
LGTM!

PR #1107, opened 2023-08-16
fixed bug that may occur when a dependency result is missing

### LINKED ISSUE #1106: error “TypeError: 'NoneType' object is not subscriptable”
### The FACT version you are using

_No response_

### Your question

image

jstucke:
resolves #1106


maringuu (approved):
While I'd like to note that this check should not be needed this lgtm.

PR #1108, opened 2023-08-17
update preset `KeyError` fix

jstucke:
fixed a `KeyError` that occurs during update preset generation when a plugin is removed (but the DB entry of the FW contains an analysis result of that plugin)


maringuu (approved):
lgtm

PR #1109, opened 2023-08-18
fs_metadata plugin: restructure code + refactoring

jstucke:
* restructured plugin "file_system_metadata"
  * instead of a class instance variable `result` let the functions return stuff
* some more refactoring and test fixes


maringuu (approved):
LGTM!

PR #1112, opened 2023-08-22
Linter plugin bugfix

jstucke:
- fixed bugs in the linter plugin:
  - fix for new schema being incompatible with the output of js and php linters
  - fix for missing file name causing the script language to not be detected correctly
- added test cases for more script types
- fixed a bug in the plugin runner that occurred when trying to receive an exception from the analysis process


src/plugins/analysis/linter/test/data/hello_world.py
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
-import os
+import os  # noqa: F401
```
maringuu, line 2: The noqa statements make no sense in a test file for linting.

src/scheduler/analysis/plugin.py
```diff
@@ -206,10 +206,12 @@ def _handle_sigterm(signum, frame):
             except queue.Empty:
                 continue
 
+            analysis_id = f'{self._plugin.metadata.name} analysis on {task.scheduler_state.uid}'
```
maringuu, line 209: nit: This is not really an id. More like analysis description or so.

src/plugins/analysis/linter/code/source_code_analysis.py
```diff
@@ -119,8 +129,7 @@ def analyze(self, file_handle: io.FileIO, virtual_file_path: str, analyses: dict
         )
 
 
-def run_linguist(file_path: str) -> dict:
-    file_name = Path(file_path).name
+def run_linguist(file_path: str, file_name: str) -> dict:
```
maringuu, line 132: Can we rename the second argument to virtual_name or virtual_file_name or sth?
Right now from looking at the signature I'd think both arguments are the same thing.

maringuu (changes requested):

maringuu:
Aside from the nits this lgtm.

src/plugins/analysis/linter/test/data/hello_world.py
jstucke: I added the file (or rather all test/data folders) to ruff's ignore list instead

src/scheduler/analysis/plugin.py
jstucke: changed

src/plugins/analysis/linter/code/source_code_analysis.py
jstucke: changed

maringuu:
LGTM!

maringuu (approved):

PR #1119, opened 2023-08-28
Fix deprecation warnings

jstucke:
[The last scheduled master CI build](https://github.com/fkie-cad/FACT_core/actions/runs/5995159722/job/16257774384) had 43624 warnings when running the tests. This PR should reduce this number to less than 30. There are some warnings that can't be fixed because they are caused by dependencies or where it is too complicated (the crypto_hints plugin should be fixed in its own PR).

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -8,8 +8,7 @@ from tempfile import TemporaryDirectory
 from typing import Type, Union
 
 import pytest
-from pydantic import BaseModel, Field
-from pydantic.v1 import ConfigDict
+from pydantic import BaseModel, ConfigDict, Field
 from pydantic.v1.utils import deep_update
 
 import config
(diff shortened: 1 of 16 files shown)


src/conftest.py
```diff
@@ -9,7 +9,8 @@
 
 import pytest
 from pydantic import BaseModel, Field
-from pydantic.utils import deep_update
+from pydantic.v1 import ConfigDict
```
maringuu, line 12: Why do we need this, what is the v2 way of doing things?

maringuu (changes requested):

src/conftest.py
jstucke: Hmm weird. According to the [docs](https://docs.pydantic.dev/latest/usage/model_config/) you can directly import it from `pydantic`:

```python
from pydantic import BaseModel, ConfigDict

class Model(BaseModel):
    model_config = ConfigDict(str_max_length=10)
```

Maybe I used the auto import fix in pycharm.

maringuu (approved):
lgtm!

PR #1123, opened 2023-09-01
Mypy hook

jstucke:
* added mypy pre-commit hook 
* added mypy config
* fixed all issues in the code base


src/analysis/plugin/plugin.py
```diff
@@ -106,34 +110,34 @@ def summarize(self, result: Schema) -> list[str]:
...
+    ) -> Schema | None:
         """Analyze a file.
         May return None if nothing was found.
 
         :param file_handle: :py:class:`io.FileIO` instance of the file to be analyzed
         :param virtual_file_path: The virtual file paths, see :py:class:`~objects.file.FileObject`
         :param analyses: A dictionary of dependent analysis
 
-        :return: The analysis if anything was found.
+        :return: The analysis results (if there are any).
```
maringuu, line 123: nit: why use parentheses?

src/analysis/plugin/plugin.py
```diff
@@ -106,34 +110,34 @@ def summarize(self, result: Schema) -> list[str]:
...
         if result is not None:
-            summary = self.summarize(result)
+            summary = self.summarize(result)  # type: ignore[unreachable]  # this is obviously reachable
             tags = self.get_tags(result, summary)
 
         # The dictionary as defined in the docs for FileObject.analyses_tags
         # Misses the root uid, which must be added by the scheduler
         tags_dict = {}
-        for tag in tags:
+        for tag in tags or []:
```
maringuu, line 140: Above we have `tags = []` so this is not needed, is it?

src/analysis/plugin/plugin.py
```diff
@@ -145,7 +149,7 @@ def get_analysis(self, file_handle: io.FileIO, virtual_file_path: dict, analyses
             'system_version': self.metadata.system_version,
             'summary': summary,
             'tags': tags_dict,
-            'result': result.dict() if result else None,
+            'result': result.dict() if isinstance(result, BaseModel) else None,
```
maringuu, line 152: Why is this needed?

src/helperFunctions/plugin.py
```diff
@@ -33,8 +37,11 @@ def _import_plugins(plugin_type):
         module_name = str(plugin_file).replace('/', '.')[len(src_dir + '/') : -len('.py')]
 
         loader = SourceFileLoader(module_name, str(plugin_file))
-        spec = importlib.util.spec_from_loader(loader.name, loader)
-        plugin_module = importlib.util.module_from_spec(spec)
+        spec = spec_from_loader(loader.name, loader)
+        if spec is None:
```
maringuu, line 41: Why and when can this happen?

src/helperFunctions/types.py
```diff
@@ -0,0 +1,60 @@
...
+from tempfile import _TemporaryFileWrapper
+from typing import Dict, List, Protocol, TypeVar, NamedTuple, TYPE_CHECKING, Union
+from unittest.mock import _patch
+
+
+if TYPE_CHECKING:
+    from analysis.plugin import AnalysisPluginV0
+
+KT = TypeVar('KT', str, tuple, bytes)  # generic key type
+VT = TypeVar('VT')  # generic value type
```
maringuu, line 14: Isn't it better to be explicit about the types?

src/test/unit/compare/compare_plugin_test_class.py
```diff
@@ -40,3 +49,4 @@ def setup_test_fw(self):
         self.fw_three = create_test_firmware(
             device_name='dev_3', bin_path='container/test.cab', all_files_included_set=True
         )
+        pass
```
maringuu, line 52: What does this do here?

src/test/unit/helperFunctions/test_task_conversion.py
```diff
@@ -53,7 +53,7 @@ def test_get_uid_of_analysis_task(self):
 
     def test_convert_analysis_task_to_firmware_object(self):
         fw_obj = Firmware()
-        fw_obj.tags = {'tag', tag.TagColor.GRAY}
+        fw_obj.tags = {'tag': tag.TagColor.GRAY}
```
maringuu, line 56: Urrg interesting that this worked.

maringuu (changes requested):
Overall lgtm but still some questions.

src/analysis/plugin/plugin.py
jstucke: In general? :sweat_smile:  

I don't know, if you want I can remove them. I originally just wanted to specify that analysis *results* are returned.

src/analysis/plugin/plugin.py
jstucke: Yes, this was a fix during the time when it was `tags = None` -> changed back

src/analysis/plugin/plugin.py
jstucke: I think during the time I wasn't sure if the plugin always returns a `BaseModel` or if it can be a `dict` if no schema is specified. I think as it is now, the schema is mandatory and the plugin always returns a `BaseModel`, right?

src/helperFunctions/plugin.py
jstucke: I'm not actually sure. Mypy claimed that `spec_from_loader()` may return `None` and I took a look in the code but wasn't really sure so I kept it.

src/helperFunctions/types.py
jstucke: This is for the generic case (i.e. the function accepts all dicts but you want to annotate e.g. that it will return the same type as the values of the dict) https://docs.python.org/3/library/typing.html#generics

src/test/unit/helperFunctions/test_task_conversion.py
jstucke: interesting indeed

src/analysis/plugin/plugin.py
maringuu: I meant that what is written in the parentheses is useful information and reads better without the parantheses.

src/analysis/plugin/plugin.py
maringuu: > right?

Yes!

src/helperFunctions/plugin.py
maringuu: I wouldn't trust mypy here.

jstucke: But do you know for sure that it isn't the case? Better safe than sorry? :sweat_smile:

maringuu: Checking for error conditions where there are none documented and we don't have any test case to produce any error does not make sense imo.
So I'd better be sorry.

src/scheduler/analysis/scheduler.py
```diff
@@ -387,10 +390,10 @@ def _set_binary(self, file_object: FileObject):
...
     @staticmethod
     def _is_forced_update(file_object: FileObject) -> bool:
-        try:
-            return bool(getattr(file_object, 'force_update', False))
-        except AttributeError:
-            return False
+        if hasattr(file_object, 'force_update'):
+            # FixMe: for backwards compatibility with older frontend. Remove in next release.
+            return bool(file_object.force_update)
+        return bool(file_object.temporary_data.get('force_update', False))
```
maringuu, line 396: Can you explain this to me?

src/statistic/analysis_stats.py
```diff
@@ -5,14 +5,17 @@
 import numpy as np
 from typing import TYPE_CHECKING
 
+
 if TYPE_CHECKING:
-    from multiprocessing import Value
+    from helperFunctions.types import MpValue, MpArray
+
+    pass
```
maringuu, line 12: Typo?

maringuu (commented):
I had a brief look at everything but didn't look in depth.
I hope I found any actual code changes and reviewed them.
For the typing changes I'm happy when mypy is.

src/helperFunctions/plugin.py
jstucke: I removed the check and added a `# type: ignore` instead

dorpvom:
We decided to split this up into smaller PRs to make each one reviewable

PR #1135, opened 2023-10-06
Clickable analysis tags

jstucke:
- Analysis tags can be clicked and link to the summary of the analysis (which set the tag)
- The summary is loaded automatically (only if a tag was clicked, not always)
- fixed a bug in the summary for file objects


maringuu (approved):
lgtm

PR #1138, opened 2023-10-09
tracebacks from analysis exceptions are now logged

jstucke:
- tracebacks from analysis exceptions were printed to the command line but not logged, resulting in them not showing up in log files or the web GUI
  - this only affected the old plugin base class but not the new base class / runner
- added a parameter `reraise` to `ExceptionSafeProcess` so that the trace is only printed once

### DIFF
--- a/src/analysis/PluginBase.py
+++ b/src/analysis/PluginBase.py
@@ -1,4 +1,6 @@
-import ctypes  # noqa: N999
+from __future__ import annotations  # noqa: N999
+
+import ctypes
 import logging
 import os
 from multiprocessing import Array, Manager, Queue, Value
@@ -17,8 +19,11 @@ from helperFunctions.process import (
     terminate_process_and_children,
 )
 from helperFunctions.tag import TagColor
-from objects.file import FileObject
 from plugins.base import BasePlugin
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from objects.file import FileObject
 
 META_KEYS = {
     'tags',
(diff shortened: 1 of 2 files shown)


src/analysis/PluginBase.py
```diff
@@ -233,10 +239,15 @@ def _update_duration_stats(self, duration):
         if self.analysis_stats_count.value < self.ANALYSIS_STATS_LIMIT:
             self.analysis_stats_count.value += 1
 
-    def _handle_failed_analysis(self, fw_object, process, worker_id, cause: str):
+    def _handle_failed_analysis(  # noqa: PLR0913
+        self, fw_object, process, worker_id, cause: str, trace: str | None = None
```
maringuu, line 243: nit: I really hate this formatting. As far as I know there is no option to force black to add the trailing comma..

src/analysis/PluginBase.py
```diff
@@ -214,7 +219,8 @@ def worker_processing_with_timeout(self, worker_id, next_task: FileObject):
         if self.timeout_happened(process):
             result_fo = self._handle_failed_analysis(next_task, process, worker_id, 'Timeout')
         elif process.exception:
-            result_fo = self._handle_failed_analysis(next_task, process, worker_id, 'Exception')
+            _, trace = process.exception
+            result_fo = self._handle_failed_analysis(next_task, process, worker_id, 'Exception', trace=trace)
```
maringuu, line 223: Why ignore the exception and use "Exception" in the line after?

maringuu (changes requested):

maringuu (approved):
lgtm

PR #1142, opened 2023-10-12
Add typehints and make mypy happy

maringuu:
This is the same as #1123 but rebased against master and better commit messages to make it reviewable


dorpvom:
We decided to split this (and #1123) up into smaller PRs.

PR #1143, opened 2023-10-16
Uefi plugin

jstucke:
- new analysis plugin based on FwHunt
  - [x] convert to V0
  - [x] write tests

### DIFF
--- a/src/plugins/analysis/uefi/code/uefi.py
+++ b/src/plugins/analysis/uefi/code/uefi.py
@@ -81,7 +81,7 @@ class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
             run_docker_container(
                 DOCKER_IMAGE,
                 combine_stderr_stdout=True,
-                timeout=self.TIMEOUT,
+                timeout=self.metadata.timeout,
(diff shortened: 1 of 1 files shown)


src/plugins/analysis/uefi/code/uefi.py
```diff
@@ -0,0 +1,161 @@
...
+        return self._analyze_uefi_module(file_handle.name, _get_analysis_mode(type_analysis.mime))
+
+    def _analyze_uefi_module(self, path: str, mode: str) -> Schema | None:
+        with TemporaryDirectory() as tmp_dir:
+            output_file = Path(tmp_dir) / 'output.json'
+            output_file.touch()
+            run_docker_container(
+                DOCKER_IMAGE,
+                combine_stderr_stdout=True,
+                timeout=self.TIMEOUT,
```
maringuu, line 84: This is part of the Mixin and should not be used.
Also I'd rather use functions instead of methods since plugins don't maintain state.

jstucke: Oops I didn't intend to use it. This is more of an artifact from the pre-V1 plugin

PR #1146, opened 2023-10-18
sort tags on analysis page

jstucke:
- sort tags on analysis page
  - to fix the problem that the order of displayed tags is random and switches around when you reload the page


maringuu (approved):
lgtm

Issue #1149, opened 2023-10-20
process stop 20hours


chenjianquan7:
### The FACT version you are using

4.1

### Your question

![image](https://github.com/fkie-cad/FACT_core/assets/26828709/0f08d191-6601-4018-a6c7-1ea903f0807c)
 all process stop, why?
it happend sometimes

chenjianquan7:
![image](https://github.com/fkie-cad/FACT_core/assets/26828709/eaae2461-338f-4011-8833-14b7d8db477c)

chenjianquan7:
![image](https://github.com/fkie-cad/FACT_core/assets/26828709/3edb96bb-749f-4dc9-9b1d-daf88d6adcaf)
![image](https://github.com/fkie-cad/FACT_core/assets/26828709/bf2237a4-66e5-4c8e-b5e3-75077c789c90)
![Uploading image.png…]()

jstucke:
Hi, from what I can tell from the logs it seems like the file you are trying to upload as firmware is already in the database as unpacked file from another firmware (this is currently not possible and may cause errors). Could this be the case?

jstucke:
What is more, the file will never leave "currently analyzed firmware" because this file "got lost" during scheduling: All unpacking and analysis tasks should actually be completed but it will not get updated because of the exception. You need to restart FACT to get rid of the entry.

Issue #1145, opened 2023-10-17
Issue when uploading large files: invalid HTTP request size (max 4096)


paulrichardo:
### The FACT version you are using

a31382a00173fa54df32dc73411833d38540cceb

### Your question

Hello,
I'm starting the server with:
~/Desktop/FACT/FACT_core$ ./start_all_installed_fact_components

All runs fine. I use python requests library with FACT's REST API.
When I query for
http://localhost:5000/rest/status
I get fine response and I see in the window of "start_all_installed_fact_components" that my GET request was processed fine.
Though when I initiate a PUT request to
http://localhost:5000/rest/firmware
with a file of 15MB in size I get 
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
in python and I see in "start_all_installed_fact_components" window an error:
invalid HTTP request size (max 4096)

After some research I understand that the problem is with uWSGI that it gets large requests. So, I used the 'uwsgi_config.ini' to set the 'buffer-size=65535' to the maximum. But again I have the same error, with the new buffer size. Then I tried to change the nginx.conf settings, but nothing seems to help.

How can I fix this error for large files?

jstucke:
Hi and thank you for reporting this. I tried to reproduce your error but without success. Do you have nginx enabled? Could you try it again with this script and see if you get the same error? I submitted a 64 MB file without issues with the script.

```python3
#!/usr/bin/env python3
from __future__ import annotations

from argparse import ArgumentParser
from base64 import b64encode
from pathlib import Path

import requests


def parse_args():
    parser = ArgumentParser(description="FACT REST firmware upload script")
    parser.add_argument("file", type=Path, help="the file to upload")
    parser.add_argument("--device_class", "-c", type=str, help="the device class", default="test_class")
    parser.add_argument("--device_name", "-n", type=str, help="the device name", default="test_device")
    parser.add_argument("--vendor", "-v", type=str, help="the device name", default="test_vendor")
    parser.add_argument("--version", "-V", type=str, help="the device name", default="1")
    return parser.parse_args()


def main():
    args = parse_args()
    data: dict[str, str | list[str]] = {
        "binary": b64encode(args.file.read_bytes()).decode(),
        "device_class": args.device_class,
        "device_name": args.device_name,
        "device_part": "complete",
        "file_name": args.file.name,
        "release_date": "1970-01-01",
        "requested_analysis_systems": [],
[...]

PR #1151, opened 2023-10-23
handle exception when part of a fw is reuploaded as fw

jstucke:
- Handle the exception that occurs when trying to upload a file as firmware that is already in the DB as included files of another firmware
  - log error instead of passing the file to the analysis scheduler
  - this should handle the exceptions from #1149 and #1145


maringuu (approved):
While I would like to mention that I really don't like that we have this problem and it should be fixed, this workaround lgtm.

PR #1152, opened 2023-10-25
Exclude Fortify Source Check

lennarthenke:
The `checksec` tool's Fortify Source check should be excluded when performing offline checks, as it doesn't function as expected in this context. The Fortify Source check operates by using the running system's libraries to check for the supported Fortify-able library functions. In the case of firmware, this approach is not applicable as the binary and the running system's libraries are not the same. Therefore, the Fortify Source check is removed for offline checks. ([source](https://slimm609.github.io/checksec.sh/))

There is a possibility that in the future, this check may be reinstated if a viable workaround is identified. This could involve leveraging the firmware's own libc through mechanisms like chroot, for instance.


jstucke (changes requested):
Looks good except for one small detail: could you bump the version number to 0.2.0?

maringuu:
Could you add the description of the PR in the commit message?

jstucke (approved):

Issue #1154, opened 2023-10-27
why process stop 


chenjianquan7:
### FACT version

4.1

### Environment

ubuntu

### Steps to reproduce

![image](https://github.com/fkie-cad/FACT_core/assets/26828709/e4ad77b5-f224-4834-a500-98444252b2cf)
why main process stop at 99.8 and other process at over 100?

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 444, in _error_catcher
    yield
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 828, in read_chunked
    self._update_chunk_length()
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 758, in _update_chunk_length
    line = self._fp.fp.readline()
  File "/usr/lib/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 816, in generate
    yield from self.raw.stream(chunk_size, decode_content=True)
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 624, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 857, in read_chunked
    self._original_response.close()
  File "/usr/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python3.8/dist-packages/urllib3/response.py", line 449, in _error_catcher
    raise ReadTimeoutError(self._pool, None, "Read timed out.")
urllib3.exceptions.ReadTimeoutError: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/chenjianquan/FACT_core-0829/src/helperFunctions/process.py", line 87, in run
    Process.run(self)
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/chenjianquan/FACT_core-0829/src/analysis/PluginBase.py", line 201, in process_next_object
    finished_task = self.analyze_file(task)
[...]

PR #1155, opened 2023-10-27
helper docker: handle/log more connection errors

jstucke:
- docker helper function: handle/log other requests connection exceptions (see #1154)


maringuu (approved):
lgtm

PR #1156, opened 2023-10-30
cwe-checker: add memory limit

jstucke:
- adds an option for configuring a memory limit for the `cwe_checker` plugin
  - sets a default value of 4GB


Enkelmann (approved):
Looks good to me. Although it could be unintuitive to users that removing the limits from the config toml will not actually remove the memory limits for the Docker containers. But right now this is probably for the best, as running the cwe_checker without any memory limits is currently not advisable.

PR #1162, opened 2023-11-02
better sanitiziation of analysis results

### LINKED ISSUE #1161: Database error when trying to write to the database
### FACT version

4.2dev

### Environment

Ubuntu20

### Steps to reproduce

image
image
image
image

### Expeced Behavior

finish

### Installation logs

<details>
    <summary>install.log</summary>

```
PASTE HERE
```
</details>

### Backend logs

<details>
    <summary>fact_main_backend.log</summary>

```
PASTE HERE
```
</details>

### Frontend logs

<details>
    <summary>fact_main_frontend.log</summary>

```
PASTE HERE
```
</details>

### Other information

_No response_

jstucke:
resolves #1161

### DIFF
--- a/src/storage/entry_conversion.py
+++ b/src/storage/entry_conversion.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import json
 import logging
+import re
 from datetime import datetime
 from time import time
 
@@ -8,6 +10,8 @@ from helperFunctions.data_conversion import convert_time_to_str
 from objects.file import FileObject
 from objects.firmware import Firmware
 from storage.schema import AnalysisEntry, FileObjectEntry, FirmwareEntry, VirtualFilePath
+
+JSON_UNICODE_REGEX = re.compile(r'\\u[0-9a-f]{4}')
 
 
 def firmware_from_entry(fw_entry: FirmwareEntry, analysis_filter: list[str] | None = None) -> Firmware:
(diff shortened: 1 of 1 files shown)


src/storage/entry_conversion.py
```diff
@@ -142,17 +146,23 @@ def _sanitize_value(analysis_data: dict, key: str, value):
...
-    string = string.replace('\0', '')
     try:
         string.encode()
     except UnicodeEncodeError:
         string = string.encode(errors='replace').decode()
+    # replace all characters that are converted to unicode characters in JSON, because unicode characters can't be
+    # saved in the PostgreSQL database
+    json_string = json.dumps(string)
+    if JSON_UNICODE_REGEX.search(json_string):
+        logging.warning(f'Sanitizing unicode characters in string {json_string[100:]}')
```
maringuu, line 157: Why the `[100:]` here?

maringuu (changes requested):
A general question: Isn't this sanitation in the wrong place?
Should it be needed at all?

As far as I understand plugin results can contain invalid things, so they are sanitized.
But shouldn't this sanitation be done by the plugin?
How does the sanitation make sure that the plugin results stay valid?

Regarding #1161 this fix would make the result invalid, wouldn't it?

src/storage/entry_conversion.py
jstucke: The string could be very long (there are some big analysis results which are stored as strings) and I thought not printing a few MBs worth of string would generally be a good idea

jstucke:
> Isn't this sanitation in the wrong place?
Should it be needed at all?
As far as I understand plugin results can contain invalid things, so they are sanitized.
But shouldn't this sanitation be done by the plugin?

I generally agree but it isn't that easy to sanitize the string (it is a regular string but with unicode characters, so it is in fact valid JSON -- just the database can't handle it) and I don't think this should be handled by each plugin individually either. Also we have the potential problem of custom plugins which we can't control. Maybe it could be handled in the analysis plugin base class but I don't know if that is a better place than the module which explicitly handles data conversion for the database

jstucke:
This seems to only be an issue if the database encoding is not set to "UTF8". My database has this encoding, so only null bytes are a problem (which should already be handled). We need to determine in which cases the database encoding is not set to "UTF8" by default. Until then I will mark this PR as draft

dorpvom:
Has not become an issue. Will close.

wuzhongjie1997:
@dorpvom please see my reply in original issue.
I'd also like to ask, if there is a way to collect all files against which a certain analysis plugin failed to operate? Say i frequently experience 'unable to write database' issues with the `crypto_hints` plugin, I wonder if pulling all the extracted files which caused this error would aid in debugging the issue.

PR #1166, opened 2023-11-07
cwe-checker: remove incompatible type x-object from whitelist

jstucke:
- removes objects of the type `application/x-object` from the MIME whitelist of the CWE Checker plugin because they are not supported


Enkelmann (approved):
LGTM

PR #1176, opened 2023-11-22
Bump werkzeug to 2.3.8 to fix slow uploads of large files via the UI

### LINKED ISSUE #1144: Bottleneck in upload process, seemingly within uWSGI
### The FACT version you are using

6cb9354d6ea5160bc56374485db42608ac1b4424

### Your question

I'm currently running version 4.3 on Ubuntu 22.04, in a VMWare VM with 12 GiB ram, 5 processors, and 40 GiB of disk space.

When working with moderately large files (500 MiB - 1 Gib) I've observed a major bottleneck in getting a file uploaded into FACT, well before any analyses take place, on the scale of 2.5 hours for a 1 GiB filesystem. 

I've done some of my own sleuthing to try and find out where this holdup is happening, and it looks like the issue lies somewhere in the act of getting the backend all the data. I added some print statements and timestamps within `./src/web_interface/components/io_routes.py` to see what functions might be causing the holdup: 

image

I've collected some logs with these print lines. The below logs are taken from trying to upload an arbitrary 1 GiB filesystem made through buildroot, which I can share if interested:

```
-----
2023-09-25 09:31:30.467716 BEGINNING POST UPLOAD
-----

2023-09-25 09:31:30.467781 BEGINNING create_analysis_task
2023-09-25 09:31:30.467834 inside create_analysis_task

[...]
[...]
[...]

2023-09-25 10:51:40.033421 got meta_from_request
2023-09-25 10:51:41.258997 got binary from request
2023-09-25 10:51:41.822617 got uid of task
2023-09-25 10:51:41.822673 FINISHING create_ analysis_task

2023-09-25 10:51:41.822703 BEGINNING convert_analysis_to_fw_obj
2023-09-25 10:51:44.209959 FINISHING convert_analysis_task_to_fw_obj

2023-09-25 10:51:44.210020 BEGINNING add_analysis_task
2023-09-25 10:51:46.507327 FINISHING add_analysis_task

-----
2023-09-25 10:51:46.507380 FINISHING POST UPLOAD
-----
```

I found it strange that it would execute parts of the HTTP route, but then hang for a couple hours. It appears that the hanging starts whenever the route tries to do anything with the `request` object, such as accessing `request.form['device_name']`. Once we are able to do one of these accesses, though, we're clearly able to do everything immediately after as well.

I'm interested in knowing whatever suspicions you might have about this. Could it be that data getting *into* uWSGI is a bottleneck, or some computation *within* it prevents the data from being passed along to the endpoint? I would appreciate your thoughts. Thank you, as well!

hairlessbear:
I'm pretty confident that this fixes #1144. I experienced the same behavior described there (i.e. slow uploads of large files, but only when performed via the web UI). Updating to werkzeug 2.3.8 fixed the issue.

#1153 will also address this problem, but since there's no ETA on when a compatible flask-restx will be released, I think it makes sense to bump to 2.3.8 in the meantime. The slow multipart parsing is [the only change 2.3.8 contains](https://github.com/pallets/werkzeug/pull/2804), so there should be no compatibility problems.

### DIFF
--- a/src/install/requirements_frontend.txt
+++ b/src/install/requirements_frontend.txt
@@ -17,7 +17,7 @@ uwsgi==2.0.22
 virtualenv
 
 # must be below dependent packages (flask, flask-login, flask-restx)
-werkzeug==2.3.7
+werkzeug==2.3.8
 
 # Used for username validation by flask-security
 bleach==5.0.1
--- a/src/install/requirements_pre_install.txt
+++ b/src/install/requirements_pre_install.txt
@@ -6,7 +6,7 @@ python-magic==0.4.27
 requests==2.31.0
 # Needed by config.py
 pydantic==2.1.1
-werkzeug==2.3.4
+werkzeug==2.3.8
 toml==0.10.2
 
 git+https://github.com/fkie-cad/common_helper_files.git


jstucke (approved):
LGTM

jstucke:
Thank you for your contribution!

PR #1181, opened 2023-12-15
print out fact version + git hash on start

jstucke:
- log fact version, git commit hash and python version on startup of each component
  - should help debugging some of the bare logs we get in GitHub issues

### DIFF
--- a/src/fact_base.py
+++ b/src/fact_base.py
@@ -3,12 +3,13 @@ import os
 import signal
 import sys
 from pathlib import Path
+from shlex import split
+from subprocess import CalledProcessError, PIPE, run, STDOUT
 from time import sleep
 
 import config
 
 try:
-    import git
     import psutil
     import psycopg2  # noqa: F401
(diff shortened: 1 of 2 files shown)


jstucke:
> Depending on gitpython installed (which in turn requires git) makes FACT even less packagable. This is not a problem per se, as FACT currently cannot be packaged anyway but I think we should have packaging in mind. I'd prefer if we dropped the gitpython dependency and use just git. In addition, FACT should not crash if git is not installed.

Why should gitpython be a problem (since it is a pip package)? Also I think we can assume that git is installed if FACT is installed (since it gets used a lot during the installation and is explicitly installed in the `pre_install.sh`).

maringuu:
> Why should gitpython be a problem but not git [...]?

It is not a problem per-se, but I do not like adding dependencies for stuff that can be done in one line without the dependency.

maringuu (approved):

PR #1188, opened 2024-01-09
advanced search: allow search on multiple plugin results

### LINKED ISSUE #1186: Advanced serach cannot intersect multiple analysis
This does not produce any results, despite there being thousands of files that match this query.
Note that it works if we leave out the query for the `file_type` plugin.
Also these kinds of queries should be documented.
```
{
    "processed_analysis.software_components.OpenSSL": {
        "$exists": "true"
    },
    "processed_analysis.software_components.OpenSSL.meta.version": {
        "$exists": "true"
    },
    "processed_analysis.file_type.mime": "application/x-executable"
}
```

jstucke:
resolves #1186

### DIFF
--- a/src/storage/query_conversion.py
+++ b/src/storage/query_conversion.py
@@ -10,6 +10,7 @@ from sqlalchemy.orm import aliased
 from storage.schema import AnalysisEntry, FileObjectEntry, FirmwareEntry
 
 if TYPE_CHECKING:
+    from sqlalchemy.orm.util import AliasedClass
     from sqlalchemy.sql import Select
 
 FIRMWARE_ORDER = FirmwareEntry.vendor.asc(), FirmwareEntry.device_name.asc()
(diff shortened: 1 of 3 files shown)


maringuu (changes requested):
Please add some tests to make sure this api does not regress.
Also this lacks documentation. I'd suggest to add this alongside the examples already displayed on the advanced search pages.

jstucke:
> Please add some tests to make sure this api does not regress. Also this lacks documentation. I'd suggest to add this alongside the examples already displayed on the advanced search pages.

I added an example to the template and a test.

maringuu:
@dorpvom @rhelmke @jstucke Im hesitating to merge this because I'm not sure what the long term plan is with the 'advanced serach'.
Do we want to have it at all? Will it be replaced with the recently merged GraphQL API?

jstucke:
> @dorpvom @rhelmke @jstucke Im hesitating to merge this because I'm not sure what the long term plan is with the 'advanced serach'. Do we want to have it at all? Will it be replaced with the recently merged GraphQL API?

Good question. The long term plan definitely is to replace the Advanced Search with GraphQL but we will probably leave it as deprecated for a while so that old REST scripts don't break immediately

dorpvom:
Closing because feature will be deprecated soon.

PR #1189, opened 2024-01-09
Ruff format

### LINKED ISSUE #1187: Use ruff format instead of black
We use ruff anyways so we can drop one dependency.

jstucke:
- updated ruff version and rules
- replaced black with ruff
- checked, fixed and reformatted code with new ruff version

resolves #1187


pyproject.toml
```diff
@@ -49,19 +44,25 @@ markers = [
...
     # pydantic only supports these from python>=3.9
     "UP006",
     "UP007",
+    # rules may cause conflicts when used with the formatter
+    "ISC001",
+    "Q001",
 ]
-fixable = ["F", "E", "W", "C90", "N", "UP", "B", "A", "PL", "C4", "EXE", "ISC", "PIE", "T20", "PT", "RET", "SIM", "TCH",
-    "PTH", "PLR", "PLW", "ARG", "ERA", "RUF", "Q"]
+fixable = ["F", "E", "W", "C90", "N", "UP", "B", "A", "C4", "EXE", "FA", "ISC", "PIE", "T20", "PT", "Q", "RET", "SIM",
```
maringuu, line 64: While we are at it, can we replace this by `ALL`?

pyproject.toml
```diff
@@ -49,19 +44,25 @@ markers = [
 ]
 
 [tool.ruff]
-select = ["F", "E", "W", "C90", "N", "UP", "B", "A", "PL", "C4", "EXE", "ISC", "PIE", "T20", "PT", "RET", "SIM", "TCH",
-    "PTH", "PLR", "PLW", "ARG", "ERA", "RUF", "Q"]
+select = ["F", "E", "W", "C90", "N", "UP", "B", "A", "C4", "EXE", "FA", "ISC", "PIE", "T20", "PT", "Q", "RET", "SIM",
```
maringuu, line 47: Could you convert this to a sorted list with one line per entry?

.pre-commit-config.yaml
```diff
@@ -43,13 +43,9 @@ repos:
...
     -   id: jshint
 
--   repo: https://github.com/ambv/black
-    rev: 23.3.0
-    hooks:
-    -   id: black
-
 - repo: https://github.com/charliermarsh/ruff-pre-commit
-  rev: 'v0.0.275'
+  rev: 'v0.1.11'
```
maringuu, line 47: Since it took me so long to review this, we can bump the version once again.

pyproject.toml
```diff
@@ -1,8 +1,3 @@
-[tool.black]
-line-length = 120
-skip-string-normalization = true
-target-version = ['py38']
-
 [tool.pylint.main]
```
maringuu, line 1: While not strictly related to this PR, can you remove this?

maringuu (changes requested):

pyproject.toml
jstucke: I don't know if sorting it alphabetically is actually helpful: Right now it is ordered the same way as the rules in [the Ruff docs](https://docs.astral.sh/ruff/rules/) which makes it easier to see changes in new ruff versions

pyproject.toml
jstucke: I tried to merge it with the ruff configuration where possible and removed the rest

maringuu:
Do not wonder, I prettified the history a bit :)

maringuu (approved):

PR #1219, opened 2024-05-02
unify nodejs installation with nodeenv

jstucke:
Updated nodejs version from 18 to 22 ([18 is no longer actively supported](https://endoflife.date/nodejs)). Installing nodejs with nodeenv has the advantage that the installed nodejs cannot clash with a possible pre-installed version in the system. What is more, it makes the installation easier (especially updating and changing the version).

### DIFF
--- a/src/install/frontend.py
+++ b/src/install/frontend.py
@@ -1,9 +1,14 @@
 import logging
 import os
+import re
+import shutil
 import subprocess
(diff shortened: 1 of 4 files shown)


dorpvom (commented):
Changes look good. I'll test on a VM to confirm.

dorpvom (approved):

PR #1224, opened 2024-06-03
Mypy: fix helper functions

jstucke:
supersedes #1123 and #1142

- added mypy config to `pyproject.toml`
- fixed mypy type issues in all modules in `helperFunctions`

this does not include the pre-commit hook yet (otherwise all other builds would fail after merging this PR)


maringuu:
In an effort to separate commits that actually change codes from commits that only change type hints, I added 41e36accf6849275608768f4a073185c4aae4320. Would you mind improving the commit message (at least removing the WIP prefix)?

PR #1225, opened 2024-06-17
db schema: added result GIN index + migration

jstucke:
- added GIN index to speed up certain queries on the JSONB results column of table analysis
- added alembic migration script

### DIFF
--- a/src/plugins/analysis/crypto_hints/code/crypto_hints.py
+++ b/src/plugins/analysis/crypto_hints/code/crypto_hints.py
@@ -1,15 +1,18 @@
-import io
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, List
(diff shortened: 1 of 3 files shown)


rhelmke (approved):
These changes provide an _enormous_ speedup for database queries that touch the analysis results and use postgres json operators. The benchmark below was executed on a FACT instance with over 10k firmware images.

Pre:
```sql
SELECT analysis.uid, analysis.plugin
FROM analysis
WHERE analysis.result ? :result_1
has_key took 252.0058159828186 sec
```

Post: 
```sql
SELECT analysis.uid, analysis.plugin
FROM analysis
WHERE analysis.result ? :result_1
has_key took 0.5301656723022461 sec
```

PR #1228, opened 2024-06-25
bug fix: pin recusive dependency numpy to <2

jstucke:
The recently released version 2 of numpy causes weird import errors because matplotlib is not yet compatible

### DIFF
--- a/src/install/requirements_frontend.txt
+++ b/src/install/requirements_frontend.txt
@@ -16,6 +16,9 @@ quantiphy~=2.20
 uwsgi~=2.0.25.1
 virtualenv~=20.26.1
 
+# FixMe: dependency of matplotlib which does not work with numpy>=2 yet
+numpy<2
+
 # must be below dependent packages (flask, flask-login, flask-restx)
 werkzeug~=3.0.3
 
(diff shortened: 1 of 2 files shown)


maringuu:
Why is this our job?
Shouldn't this be part of matplotlib's dependency specification.

jstucke:
> Why is this our job? Shouldn't this be part of matplotlib's dependency specification.

You are right, normally this shouldn't be our job. The CI was failing because of this, though: https://github.com/fkie-cad/FACT_core/actions/runs/9655874153/job/26632485364

But since then, there were many successful CI jobs without this fix, so I don't know if we need this after all? Maybe it could also be related to caching pip packages since there was this warning in the build log:

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
```

So I will close this PR for now.

PR #1231, opened 2024-06-28
postgres comments

jstucke:
- added table and column comments for documentation purposes and for displaying in the GraphiQL docs

ToDo:
- [x] generate alembic migration script

### DIFF
--- a/src/storage/schema.py
+++ b/src/storage/schema.py
@@ -22,6 +22,7 @@ from sqlalchemy.orm import Session, backref, declarative_base, mapped_column, re
 Base = declarative_base()
 UID = VARCHAR(78)
 
+
 # primary_key=True implies `unique=True` and `nullable=False`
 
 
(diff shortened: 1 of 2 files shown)


src/storage/schema.py
```diff
@@ -28,20 +28,57 @@
...
-    summary = mapped_column(ARRAY(VARCHAR, dimensions=1), nullable=True)
-    tags = mapped_column(MutableDict.as_mutable(JSONB), nullable=True)
-    result = mapped_column(MutableDict.as_mutable(JSONB), nullable=True)
+    uid = mapped_column(
+        UID,
+        ForeignKey('file_object.uid', ondelete='CASCADE'),
+        index=True,
+        comment='The UID of the file object on which this analysis was performed.',
+    )
+    plugin = mapped_column(VARCHAR(64), nullable=False, comment='The name of the analysis plugin (e.g. "file_type").')
```
maringuu, line 37: As a personal style preference, I like to put every argument on its own regardless of the linelength.
Would you mind splitting the lines here too?

dorpvom (approved):

PR #1234, opened 2024-07-03
backend start script: check ulimit and increase soft limit if possible

jstucke:


maringuu:
I think this code would be a good place to document _why_ we need to increase the limit.

jstucke:
> I think this code would be a good place to document _why_ we need to increase the limit.

I added a docstring to the `_check_ulimit` function

src/start_fact_backend.py
```diff
@@ -110,6 +114,31 @@ def _exception_occurred(self):
...
 
 
+def _check_ulimit():
+    """
+    Each process has a hard limit and a soft limit for the maximum number of files opened at the same time. Since
+    FACT makes extensive use of multiprocessing features, it uses up a lot of those file descriptors and if we run
+    out, this raises an OSError. To mitigate this, we try to increase the soft limit and print a warning if the
+    hard limit is low. With the default configuration, FACT uses about 560 file descriptors (and potentially many
+    more if you crank up the worker counts).
+    """
```
maringuu, line 124: Can specify which multiprocessing features use file descriptors?
While I know that its queues, I would not expect the average reader to think of queues when thinking about multiprocessing.

Also, can you document the calculation that lead to 560 descriptors by default?

jstucke: I extended the docstring quite a bit and added information on how the FDs are distributed

maringuu (approved):

PR #1241, opened 2024-07-29
plugin users&pws: ported to new base class

jstucke:


src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -6,113 +6,158 @@
...
-        file_object.processed_analysis[self.NAME]['summary'].extend(list(result_entry))
-        self._add_found_password_tag(file_object, result_entry)
-
-
-def generate_unix_entry(entry: bytes) -> dict:
+class CredentialResult(pydantic.BaseModel):
+    username: str = Field(description='The username.')
+    full_entry: str = Field(description='The full entry in unparsed form.')
+    type: str = Field(description='The type of credential (UNIX, htpasswd, etc.).')
+    # FixMe: update type hint when we drop support for python <3.10
```
maringuu, line 48: I think this comment is made obsolete by pyupgrade which changes typehints automatically.

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -6,113 +6,158 @@
...
+    password_hash: Optional[str] = Field(description='The password in hashed form.', default=None)
+    password: Optional[str] = Field(description='The password (if the hash was available and cracked).', default=None)
+    error: Optional[str] = Field(
+        description='Error message (if cracking the password hash was not successful).', default=None
```
maringuu, line 52: nit: Please add a trailing comma.

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -6,113 +6,158 @@
...
+    def analyze(self, file_handle: FileIO, virtual_file_path: dict[str, list[str]], analyses: dict) -> Schema:
+        del virtual_file_path, analyses
+        file_contents = Path(file_handle.name).read_bytes()
+        credentials: list[CredentialResult] = []
+        for regex_list, entry_gen_function in [
+            (UNIX_REGEXES, generate_unix_entry),
+            (HTPASSWD_REGEXES, generate_htpasswd_entry),
+            (MOSQUITTO_REGEXES, generate_mosquitto_entry),
+        ]:
+            credentials.extend(find_password_entries(file_contents, regex_list, entry_gen_function))
```
maringuu, line 82: I personally find this less readable than not using a for loop. Also not using a for-loop is three lines less.

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -6,113 +6,158 @@
...
+    password_hash: Optional[str] = Field(description='The password in hashed form.', default=None)
+    password: Optional[str] = Field(description='The password (if the hash was available and cracked).', default=None)
+    error: Optional[str] = Field(
+        description='Error message (if cracking the password hash was not successful).', default=None
+    )
+
+
+class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
+    class Schema(pydantic.BaseModel):
+        credentials: List[CredentialResult] = Field(description='The list of found credentials.')
```
maringuu, line 58: I'd propose a different schema:

```python
class Schema(pydantic.BaseMode):
    unix: list[CreadentialResult]
    httpd: list[CredentialResult]
    mosquito: list[CredentialResult]
```

The class `CrednetialResult` would not have the `type` field anymore.

This way we could delete the `_group_user_and_pw_results` function and the schema would be more expressive imo.

maringuu (changes requested):

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
```diff
@@ -6,113 +6,158 @@
...
-    def update_file_object(self, file_object: FileObject, result_entry: dict):
-        file_object.processed_analysis[self.NAME].update(result_entry)
-        file_object.processed_analysis[self.NAME]['summary'].extend(list(result_entry))
-        self._add_found_password_tag(file_object, result_entry)
-
-
-def generate_unix_entry(entry: bytes) -> dict:
+class CredentialResult(pydantic.BaseModel):
+    username: str = Field(description='The username.')
+    full_entry: str = Field(description='The full entry in unparsed form.')
```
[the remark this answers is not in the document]
jstucke: "full_entry" is a full, unparsed entry, e.g.

```
admin:$5$Mqs9JzinZfyjSMNR$M7DWqlMjKmOHwfzXnts89eQImN2QNDATeN/Z/BX0Jb.:12332:0:99999:7:::
```

or if it doesn't contain a pw hash:

```
admin:*:12332:0:99999:7:::
```

This can be interesting if something went wrong (or sees to have gone wrong) during parsing or if the result is empty (because there is no hash)

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
jstucke: The `find_password_entries` was also ugly and kinda useless, since it had only one line. I refactored it in a way that made more sense to me

src/plugins/analysis/users_and_passwords/code/password_file_analyzer.py
jstucke: Changing the schema in this way also has disadvantages (like making it more complicated to add more password types or making it more complicated to iterate through all found pw entries) but I changed it (minus the typos :sweat_smile:)

jstucke: I kept the type field, though, because without it, summary generation becomes unnecessarily ugly.

maringuu (approved):

PR #1245, opened 2024-08-08
Update postgres to v17

jstucke:
- update postgres from 14 to 17
- also fixed bugs:
  - fixed a bug in the init postgres script when the local postgres instance does not run on the default port
  - fixed a copy&paste bug in a test fixture

### DIFF
--- a/src/install/db.py
+++ b/src/install/db.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from contextlib import suppress
 from pathlib import Path
 from shlex import split
@@ -6,8 +7,10 @@ from subprocess import PIPE, CalledProcessError, run
 
 from helperFunctions.install import InstallationError, OperateInDirectory, check_distribution
 
+POSTGRES_VERSION = 16
 
-def install_postgres(version: int = 16):
+
+def install_postgres(version: int = POSTGRES_VERSION):
     # based on https://www.postgresql.org/download/linux/ubuntu/
     codename = check_distribution()
     command_list = [
@@ -21,7 +24,7 @@ def install_postgres(version: int = 16):
             raise InstallationError(f'Failed to set up PostgreSQL: {process.stderr}')
 
 
-def configure_postgres(version: int = 14):
+def configure_postgres(version: int = POSTGRES_VERSION):
     config_path = f'/etc/postgresql/{version}/main/postgresql.conf'
     # increase the maximum number of concurrent connections
     run(f'sudo sed -i -E "s/max_connections = [0-9]+/max_connections = 999/g" {config_path}', shell=True, check=True)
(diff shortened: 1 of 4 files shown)


src/install/db.py
```diff
@@ -4,14 +4,15 @@
 from shlex import split
 from subprocess import PIPE, CalledProcessError, run
 
-from helperFunctions.install import InstallationError, OperateInDirectory
+from helperFunctions.install import InstallationError, OperateInDirectory, check_distribution
 
 
-def install_postgres(version: int = 14):
+def install_postgres(version: int = 16):
```
dorpvom, line 10: This has to also be used for the `postgres_is_installaed` function, since no ugrade to 16 will be done on existing 14 installation.

dorpvom (changes requested):

src/install/db.py
jstucke: I added a version check with a warning it the installed version is outdated

src/install/db.py
```diff
@@ -39,8 +43,24 @@ def postgres_is_installed():
         return False
 
 
+def postgres_is_up_to_date():
+    with suppress(CalledProcessError, FileNotFoundError):
+        proc = run(split('psql --version'), text=True, capture_output=True, check=False)
```
dorpvom, line 48: Ok. I get that this is super nitpicky, but there is really no need to call `psql --version` twice :wink:

src/install/db.py
```diff
@@ -39,8 +43,24 @@ def postgres_is_installed():
...
+    logging.warning('PostgreSQL version could not be identified. Is it installed?')
+    return True
+
+
 def main():
     if postgres_is_installed():
+        if not postgres_is_up_to_date():
+            logging.warning(
+                'PostgreSQL is installed but the version is not up to date. Please see '
+                'https://github.com/fkie-cad/FACT_core/wiki/Upgrading-the-PostgreSQL-Database for information on how'
```
dorpvom, line 61: Set link into some sort of quote please

src/install/db.py
```diff
@@ -1,17 +1,21 @@
...
+import re
 from contextlib import suppress
 from pathlib import Path
 from shlex import split
 from subprocess import PIPE, CalledProcessError, run
 
-from helperFunctions.install import InstallationError, OperateInDirectory
+from helperFunctions.install import InstallationError, OperateInDirectory, check_distribution
 
+POSTGRES_VERSION = 16
```
dorpvom, line 10: I think we discussed it should be 17

dorpvom (changes requested):

src/install/db.py
jstucke: removed one call

dorpvom (approved):
In the process of checking if the update works correctly

PR #1247, opened 2024-08-20
crypto material plugin v1

jstucke:
- converted `crypto_material` plugin to new base class
- [x] requires #1248 for the links in the template to work


euwint (approved):
Looks good to me

PR #1251, opened 2024-08-27
intercom backend refactoring: make data flow comprehensible

jstucke:


src/intercom/back_end_binding.py
```diff
@@ -34,90 +33,66 @@ def __init__(
...
+            InterComBackEndAnalysisTask(self.unpacking_service.add_task),
+            InterComBackEndReAnalyzeTask(self.unpacking_service.add_task),
+            InterComBackEndCompareTask(self.compare_service.add_task),
+            InterComBackEndRawDownloadTask(),
+            InterComBackEndFileDiffTask(),
+            InterComBackEndTarRepackTask(),
+            InterComBackEndBinarySearchTask(),
+            InterComBackEndUpdateTask(self.analysis_service.update_analysis_of_object_and_children),
+            InterComBackEndDeleteFile(
+                self.analysis_service.update_analysis_of_object_and_children,
```
dorpvom, line 51: Why do I need to `update_analysis_of_object_and_children` if I want to delete a firmware?

src/intercom/common_redis_binding.py
```diff
@@ -1,31 +1,63 @@
...
+        self.process = None
+        self.processing_function = processing_function
+        self.stop_condition = Value('i', 0)
+
+    def start(self):
+        self.process = Process(target=self._worker)
+        self.process.start()
+
+    def shutdown(self):
+        self.stop_condition.value = 1
```
dorpvom, line 49: I guess this plays into the _make data flow comprehensible_ goal, but this does create another dozen or so file descriptors compared to sharing one stop condition over all listeners, right?

dorpvom (commented):
Consider questions and give me some feedback

src/intercom/back_end_binding.py
jstucke: Good catch. This is obviously wrong. Maybe it snuck in during a rebase/merge

src/intercom/common_redis_binding.py
jstucke: I also wondered about that and thought about using one central value, that is passed to the individual listeners, but the `multiprocessing.Value` objects by themselves do not seem to use additional file descriptors

PR #1252, opened 2024-08-29
wait for single file analysis (update)

jstucke:
- makes intercom backend listener "responder" asynchronous (so it does not block when waiting for the result)
- converts the "single file analysis task" listener into a responder
- changes the show analysis page so that single file analyses are awaited and the page is automatically reloaded
- adds a button to the "analysis is outdated" message to quickly update it

- [x] based on #1251 so that one should be merged first

### DIFF
--- a/src/intercom/back_end_binding.py
+++ b/src/intercom/back_end_binding.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import difflib
 import logging
+from multiprocessing import Manager
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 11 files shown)


euwint (changes requested):
Looks good, but when I run pytest in _src/test/integration/intercom/_, the test does not terminate. This is probably due to a threading problem

jstucke:
> Looks good, but when I run pytest in _src/test/integration/intercom/_, the test does not terminate. This is probably due to a threading problem

should hopefully be fixed now

euwint (approved):
Problem is solved

PR #1255, opened 2024-09-02
Revised upload page

jstucke:
- removed `moment.js` package (apparently unused for quite a while)
- revised upload page:
  - replaced "select" elements on upload page with dropdown menus
  - inputs are always visible
  - changed the styling to be more coherent
  - removed validity check (it is not possible to send an invalid request using the unedited upload page because of the `required` tags)
    - it could be possible to send an invalid request by manually crafting one but that would only generate an error in the frontend
  - added optional filter inputs for device classes, names and vendors


maringuu:
Nice!

Can we make the right arrow trigger the list of all possible entries?
The dropdown search on the right looks redundant to me.
image

maringuu (approved):
In addition to the commend above, I have two request but I am approving anyways to not get this stuck another week.

1) Please make the button _toggle_ the list. See the example code linked in the above comment.
2) Please make the list scroll if it overflows.
For example, the following cs makes it show 10 items and add a scrollbar if it overflows.

```
.autocomplete-items {
         // [...]
	max-height: 1000%;
	overflow-x: hidden;
	overflow-y: auto;
}
```

jstucke:
> In addition to the commend below, I have two request but I am approving anyways to not get this stuck another week.
> 
>     1. Please make the button _toggle_ the list. See the example code linked in the above comment.
> 
>     2. Please make the list scroll if it overflows.
>        For example, the following cs makes it show 10 items and add a scrollbar if it overflows.

I think I found a way to fix both issues (see https://github.com/fkie-cad/FACT_core/pull/1255/commits/144dbb73099a2ee59e8e8ec09c59674fa1831a3f)

PR #1262, opened 2024-09-05
Hash plugin v1

jstucke:
- fixed a bug in AnalysisPluginV0 (summarize should not be abstract, since the summary is optional)
- migrated the hash plugin to the new base class

### DIFF
--- a/src/plugins/analysis/hash/code/hash.py
+++ b/src/plugins/analysis/hash/code/hash.py
@@ -1,43 +1,113 @@
+from __future__ import annotations
+
 import logging
 from hashlib import algorithms_guaranteed
+from typing import TYPE_CHECKING, Optional
+
+import lief
+import ssdeep
+import tlsh
+from pydantic import BaseModel, Field
+from semver import Version
 
 import config
-from analysis.PluginBase import AnalysisBasePlugin
-from helperFunctions.hash import get_hash, get_imphash, get_ssdeep, get_tlsh
+from analysis.plugin import AnalysisPluginV0
+from analysis.plugin.compat import AnalysisBasePluginAdapterMixin
+from helperFunctions.hash import get_hash, get_md5
+
+if TYPE_CHECKING:
+    from io import FileIO
+
+ELF_MIME_TYPES = [
+    'application/x-executable',
+    'application/x-object',
+    'application/x-pie-executable',
+    'application/x-sharedlib',
+]
 
 
-class AnalysisPlugin(AnalysisBasePlugin):
+class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
+    class Schema(BaseModel):
+        # The supported hashes are the ones from helperFunctions.hash and hashlib (except "shake" which is of
+        # little use considering its variable length).
+        # If they are not supported on the platform or not selected in the configuration of the plugin, the value will
+        # be `None`.
+        # Only the md5 and sha256 hashes are guaranteed to be available (since they are required down the line)
+
+        # from hashlib
+        md5: str = Field(description="md5 hash of the file's content")
+        sha256: str = Field(description="sha256 hash of the file's content")
+        sha1: Optional[str] = Field(description="sha1 hash of the file's content", default=None)
+        sha224: Optional[str] = Field(description="sha224 hash of the file's content", default=None)
+        sha384: Optional[str] = Field(description="sha384 hash of the file's content", default=None)
+        sha512: Optional[str] = Field(description="sha512 hash of the file's content", default=None)
+        blake2b: Optional[str] = Field(description="blake2b hash of the file's content", default=None)
+        blake2s: Optional[str] = Field(description="blake2s hash of the file's content", default=None)
+        sha3_224: Optional[str] = Field(description="sha3_224 hash of the file's content", default=None)
+        sha3_256: Optional[str] = Field(description="sha3_256 hash of the file's content", default=None)
+        sha3_384: Optional[str] = Field(description="sha3_384 hash of the file's content", default=None)
+        sha3_512: Optional[str] = Field(description="sha3_512 hash of the file's content", default=None)
+
+        ssdeep: Optional[str] = Field(description="ssdeep hash of the file's content", default=None)
+        tlsh: Optional[str] = Field(description="tlsh hash of the file's content", default=None)
+        imphash: Optional[str] = Field(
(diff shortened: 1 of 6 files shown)


src/plugins/analysis/hash/code/hash.py
```diff
@@ -1,43 +1,134 @@
...
+def get_imphash(file: FileIO, type_analysis: BaseModel | None) -> str | None:
     """
-    This Plugin creates several hashes of the file
+    Generates and returns the md5 hash of the (sorted) imported functions of an ELF file represented by `file_object`.
+    Returns `None` if there are no imports or if an exception occurs.
     """
+    if type_analysis is not None and _is_elf_file(type_analysis):
+        try:
+            with suppress_stdout():
+                functions = [f.name for f in lief.ELF.parse(file).imported_functions]
```
maringuu, line 96: Why not use `lief.logging.disable()`?

src/plugins/analysis/hash/test/test_plugin_hash.py
```diff
@@ -23,20 +31,51 @@
...
-        assert 'sha1' in result, 'sha1 not in result'
-        assert 'foo' not in result, 'foo in result but not available'
-        assert result['md5'] == '6f8db599de986fab7a21625b7916589c', 'hash not correct'
-        assert 'ssdeep' in result, 'ssdeep not in result'
-        assert 'imphash' in result, 'imphash not in result'
+        assert isinstance(result.md5, str)
+        assert isinstance(result.sha1, str), 'sha1 not in result'
+        assert isinstance(result.ssdeep, str), 'ssdeep not in result'
+        assert isinstance(result.imphash, str), 'imphash not in result'
+        assert not hasattr(result, 'foo')
```
maringuu, line 41: This is redundant, isn't it?
Also, what was this supposed to test in the first place?! Someone adding a hash named `foo` accidentally?

src/plugins/analysis/hash/code/hash.py
```diff
@@ -1,43 +1,134 @@
...
+        sha3_384: Optional[str] = Field(description="sha3_384 hash of the file's content", default=None)
+        sha3_512: Optional[str] = Field(description="sha3_512 hash of the file's content", default=None)
+
+        # from helperFunctions.hash
+        ssdeep: Optional[str] = Field(description="ssdeep hash of the file's content", default=None)
+        tlsh: Optional[str] = Field(description="tlsh hash of the file's content", default=None)
+        imphash: Optional[str] = Field(
+            description="import hash of the executable's imported functions (only for ELF files)",
+            default=None,
+        )
```
maringuu, line 58: Two things:
1) I do not think this belongs in this plugin but rather in an elf specific plugin. But since this is status quo this is fine for now.
2) This needs some clarification. How exactly are the functions hashed? Which hash functions is used?

src/plugins/analysis/hash/test/test_plugin_hash.py
```diff
@@ -23,20 +31,51 @@
...
-        assert 'md5' in result, 'md5 not in result'
-        assert 'sha1' in result, 'sha1 not in result'
-        assert 'foo' not in result, 'foo in result but not available'
-        assert result['md5'] == '6f8db599de986fab7a21625b7916589c', 'hash not correct'
-        assert 'ssdeep' in result, 'ssdeep not in result'
-        assert 'imphash' in result, 'imphash not in result'
+        assert isinstance(result.md5, str)
+        assert isinstance(result.sha1, str), 'sha1 not in result'
+        assert isinstance(result.ssdeep, str), 'ssdeep not in result'
+        assert isinstance(result.imphash, str), 'imphash not in result'
```
maringuu, line 40: Why not use e.g. `resut.md5 is not None`?
Also, I do not think that any of the strings are helpful.

src/plugins/analysis/hash/view/hash.html
```diff
@@ -0,0 +1,14 @@
...
+{% block analysis_result_details %}
+
+    {% for key, value in analysis_result.items() | sort %}
+        {% if value %}
+            <tr>
+                <td>{{ key }}</td>
+                <td style="font-family: monospace">{{ value }}</td>
+            </tr>
+        {% endif %}
+    {% endfor %}
```
maringuu, line 12: What do you think of explicitly listing all hashes here?
I think this improves readability.

maringuu (changes requested):

src/plugins/analysis/hash/code/hash.py
```diff
@@ -1,43 +1,134 @@
...
+        sha384: Optional[str] = Field(description="sha384 hash of the file's content", default=None)
+        sha512: Optional[str] = Field(description="sha512 hash of the file's content", default=None)
+        blake2b: Optional[str] = Field(description="blake2b hash of the file's content", default=None)
+        blake2s: Optional[str] = Field(description="blake2s hash of the file's content", default=None)
+        sha3_224: Optional[str] = Field(description="sha3_224 hash of the file's content", default=None)
+        sha3_256: Optional[str] = Field(description="sha3_256 hash of the file's content", default=None)
+        sha3_384: Optional[str] = Field(description="sha3_384 hash of the file's content", default=None)
+        sha3_512: Optional[str] = Field(description="sha3_512 hash of the file's content", default=None)
+
+        # from helperFunctions.hash
```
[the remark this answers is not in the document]
jstucke: removed the comment

src/plugins/analysis/hash/code/hash.py
jstucke: I removed the call and everything related (function, test)

src/plugins/analysis/hash/code/hash.py
jstucke: I tried to improve the wording of the description to make it more specific

src/plugins/analysis/hash/test/test_plugin_hash.py
jstucke: changed

src/plugins/analysis/hash/view/hash.html
jstucke: I don't like the idea: Not only is it a lot longer and very repetitive, it would also mean that we need to change the file each time we change the computed hashes

maringuu (approved):

maringuu:
@jstucke I'm sorry, this needs another rebase.

PR #1263, opened 2024-09-06
feat: converted ipc plugin to new base class

jstucke:


src/plugins/analysis/ipc/code/ipc_analyzer.py
```diff
@@ -3,44 +3,69 @@
...
+            name='ipc_analyzer',
+            dependencies=['file_type'],
+            description='Inter-Process Communication Analysis',
+            mime_whitelist=[
+                'application/x-executable',
+                'application/x-object',
+                'application/x-pie-executable',
+                'application/x-sharedlib',
+            ],
+            timeout=600,  # 10 minutes
```
maringuu, line 47: Unnecessary comment

src/plugins/analysis/ipc/code/ipc_analyzer.py
```diff
@@ -3,44 +3,69 @@
...
-    MIME_WHITELIST = [  # noqa: RUF012
-        'application/x-executable',
-        'application/x-object',
-        'application/x-sharedlib',
-    ]
-    DEPENDENCIES = ['file_type']  # noqa: RUF012
-    TIMEOUT = 600  # 10 minutes
+class IpcCall(BaseModel):
+    type: str = Field(description='a type of call (usually a POSIX system call)')
+    arguments: List[Union[str, int, List[Union[str, int]]]] = Field(description='arguments of the IPC call')
```
maringuu, line 29: I do not think this field is useful in its current form. The typehint is too complex.
Wouldn't a `list[Union[str, int]]` suffice? Why have a special case for single arguments?

src/plugins/analysis/ipc/code/ipc_analyzer.py
```diff
@@ -3,44 +3,69 @@
...
 
-    def _run_ipc_analyzer_in_docker(self, file_object: FileObject) -> dict:
+
+class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
+    class Schema(BaseModel):
+        targets: List[IpcTarget] = Field(description='An array of targets of IPC calls')
```
maringuu, line 34: Wouldn't it be useful to make this a dictionary?
E.g. to find out which IPC calls target a particular pipe one could simply check whether the path is in the targets dictionary.

src/plugins/analysis/ipc/test/test_ipc_analyzer.py
```diff
@@ -2,32 +2,36 @@
...
-    'echo': [{'type': 'system', 'arguments': ['hello']}],
-    'id': [{'type': 'system', 'arguments': ['']}],
-    'pwd': [{'type': 'system', 'arguments': ['']}],
+    'targets': [
+        {'name': 'whoami', 'ipc_calls': [{'type': 'system', 'arguments': ['']}]},
+        {'name': 'ls', 'ipc_calls': [{'type': 'system', 'arguments': ['-l']}]},
+        {'name': 'echo', 'ipc_calls': [{'type': 'system', 'arguments': ['hello']}]},
+        {'name': 'id', 'ipc_calls': [{'type': 'system', 'arguments': ['']}]},
+        {'name': 'pwd', 'ipc_calls': [{'type': 'system', 'arguments': ['']}]},
+    ]
```
maringuu, line 16: > target of the IPC (usually a file path or address)

How do these names (`whoami`, `ls`, etc.) match the documentation?

src/web_interface/filter.py
```diff
@@ -433,6 +433,13 @@ def _cve_score_to_float(score: float | str) -> float:
         return 0.0
 
 
+def sort_dict_list_by_key(dict_list: list[dict], key: Any) -> list[dict]:
+    types = {type(d.get(key)) for d in dict_list if key in d}
+    if types not in [set(), {str}, {int}]:
+        raise ValueError(f'Dict values must be of the same type, not {types}')
+    return sorted(dict_list, key=lambda d: d.get(key, -9999 if types == {int} else ''))
```
maringuu, line 440: The -9999 looks suspicious.

maringuu (changes requested):

src/plugins/analysis/ipc/code/ipc_analyzer.py
jstucke: I don't think so: JSON objects are not intended to be used with unknown keys -- that's what arrays are for.
> E.g. to find out which IPC calls target a particular pipe one could simply check whether the path is in the targets dictionary.
If you want to find them using a Postgres or GraphQL query, it doesn't really matter, because it works either way.

src/plugins/analysis/ipc/code/ipc_analyzer.py
jstucke: > Wouldn't a list[Union[str, int]] suffice?

No because it can be a str, an int, a list of strings and a list of ints (so the type hint wasn't even technically correct; the correct one would be `List[Union[str, int, List[str], List[int]]]`)

Maybe we can just use a `: list`? :see_no_evil:

maringuu (approved):

PR #1266, opened 2024-09-11
converted elf_analysis plugin to new base class

jstucke:


src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,233 @@
...
-            if str(sv) != '* Local *' and str(sv) != '* Global *':
-                imported_libs.append(str(sv).split('(', maxsplit=1)[0])
-        return list(set(imported_libs))
+def _get_modinfo(elf: lief.ELF) -> list[str] | None:
+    # getting the information from the *.ko files .modinfo section
+    modinfo = None
+    for section in elf.sections:
+        if section.name == '.modinfo':
+            modinfo = section.content.tobytes()
+            modinfo = [entry.decode() for entry in modinfo.split(b'\x00') if entry]
```
maringuu, line 216: Should we handle decoding errors here?

src/helperFunctions/hash.py
```diff
@@ -32,7 +32,7 @@ def get_tlsh_comparison(first, second):
     return tlsh.diff(first, second)
 
 
-def normalize_lief_items(functions):
+def normalize_lief_items(functions) -> list[str]:
```
maringuu, line 35: Couldn't we remove this function completely?
Why is this in `helperFunctions.hash` anyways?

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,233 @@
...
+    class Schema(BaseModel):
+        header: ElfHeader
+        sections: List[ElfSection]
+        segments: List[ElfSegment]
+        dynamic_entries: List[DynamicEntry]
+        exported_functions: List[ElfSymbol]
+        imported_functions: List[str]
+        mod_info: Optional[List[str]]
+        note_sections: List[InfoSectionData]
+        behavior_classes: List[str]
```
maringuu, line 101: Shouldn't we document these?
Imo, we should at least document the non-standart ones like `mod_info` and `behavior_classes`.

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,233 @@
...
+        elf = lief.parse(file_handle.name)
+        json_dict = json.loads(lief.to_json(elf))
+        _convert_flags(json_dict)
+        return self.Schema(
+            header=ElfHeader.model_validate(json_dict['header']),
+            exported_functions=[ElfSymbol(name=f.name, offset=f.address) for f in elf.exported_functions],
+            imported_functions=[f.name for f in elf.imported_functions],
+            sections=[ElfSection.model_validate(s) for s in json_dict['sections']],
+            segments=[ElfSegment.model_validate(s) for s in json_dict['segments']],
+            dynamic_entries=[DynamicEntry.model_validate(e) for e in json_dict['dynamic_entries']],
```
maringuu, line 131: Imo the code gets more readable if we do not serialize the lief elf representation to json.
E.g. like the `exported_functions` above.

maringuu (changes requested):

src/plugins/analysis/elf_analysis/code/elf_analysis.py
jstucke: changed

src/helperFunctions/hash.py
jstucke: removed (or rather inlined)

src/plugins/analysis/elf_analysis/code/elf_analysis.py
jstucke: We will probably encounter only ascii but it won't hurt either

src/plugins/analysis/elf_analysis/code/elf_analysis.py
jstucke: documented both fields

jstucke:
I made some additional changes:
- libraries were missing
- fixed the `_get_elf_analysis_libraries()` method to work with the new result structure
- deduplicated imports and exports (for whatever reason, they can sometimes contain duplicates in lief)

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,288 @@
...
+                'application/x-sharedlib',
+            ],
+        )
+        super().__init__(metadata=metadata)
+
+    def analyze(self, file_handle: FileIO, virtual_file_path: str, analyses: dict) -> Schema:
+        del virtual_file_path, analyses
+        elf = lief.ELF.parse(file_handle.name)
+        if elf is None:
+            raise TypeError('not a valid ELF file')
```
maringuu, line 178: This should be a `ValueError` instead of a `TypeError` [1].
[1]: https://docs.python.org/3/library/exceptions.html#TypeError

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,288 @@
...
+
+def _get_note_sections_content(elf: lief.ELF) -> Iterable[InfoSectionData]:
+    for section in elf.sections:  # type: lief.ELF.Section
+        if section.type == lief.ELF.Section.TYPE.NOTE:
+            readable_content = bytes([c for c in section.content.tobytes() if c in PRINTABLE_BYTES])
+            yield InfoSectionData(name=section.name, contents=readable_content.decode())
+
+
+def _deduplicate_functions(lief_functions: Iterable[lief.Function]) -> list[tuple[int, str]]:
+    return sorted({(f.address, f.name) for f in lief_functions})
```
maringuu, line 288: Can you file a bug report upstream and put a link to it in the code?
Then when someone stumbles upon this code again, they can see if it is still needed.

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,288 @@
...
+class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
+    class Schema(BaseModel):
+        header: ElfHeader
+        sections: List[ElfSection]
+        segments: List[ElfSegment]
+        dynamic_entries: List[DynamicEntry]
+        exported_functions: List[ElfSymbol]
+        imported_functions: List[ElfSymbol]
+        libraries: List[str]
+        mod_info: Optional[List[str]] = Field(description='Key value pairs with module information.')
```
maringuu, line 155: Could you reword the description and make more clear that this is related to _Linux kernel_ modules (Isn't it?).
To me the description does not convey much information that is not conveyed by the type and name of the field. 

Also, how do 'key value pairs' fit in an `Optinal[List[str]]` ?

maringuu (changes requested):

src/plugins/analysis/elf_analysis/code/elf_analysis.py
jstucke: I'm actually not sure whether or not this is intended. In the old plugin this wasn't an issue, since it also "deduplicated" the results using sets, so in the plugin this is a new bug

src/plugins/analysis/elf_analysis/code/elf_analysis.py
jstucke: I agree, a list of key value pairs doesn't make a lot of sense. I converted it to a dict.

maringuu (dismissed):

src/plugins/analysis/elf_analysis/code/elf_analysis.py
```diff
@@ -1,161 +1,288 @@
...
+            machine_type=header.machine_type.__name__.lower(),
+            numberof_sections=header.numberof_sections,
+            numberof_segments=header.numberof_segments,
+            object_file_version=header.object_file_version.__name__,
+            processor_flag=header.processor_flag,
+            program_header_size=header.program_header_size,
+            program_headers_offset=header.program_header_offset,
+            section_header_size=header.section_header_size,
+            section_headers_offset=header.section_header_offset,
+            section_name_table_idx=header.section_name_table_idx,
```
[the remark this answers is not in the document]
jstucke: > Not a huge fan of using __name__ to convert the constants to human readable names.
The lief names are not documented somehwere accessible (are they?).

neither am I, but sadly this seems to be the only way to reproduce the old output

> I'm not sure what the right way to proceed is here. From a GUI user perspective human readable names are important. From a API user perspective the numeric values would be preferable.

since the barrier for a GUI user to work with numeric values is arguably higher, IMHO using strings is the better approach

> Also, why does this work at all? It does not work with standard enum.Enum's and is not defined in lief's typing stubs.

It probably has something to do with the fact that lief is not actually python code, but rather a shared library (.so) with python bindings

jstucke:
> Just my two cents on how multiple symbols can have the same name.
The ELF specification is defined as part of the System V ABI [[1](https://wiki.osdev.org/System_V_ABI#Documents)].
According to it, multiple non-global symbols can have the same name in multiple object files [[2](https://www.sco.com/developers/gabi/latest/ch4.symtab.html)].

I'm not sure this is the problem here. In cases of duplicate entries, every entry seems to have a duplicate (with the same offset). I'm still not sure why

euwint (approved):
Looks good & works flawlessly now. We should bear in mind that we should also update this HTML code soon to force line breaks. On narrow windows, the text currently extends into the next column within the table.

PR #1274, opened 2024-09-27
Bugfix: file_system_meta_data did not show all results for a file object across firmwares

rhelmke:
When an unpacked and analyzed file has the same name across firmware images, the `file_system_meta_data` plugin did not show the results for all firmware images in the file object view.


jstucke (changes requested):
Changes look good but it seems `test_file_system_metadata_routes.py` needs a fix

rhelmke:
Should be fixed now

jstucke (approved):
LGTM!

PR #1279, opened 2024-10-07
tree structure for comparison changed text files

jstucke:


euwint (changes requested):
I have a few change requests:

1. wider “show file diff” buttons so that the text fits in one line.
2. reduction of the highlighting so that it no longer looks so wild or restless when you hover over objects with the mouse
3. inserting folder and file type icons so that the look is closer to the FACT-typical file tree

Further note: We should also think about changing the appearance of the lower tables “files in common” and “non-zero files in common”! Perhaps we could also remove the “files in common” table.

euwint (approved):
Looks much better

PR #1280, opened 2024-10-11
Fs metadata yaffs

jstucke:
- added support for YAFFS filesystems to the filesystem metadata plugin

### DIFF
--- a/src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
+++ b/src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
@@ -49,7 +49,7 @@ FS_MIME_TYPES = [
     'filesystem/xfs',
     'filesystem/squashfs',
 ]
-YAFFS_REGEX = re.compile(r'([rwxtTsS?-]{10}) +\d+ (\d{4}-\d{2}-\d{2} \d{2}:\d{2}) ([^\n]+)')
+YAFFS_REGEX = re.compile(r'([rwxtTsSl?-]{10}) +\d+ (\d{4}-\d{2}-\d{2} \d{2}:\d{2}) ([^\n]+)')
 REVERSE_FILEMODE_LOOKUP = [{char: mode for mode, char in row} for row in stat._filemode_table]
 
 
@@ -162,11 +162,14 @@ class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
         -rw-r--r--       38 2014-07-19 01:31 build.prop
         drwxr-xr-x        0 2016-10-11 03:12 sbin
         lrwxrwxrwx        0 2014-07-19 01:28 sbin/ifconfig -> ../bin/busybox
-        we ignore directories and symlinks
+        we ignore directories and block devices
         """
         for match in YAFFS_REGEX.finditer(unpacker_result['output']):
             mode_str, date, path = match.groups()
             mode = oct(_filemode_str_to_int(mode_str))
+            if '->' in path:
+                # symlink entries have a " -> [target]" after their path (see comment above)
+                path = path.split('->')[0].strip()
             result.append(
                 FileMetadata(
                     mode=mode,
--- a/src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
+++ b/src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
@@ -134,13 +134,15 @@ class AnalysisPlugin(AnalysisPluginV0, AnalysisBasePluginAdapterMixin):
         )
         super().__init__(metadata=metadata)
 
-    def analyze(self, file_handle: FileIO, virtual_file_path: dict, analyses: dict[str, BaseModel]) -> Schema:
+    def analyze(self, file_handle: FileIO, virtual_file_path: dict, analyses: dict[str, BaseModel | dict]) -> Schema:
         del virtual_file_path
         file_type = analyses['file_type'].mime
         result = self._extract_metadata(file_handle, file_type, analyses)
         return self.Schema(files=result)
 
-    def _extract_metadata(self, file_handle: FileIO, file_type: str, analyses) -> list[FileMetadata]:
+    def _extract_metadata(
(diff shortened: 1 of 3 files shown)


src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -129,16 +137,48 @@ def __init__(self):
...
+        unpacker_result = analyses.get('unpacker')
+        if not unpacker_result or unpacker_result['plugin_used'] != 'YAFFS':
+            return result
+        """
+        the output of unyaffs from the unpacker log has the following structure:
+        brw-r-----  31,   4 2014-07-19 01:28 dev/mtdblock4
+        -rw-r--r--       38 2014-07-19 01:31 build.prop
+        drwxr-xr-x        0 2016-10-11 03:12 sbin
+        lrwxrwxrwx        0 2014-07-19 01:28 sbin/ifconfig -> ../bin/busybox
+        we ignore directories and symlinks
```
rhelmke, line 163: I'm not entirely conviced that it is a wise idea to ignore symlinks. It is true that most unix systems ignore symlink permissions, since they are so transparent that ownership and read changes affect the linked file instead.

_Yet_, which permissions are checked when we _create_ or _overwrite_ a symlink? Those permissions are inherited from the parent directory. Meaning that once users have write access to the parent directory, they can overwrite the symlink. This is, indeed, a security-relevant path and it is a pity that this information gets lost here.

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -270,3 +310,10 @@ def _get_tar_file_mode_str(file_info: tarfile.TarInfo) -> str:
 
 def _has_correct_type(mime_type: str) -> bool:
     return mime_type in ARCHIVE_MIME_TYPES + FS_MIME_TYPES
+
+
+def _filemode_str_to_int(filemode: str) -> int:
```
rhelmke, line 315: This function has potential to be of common usability. Consider relocating it to its inverse counterpart, which is already a common function.

rhelmke (changes requested):

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
jstucke: Since directories have no real representation in FACT (analogous to files which are represented as FileObjects), I don't really see a way to fix this. 

We could simply add them to the result of the archive/FS (where they are included), though. We could even check the permissions for inconsistencies. IMHO that would be out of scope for this PR, though.

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
jstucke: Putting the function in the same place (`filter.py`) as `octal_to_readable()` is problematic, because that function is located in the frontend code and we should not import from there from a plugin (backend)

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -129,16 +137,48 @@ def __init__(self):
...
+    def _extract_metadata(self, file_handle: FileIO, file_type: str, analyses) -> list[FileMetadata]:
         if file_type in FS_MIME_TYPES:
             return self._extract_metadata_from_file_system(file_handle)
         if file_type in ARCHIVE_MIME_TYPES:
             return _extract_metadata_from_tar(file_handle)
+        if file_type == 'filesystem/yaffs':
+            return self._extract_metadata_from_yaffs(analyses)
         return []
 
+    def _extract_metadata_from_yaffs(self, analyses) -> list[FileMetadata]:
```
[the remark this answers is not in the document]
jstucke: fixed

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -120,7 +128,7 @@ def __init__(self):
             description=(
                 'extract file system metadata (e.g. owner, group, etc.) from file system images contained in firmware'
             ),
-            version='1.0.0',
+            version='1.1.0',
```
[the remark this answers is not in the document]
rhelmke: discussed offline

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
rhelmke: I understand. Then leave it as is - for the time being.

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
```diff
@@ -270,3 +315,10 @@ def _get_tar_file_mode_str(file_info: tarfile.TarInfo) -> str:
 
 def _has_correct_type(mime_type: str) -> bool:
     return mime_type in ARCHIVE_MIME_TYPES + FS_MIME_TYPES
+
+
+def _filemode_str_to_int(filemode: str) -> int:
```
rhelmke, line 320: I believe this function is buggy. It returns an invalid octal value for, e.g., `-rwsr-xr-x`

```python
In [6]: oct(filemode_str_to_int("-rwsr-xr-x"))
Out[6]: '0o104755'
```

It should be `0o4755`.

rhelmke: Unix permissions are limited to four octal digits.

These bad permissions are persisted in the plugin results.

rhelmke (changes requested):

src/plugins/analysis/file_system_metadata/code/file_system_metadata.py
rhelmke: Okay. So for reference: the `filemode` is correct! Because the [leftmost two octals describe the file type](https://stackoverflow.com/a/4087441).
However, this value describes more than a `permission`. We decided to create a new PR that introduces a new analysis result field that separates `permissions` from `file types`.

rhelmke (approved):

PR #1282, opened 2024-10-15
show link to symlink target on analysis page

jstucke:


src/storage/db_interface_frontend.py
```diff
@@ -453,3 +456,33 @@ def _get_mode_dict(self, parent_uid: str | None) -> dict[str, str]:
...
+                    or_(
+                        VirtualFilePath.parent_uid == fw_files_table.c.file_uid,
+                        VirtualFilePath.parent_uid == root_uid,  # special case: parent is also root
+                    )
+                )
+                .filter(VirtualFilePath.parent_uid.in_(parents))
+                .filter(VirtualFilePath.file_path.in_(candidate_paths))
+            )
+            for uid in session.execute(query.limit(1)).scalars():
+                return uid
```
maringuu, line 487: I am aware that we, by design, can not get this right, but I'd just like to note that using `limit(1)` to get a random target file is not necessarily correct.

src/web_interface/components/analysis_routes.py
```diff
@@ -232,6 +233,25 @@ def show_elf_dependency_graph(self, uid: str, root_uid: str):
...
+
+    def _get_link_target(self, file_obj: FileObject, root_uid: str) -> str | None:
+        if not root_uid:
+            return None
+        try:
+            full_type = file_obj.processed_analysis['file_type']['result']['full']
+            # if FO is a symlink, file_type analysis "full" will be something like "symbolic link to 'busybox'"
+            target_path = full_type.split("'")[1]
+        except (IndexError, KeyError):
+            return None
```
maringuu, line 251: Again, we can't really do much about this, but this prone to be broken.
For example, it does not work with the link created by `ln -s "./crash'es" "foo'bar"`.
Furthermore, I'd like to note that this is actually not standard `file` output,
but defined in `src/install/internal_symlink_magic`.

src/web_interface/templates/show_analysis.html
```diff
@@ -112,7 +112,11 @@
         {# Header section #}
         <div class="header mb-4" style="word-wrap: break-word">
             <h3>
-                {{ firmware.uid | replace_uid_with_hid(root_uid=root_uid) | safe }}<br />
+                {{ firmware.uid | replace_uid_with_hid(root_uid=root_uid) | safe }}
+                {% if link_target %}
+                    ({{ link_target | safe }})
```
maringuu, line 117: This is not `safe`:

```sh
printf "\n" > '<script>alert("injection");'
ln -s \<script\>alert\(\"injection\"\)\; foobar

tar -cvf x.tar ./*
```
While this does not run `altert`, it still breaks the whole site.

maringuu (changes requested):
Overall I am hesitating to merge this as FACT's internals are not really built with support for symlinks in mind as can be seen by the above comments.
Of course, this feature is useful and desirable, but not possible to implement correctly atm.
I'll leave it to you or @dorpvom to merge this.

src/storage/db_interface_frontend.py
jstucke: Normally, there should only be one result at this point (but I there may be so weird edge cases that I haven't thought of)

src/web_interface/templates/show_analysis.html
jstucke: Added a `html.escape()`

src/web_interface/components/analysis_routes.py
jstucke: > prone to be broken

This is a bit of an overstatement. I have yet so to come across a file with a single quote in its file name in a firmware image. And even if that happened, we would simply not find the linked file and nothing more would break. Nevertheless, I changed the code so that it won't break even in this case.

src/storage/db_interface_frontend.py
jstucke: To elaborate a bit on the underlying idea: at this point we know the root FW of the file and the path that is mentioned in the file type ("link to '/a/b/c'). The idea is to look in the "virtual path" for paths that line up with this. Usually, this should be only one file. But there are two other possibilities: the file is missing from the file system or there are multiple file systems or containers with this path contained in the firmware. In the latter case, there isn't really a right or wrong result, since the file object represents all of these files at once. We could display *all* link targets, though.

maringuu (approved):

jstucke:
rebased on master to fix the bug that let the CI fail

PR #1283, opened 2024-10-17
fix flaky helper process tests

jstucke:
- fix flaky helper process tests and also speed them up a bit with events

### DIFF
--- a/src/test/unit/helperFunctions/test_process.py
+++ b/src/test/unit/helperFunctions/test_process.py
@@ -1,14 +1,18 @@
 import logging
+from multiprocessing import Event
 from time import sleep
 
 import pytest
+from flaky import flaky
 
 from helperFunctions.process import ExceptionSafeProcess, check_worker_exceptions, new_worker_was_started
+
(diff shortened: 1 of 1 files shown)


src/test/unit/helperFunctions/test_process.py
```diff
@@ -1,14 +1,18 @@
...
 
 from helperFunctions.process import ExceptionSafeProcess, check_worker_exceptions, new_worker_was_started
 
+check_exceptions_finished = Event()
+
 
 def breaking_process(wait: bool = False):
     if wait:
-        sleep(0.5)
+        check_exceptions_finished.wait(timeout=5)
```
euwint, line 15: Maybe we could use adjustable timeouts through environment variables?

euwint (approved):

src/test/unit/helperFunctions/test_process.py
jstucke: Good idea but it is probably best to do this for all sleep/wait times in the tests at once (and that is probably out of scope for this PR)

PR #1284, opened 2024-10-23
copy raw result button

jstucke:


src/web_interface/static/js/show_analysis_preview.js
```diff
@@ -70,3 +70,9 @@ window.onload = function() {
     // make sure the switch is off when the page is reloaded
     toggleSwitch.checked = false;
 };
+
+function copyRawAnalysis() {
+    const code = document.getElementById("raw-analysis");
+    // Copy the raw analysis data to clipboard
+    navigator.clipboard.writeText(code.innerText);
```
rhelmke, line 77: The button is defunct when the raw analysis is not shown. It's probably better to fetch the result from the endpoint.

rhelmke (changes requested):

rhelmke (approved):
thanks!

PR #1285, opened 2024-10-23
fs metadata: show mode incl. type

jstucke:
- also show file type if the file mode data includes this
- change the label in the result from "permissions" to "file mode"

### DIFF
--- a/src/plugins/analysis/file_system_metadata/view/file_system_metadata.html
+++ b/src/plugins/analysis/file_system_metadata/view/file_system_metadata.html
@@ -12,7 +12,7 @@
 {% endif %}
 {% for metadata in metadata_list %}
     <tr>
-        <td>{{ metadata['path'] }}</td>
+        <td style="overflow-wrap: anywhere;">{{ metadata['path'] }}</td>
         <td class="p-0">
             {# this should be the same as in ../routes/ajax_view.html #}
             <table class="table table-bordered table-sm mb-0">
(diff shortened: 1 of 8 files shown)


src/plugins/analysis/file_system_metadata/view/file_system_metadata.html
```diff
@@ -28,9 +28,9 @@
                     <td class="text-left" style="width: 25%">{{ metadata.creation_time | nice_unix_time | safe }}</td>
                 </tr>
                 <tr>
-                    <td class="text-right">permissions</td>
+                    <td class="text-right">file mode</td>
```
rhelmke, line 31: in most instances i've seen, `filemode` is written as one word.
```suggestion
                    <td class="text-right">filemode</td>
```

src/web_interface/components/jinja_filter.py
```diff
@@ -215,11 +215,11 @@ def _setup_filters(self):  # noqa: PLR0915
         self._app.jinja_env.filters['nice_unix_time'] = flt.nice_unix_time
         self._app.jinja_env.filters['nice_virtual_path_list'] = self._nice_virtual_path_list
         self._app.jinja_env.filters['number_format'] = flt.byte_number_filter
```
rhelmke, line 217: `octal_to_readable` is still a valuable filter. Do not remove it.

```suggestion
        self._app.jinja_env.filters['number_format'] = flt.byte_number_filter
        self._app.jinja_env.filters['octal_to_readable'] = flt.octal_to_readable
```

rhelmke (changes requested):

rhelmke (approved):
Ok, looks good! Thanks!

PR #1287, opened 2024-10-25
sped up highlighting on analysis page

jstucke:
- sped up highlighting on analysis page
  - optimized code
  - replaced custom line numbering function with much faster library
  - don't highlight if the block is too large (>1MiB)


src/web_interface/static/js/show_analysis_preview.js
```diff
@@ -43,16 +57,16 @@ function load_preview(offset = null, focus = false) {
     preview_loading_gif.style.display = "block";
     $("#preview-content").load(resourcePath, init_preview);
 }
+
 preview_button.onclick = load_preview;
 let rawResultIsHighlighted = false;
 const toggleSwitch = document.getElementById("rawResultSwitch");
+const analysisTable = document.getElementById("analysis-table-body");
```
maringuu, line 64: This is `null` if no analysis is selected (by clicking on the file name).

maringuu (approved):
Nice! Just one commend below.

maringuu (approved):

PR #1288, opened 2024-10-29
converted IP&URI plugin to new base class

jstucke:


src/plugins/analysis/ip_and_uri_finder/code/ip_and_uri_finder.py
```diff
@@ -1,105 +1,110 @@
...
-        'application/x-dosexec',
-    ]
-    DESCRIPTION = 'Search file for IP addresses and URIs based on regular expressions.'
-    VERSION = '0.4.2'
-    FILE = __file__
-
-    def additional_setup(self):
+class IpAddress(BaseModel):
+    address: str
+    location: Optional[str]
```
maringuu, line 30: It would make more sense to have this be two fields (longitude and latitude) insead of a comma separated string, wouldn't it?

maringuu (changes requested):
One thing, otherwise lgtm.

src/plugins/analysis/ip_and_uri_finder/code/ip_and_uri_finder.py
jstucke: I changed it

maringuu (approved):

PR #1301, opened 2024-11-18
more software signatures

jstucke:
- added some software signatures (mbed TLS, file, opkg)


jstucke:
> Can you document somewhere (commit message/code comments), why these signatures are correct? I think it is important to not only have signatures, but also be able to verify why they are correct. The most preferable proof would we permalinks to source code, but I'm happy with anything that allows me to understand why the signatures are as they are.

I tried to find the according places in the source code but I also found that in libmagic the version is stored as an integer instead of a string. The FSR currently does not have the capability to extract integers, so I added it.

src/plugins/analysis/software_components/code/software_components.py
```diff
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import json
 import re
 import string
 from typing import TYPE_CHECKING
 
+from sqlalchemy.testing.plugin.plugin_base import logging
```
maringuu, line 8: Wrong auto import?

maringuu:
Does this need a rebase after #1263?

src/plugins/analysis/software_components/code/software_components.py
jstucke: Yes, exactly. Good catch!

jstucke:
> Does this need a rebase after #1263?

I don't think so, since the PR didn't change any internals, but I did one anyway.

src/plugins/analysis/ipc/docker/ipc_analyzer/ipc_analyzer.py
```diff
@@ -336,6 +336,38 @@ def find_function_ref_strings(function_name):
...
+        for instruction in instruction_iterator:
+            for i in range(instruction.getNumOperands()):
+                for operand in instruction.getOpObjects(i):
+                    try:
+                        value = operand.getValue()
+                    except AttributeError:
+                        continue
+                    if value is not None and isinstance(value, (int, long)):
+                        constants.append(str(value))
+    return constants
```
maringuu, line 368: Nit: I personally prefer less indentation, which in this case can be archived by adding:
```python
if function is None:
    return []
```

Also: Why can function be none if we did not get an exception in the call to `getGlobalFunctions`?!
I think it [cannot](https://ghidra.re/ghidra_docs/api/ghidra/program/flatapi/FlatProgramAPI.html#getGlobalFunctions(java.lang.String)).

src/plugins/analysis/ipc/docker/ipc_analyzer/ipc_analyzer.py
```diff
@@ -336,6 +336,38 @@ def find_function_ref_strings(function_name):
...
+    Get all constants that are used as operands in the function with name `function_name`.
+
+    :param function_name: The name of the function.
+    :type function_name: str
+    :return: a list of int/long constants referenced in the function as strings
+    :rtype: list[str]
+    """
+    try:
+        function = getGlobalFunctions(function_name)[0]
+    except (IndexError, TypeError):
```
maringuu, line 350: Why does this raise a type error?

src/plugins/analysis/software_components/code/software_components.py
```diff
@@ -50,9 +52,26 @@ def get_version(self, input_string: str, meta_dict: dict) -> str:
...
+        """
+        The metadata entry "_sub_regex" can be used to change the version string if it does not have the expected
+        format (e.g. add dots). The entry should contain a regex and replacement for `re.sub()` as JSON string
+        """
+        try:
+            sub_regex, replacement = json.loads(meta_dict['_sub_regex'])
+            return re.sub(sub_regex, replacement, version_str)
+        except json.JSONDecodeError:
+            logging.warning(f'[{self.NAME}]: signature has invalid substitution regex: {meta_dict}')
+            return ''
```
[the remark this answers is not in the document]
maringuu: Also, what is the difference between the version_regex and the _sub_regex?
Could (should?) they be merged?

maringuu (changes requested):
A few comments otherwise lgtm!

src/plugins/analysis/ipc/docker/ipc_analyzer/ipc_analyzer.py
jstucke: > Also: Why can function be none if we did not get an exception in the call to `getGlobalFunctions`?!
> I think it [cannot](https://ghidra.re/ghidra_docs/api/ghidra/program/flatapi/FlatProgramAPI.html#getGlobalFunctions(java.lang.String)).

I think you are right -> changed

src/plugins/analysis/ipc/docker/ipc_analyzer/ipc_analyzer.py
jstucke: I don't think this raises a type error if `function_name` is a string -> changed

PR #1307, opened 2024-12-03
Updated .gitignore

dorpvom:
On new installations paths to the local cve database and node modules / nodeenv has changed


jstucke (approved):
LGTM! :star2:

PR #1312, opened 2024-12-09
added coverage multiprocessing config for CI master builds

jstucke:


maringuu:
Why not enable it in general?
Using `sed` to change the pyproject.toml does not seem like a good idea to me.

jstucke:
> Why not enable it in general? Using `sed` to change the pyproject.toml does not seem like a good idea to me.

Good question! This is more or less a hack. The simple reason is, that with the option `multiprocessing` enabled, CI builds take 6--8 minutes longer which is a bit too much. For the nightly CI build this is not an issue, though.

jstucke:
> One could e.g., change the CI script to run pytest as follows:
> 
> ```shell
> pytest \
>     --cov=. \
>     --cov-config=.github/.coveragerc
> ```
> 
> The .coveragerc would then contain any configuration that diverges from the pyproject.toml.

Interesting proposal. I'm don't like the idea of using `sed` either (because it can easily break when the config file changes. I think the return value should indicate when it failed, though?). This on the other hand has the disadvantage that you need to maintain two config files henceforth, though.

PR #1313, opened 2024-12-10
additional dark mode fixes

jstucke:
- dependency graph
  - fix node text color in dependency graph for dark mode
  - some jshint fixes
  - moved the dark mode JavaScript code into its own file (so it can also be used in `dependency_graph.js`)
- GraphQL search
  - fixed `<pre>` elements for dark mode


maringuu:
The `Browse` button on the upload also does not change color as opposed to the other buttons.

jstucke:
> The `Browse` button on the upload also does not change color as opposed to the other buttons.

Yes, I know. But sadly I was unable to fix this, as this is some internal Bootstrap black magic thing that does not seem to be addressable with CSS. Feel free to try.

maringuu (approved):

PR #1317, opened 2024-12-12
System version more plugins

jstucke:
- added system_version for more plugins
  - file_type: used version of [FW magic DB](https://github.com/fkie-cad/firmware-magic-database)
  - cve_lookup: used version of [NVD JSON data feeds](https://github.com/fkie-cad/nvd-json-data-feeds)

### DIFF
--- a/src/plugins/analysis/file_type/code/file_type.py
+++ b/src/plugins/analysis/file_type/code/file_type.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+import json
 import typing
-from typing import List
+from pathlib import Path
 
 import pydantic
 from pydantic import Field
@@ -9,6 +10,7 @@ from semver import Version
 
 from analysis.plugin import AnalysisPluginV0
 from helperFunctions import magic
+from helperFunctions.fileSystem import get_bin_dir
 
 if typing.TYPE_CHECKING:
     import io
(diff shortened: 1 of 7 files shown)


maringuu:
What problem does this solve?
I do not quite understand how the _system version_ is useful.

jstucke:
> What problem does this solve? I do not quite understand how the _system version_ is useful.

It is useful for two things:
- track the internal version of the data that the analysis is based on (since the outcome of the analysis may change based on this data)
- if you run the analysis again with the same version of the plugin, it depends on the system version whether or not the analysis is run again or skipped (and since the outcome may change, it makes sense to update it)

maringuu:
To me the system version still seems like a design flaw.
Imo it would make more sense to fix the version of whatever tool a plugin needs and bump the plugin version if we update the tool.

Let's consider the `file_type` plugin for example.
If I'm on ubuntu and happen to update my system (via `apt upgrade`) and the `file` package is updated, this results in a new system version for the `file_type` plugin.
When I then upload a new firmware image to FACT, the results are different than what I would have gotten when I uploaded the same firmware before upgrading `file`.
To me this does not make sense. The analysis results should not change, if I didn't touch the fact install.

jstucke:
I tend to agree with you. That is one of the reasons why we don't install some tools through apt: To get reproducible results. In some cases it is hard to address this, e.g. in case of the cve_lookup plugin it wouldn't make a lot of sense to pin the commit of the CVE sources (because you would need to update the plugin all the time to get the latest entries). But it adds some meta information which makes it easier to track why we see certain results. Maybe it would make sense to remove the "system version" from scheduling. 

And I think that we should try to eliminate as many sources of indeterminism/uncertainty as possible (like the version of file/libmagic). In the case of file/libmagic I'm not sure what the best way to do that would look like, though: compiling it adds unnecessary installation overhead and running it inside a docker container adds runtime overhead during analysis (which is even worse). Maybe downloading a Debian/Ubuntu package with the pre-compiled magic file and extracting it could work :see_no_evil:

maringuu:
I opened #1349 to track the discussion.

src/plugins/analysis/file_type/code/file_type.py
```diff
@@ -24,11 +28,17 @@ class Schema(pydantic.BaseModel):
...
+            fw_magic_db_version = json.loads(version_file.read_text()).get('version')
+        except (json.JSONDecodeError, FileNotFoundError):
+            fw_magic_db_version = None
         super().__init__(
             metadata=AnalysisPluginV0.MetaData(
                 name='file_type',
                 description='identify the file type',
-                version='1.0.0',
+                version=Version(1, 0, 1),
+                system_version=fw_magic_db_version,
```
maringuu, line 41: Isn't this technically wrong?
While the plugin defensively depends on the version of the  8firmware-magic-database ](https://github.com/fkie-cad/firmware-magic-database), it also depends on the version of file.

maringuu (changes requested):
One comment above, otherwise lgtm.

PR #1318, opened 2024-12-13
Update browse comparisons view

dorpvom:
A real update might come later, but for now just using the slightly nicer listing from the home / dashboard page to be more consistent with general fact look.


jstucke (approved):
Looks good to me, but maybe I'm biased :sweat_smile:

euwint (approved):
I agree, it looks good.

jstucke:
the noble CI worker only failed because of a bug with the package sources (which was fixed in the meantime). Shouldn't prevent us from merging I think

PR #1320, opened 2024-12-16
forward to analysis page (missing analysis)

jstucke:
- if an analysis did not run for a file, the user is directed to an error page (a page empty except for an error message)
- this is not helpful, so this PR changes this behavior as follows: Instead of showing the error page the user is forwarded back to the analysis page (without selected plugin) and the error is displayed as flash message
- steps to (re)produce the error: Go to some file in a FW, run a "single file analysis" and go back to the root object through the link in the "general information" section


maringuu:
Please put the PR description in the commit message.
If you do this before opening the commit, GitHub will automatically use the commit message as PR description.

maringuu (approved):
Otherwise lgtm.

maringuu:
Thanks!

PR #1321, opened 2024-12-16
add db view unpacking

jstucke:


maringuu:
What is this used for?

jstucke:
> What is this used for?

I should probably have included a description :sweat_smile: 
I only talked with @dorpvom about this. The problem this PR addresses is the following: The metadata from unpacking (entropy, extracted file count and other stuff that lands in `fo.processed_analysis["unpacking"]`) cannot be used together with numerical comparison operators (like "greather than" or "less equal than"). The reason for this is simple: GraphQL does not support this on JSON data. This is the simplest workaround for using this data without creating more tables/etc.: Creating a "view" for this data, which is a PostgreSQL construct, that allows using the data that is stored in a JSONB column as if it was stored as a number in a regular table in queries (and also GraphQL). 

Note: in PostgreSQL we can absolutely formulate such a query, only GraphQL (or more specifically the default schema created by Hasura) is the problem

What can we do with this? We can use this to e.g. run queries like "find all files with an entropy greater or equal 0.8 from which no files were unpacked".

maringuu:
Shouldn't there be an alembic migration for this?

Asides, I am happy with this if you put your description in the commit message.

jstucke:
> Shouldn't there be an alembic migration for this?

That's a good question. Does this need a migration if it is not part of the SQLAlchemy schema? Making it part of the schema is not really necessary, since the query limitations of GraphQL on JSON data do not apply to SQLAlchemy.

PR #1323, opened 2025-01-10
feat: converted strings plugin to v1 and integrated strings eval

jstucke:
* migrates the "printable_strings" plugin to the new base class
* integrates the "string_evaluation" plugin into the "printable_strings" plugin
* uses bootstrap-table to display the new result structure in the template


src/plugins/analysis/strings/internal/string_eval.py
```diff
@@ -0,0 +1,132 @@
...
+    r'@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))'
+)
+PREFIX_REGEX = re.compile(r'(^_+)|^\.')
+PARAMETER_REGEX = re.compile(r'^\s*-{1,2}')
+HTML_REGEX = re.compile(r'</?[^\\()$\[\]§.,?<>;|!]+>')
+
+DICTIONARY = {'version', 'v.', 'http', 'ftp', 'usage', 'Usage', 'ssh', 'SSH', 'password', 'Version'}
+
+
+def evaluate_string(string: str) -> float:
```
maringuu, line 33: nit: What about renaming this to `string_relevance`?

src/plugins/analysis/strings/code/strings.py
```diff
@@ -2,54 +2,75 @@
...
+                    version=Version(1, 0, 0),
+                    mime_blacklist=MIME_BLACKLIST_COMPRESSED,
+                    Schema=self.Schema,
+                )
+            )
+        )
         self.regexes = self._compile_regexes()
 
     def _compile_regexes(self) -> list[tuple[Pattern[bytes], str]]:
         min_length = getattr(config.backend.plugin.get(self.NAME, {}), 'min-length', 8)
```
maringuu, line 50: This only works as long as we use `AnalysisBasePluginAdapterMixin`. 
Instead this should use `self.metadata.name`.

src/plugins/analysis/strings/code/strings.py
```diff
@@ -2,54 +2,75 @@
...
+    class Schema(BaseModel):
+        strings: List[StringMatch] = Field(description='An array of ASCII strings contained in this file.')
 
-    def additional_setup(self):
+    def __init__(self):
+        super().__init__(
+            metadata=(
+                self.MetaData(
+                    name='printable_strings',
+                    description='extracts strings and their offsets from the files consisting of printable characters',
```
maringuu, line 40: What do you think?
```suggestion
                    description='Extracts printable strings from a file and assigns a relevance score based on a predefined ruleset.',
```

src/plugins/analysis/strings/internal/string_eval.py
```diff
@@ -0,0 +1,132 @@
+"""
+evaluate relevance of strings
+
+Credits:
+Original version by Paul Schiffer created during Firmware Bootcamp WT16/17 at University of Bonn
+Refactored and improved by Fraunhofer FKIE
+"""
```
maringuu, line 7: I don't think that doc comment is that useful.

maringuu (changes requested):
Overall lgtm (The table is especially nice!). Just a few nits.

src/plugins/analysis/strings/code/strings.py
jstucke: Good catch. Totally overlooked that bit

src/plugins/analysis/strings/internal/string_eval.py
jstucke: `string_relevance` is lacking a verb. I renamed it to `calculate_relevance_score`

src/plugins/analysis/strings/code/strings.py
jstucke: I think your change better reflects what the plugin does -> changed

src/plugins/analysis/strings/internal/string_eval.py
jstucke: Removed the doc part and added a more informative docstring to the function instead.

maringuu (approved):
Needs a rebase, otherwise lgtm.

PR #1324, opened 2025-01-14
Checksec v1

jstucke:
- [converted exploit mitigations plugin to new base class](https://github.com/fkie-cad/FACT_core/commit/137fe93f13aeb57604fdc2b151b08bf45bded539)
- [renamed directory of exploit mitigations plugin to fit plugin name](https://github.com/fkie-cad/FACT_core/commit/ef767b6bdf118467da2aaad20fe0958af63513c1) ("checksec" -> "exploit_mitigations")


src/plugins/analysis/exploit_mitigations/code/exploit_mitigations.py
```diff
@@ -1,174 +1,125 @@
...
-        mitigations.update({'PIE': 'invalid ELF file'})
-
+        return 'enabled'
+    if checksec_result['pie'] == 'no':
+        return 'disabled'
+    if checksec_result['pie'] == 'dso':
+        return 'DSO'
+    if checksec_result['pie'] == 'rel':
+        return 'REL'
+    return 'invalid ELF file'
```
maringuu, line 117: On which files does this function return `'invalid ELF file'`? If this final return statement did not exist, we could ditch the function and replace it with a dictionary instead.

src/plugins/analysis/exploit_mitigations/view/exploit_mitigations.html
```diff
@@ -2,15 +2,27 @@
...
+            <td>{{ key }}</td>
+            <td class="table-{{ bs_class }}">
+                {% if value == True %}
+                    enabled
+                {% elif value == False %}
+                    disabled
+                {% else %}
+                    {{ value }}
+                {% endif %}
+            </td>
```
maringuu, line 24: I don't see the benefit of hiding the non-boolean information from relro and pie here. Couldn't we just add this?

maringuu (commented):
Just two questions, otherwise lgtm

src/plugins/analysis/exploit_mitigations/code/exploit_mitigations.py
jstucke: Good question. Actually I have no idea, since I did not originally write it. I also couldn't reproduce it (checksec will usually fail to produce any results if the file is not a valid ELF file)

src/plugins/analysis/exploit_mitigations/view/exploit_mitigations.html
jstucke: I'm not sure I'm following. The output should be "enabled" if it `True`, "disabled" if it is `False` or the value (a string in this case) otherwise. What are we hiding?

src/plugins/analysis/exploit_mitigations/code/exploit_mitigations.py
jstucke: I checked all values that are in my local DB and this is what I got:
`{'DSO': 2472, 'REL': 3874, 'disabled': 2478, 'enabled': 370, None: 494}`
Maybe it is safe to assume that this cannot happen and we can throw it out?

src/plugins/analysis/exploit_mitigations/view/exploit_mitigations.html
maringuu: Nevermind, I missed the `else`.

maringuu (approved):

PR #1326, opened 2025-01-14
feat: integrated interesting URIs plugin into IP&URI finder plugin

jstucke:


maringuu (approved):
Lgtm. Not merging because I'm not sure if this needs @dorpvom's approval.

PR #1327, opened 2025-01-15
feat: removed tlsh plugin

jstucke:
The plugin does not make a lot of sense in its current form, since the results are only based on the current contents of the DB


maringuu:
Nice!

maringuu (approved):

PR #1328, opened 2025-01-16
chore(plugins): Do not depend on `AnalysisBasePluginAdapterMixin`

maringuu:
The mixin is an internal thing that should not be used in plugins. Moving it in the abstract class, allows us to silently remove it in the future.

@jstucke Iirc, we discussed this some time ago but apparently forgot to ever to this :)


jstucke (approved):
LGTM!

PR #1330, opened 2025-01-20
Yara python fix

jstucke:
* [install yara-python from source](https://github.com/fkie-cad/FACT_core/commit/ecafc7c6d54d7d6840a9dfefd2f9425d7c96680f)
  * Why ist this necessary? Plugins like software_components need support for YARA modules (especially magic) and the pre-built version of yara-python from PyPI does not include this
* [moved get_yara_error method to backend (intercom)](https://github.com/fkie-cad/FACT_core/commit/700adcad062220f8cb9360dbdde237082e31c6e9)
  * having the method in the frontend does not make a lot of sense, because the yara version of the backend could be different and compiling the rules could still fail

### DIFF
--- a/src/helperFunctions/yara_binary_search.py
+++ b/src/helperFunctions/yara_binary_search.py
@@ -130,29 +130,3 @@ class YaraBinarySearchScanner:
         compiled_rules = yara.compile(source=yara_rules.decode())
         compiled_rules.save(file=temp_rule_file)
         temp_rule_file.flush()
-
-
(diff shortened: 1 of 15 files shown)


dorpvom (approved):

src/helperFunctions/yara_binary_search.py
Fukulngx1, line 1: https://github.com/fkie-cad/FACT_core.wiki.git

PR #1335, opened 2025-01-27
chore: updated pip dependencies

jstucke:
* replaced flaky with pytest-rerunfailures (flaky does not work with newer versions of pytest)

* removed redundant dependencies

### DIFF
--- a/src/install/requirements_common.txt
+++ b/src/install/requirements_common.txt
@@ -2,29 +2,21 @@
 testresources==2.0.1
 
 # General python dependencies
-appdirs==1.4.4
 docker~=7.1.0
-flaky==3.7.0
-lief==0.15.0
-psutil==5.9.4
-psycopg2-binary==2.9.9
-pylint==2.15.5
-pytest-cov==4.0.0
-pytest-timeout==2.1.0
-pytest==7.2.0
-python-magic==0.4.27
+pytest-rerunfailures~=14.0
+lief~=0.16.2
+psutil~=6.1.1
+psycopg2-binary~=2.9.10
+pylint~=3.2.7
+pytest-cov~=5.0.0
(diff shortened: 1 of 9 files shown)


src/install/requirements_common.txt
```diff
@@ -2,34 +2,26 @@
 testresources==2.0.1
 
 # General python dependencies
-appdirs==1.4.4
```
maringuu, line 5: There is a corresponding mock in `docsrc/conf.py` that needs to be removed.
Same for `flaky`. I'm not sure if there are others since it is hard to see which dependencies were removed.

maringuu (commented):
Iirc the versions were pinned using `==` intentionally to have a reproducible environment for the CI. On the other hand using `~=` should work as well.

I personally use [poetry](https://python-poetry.org/) to define my dependencies using compatibility specifiers in `pyproject.toml` and lock the dependencies using hashes in a `poetry.lock` lockfile.
I'd suggest to replace our various requirements.txt with several poetry groups.
What do you think?

jstucke:
> Iirc the versions were pinned using `==` intentionally to have a reproducible environment for the CI. On the other hand using `~=` should work as well.
> 
> I personally use [poetry](https://python-poetry.org/) to define my dependencies using compatibility specifiers in `pyproject.toml` and lock the dependencies using hashes in a `poetry.lock` lockfile. I'd suggest to replace our various requirements.txt with several poetry groups. What do you think?

Interesting idea. A bit out of scope for this PR but we should definitely discuss this further. This could also fix the problems that come from a missing venv.

maringuu:
> Interesting idea. A bit out of scope for this PR but we should definitely discuss this further. 

I opened https://github.com/fkie-cad/FACT_core/issues/1343 to track this

PR #1337, opened 2025-01-28
feat: use CodeMirror for displaying raw result

jstucke:


dorpvom:
I was curious what PR #1337 would be out of principal.. Now. I don't get it. What does this do @jstucke?

jstucke:
> I was curious what PR #1337 would be out of principal.. Now. I don't get it. What does this do @jstucke?

Seems I forgot to include a description, sorry. The raw JSON result preview currently works with syntax highlighting (the same as the file preview). That isn't bad, but large results can get kinda unwieldy fast and it would be nice if it was possible to fold in some parts of it. This isn't really possible with simple highlighted code. But there is an alternative: on the GraphQL Search page we use CodeMirror to display code. CodeMirror *does* have an option to enable folding in parts of the code. Thus, I simply replaced the raw result element with CodeMirror.

PR #1338, opened 2025-01-31
System health stage3

jstucke:
- [moved queues into own table](https://github.com/fkie-cad/FACT_core/pull/1338/commits/8c42c5a8e706470d29d879a2027af1440e85ecf1)
- [moved plugin cards from system health to their own page](https://github.com/fkie-cad/FACT_core/pull/1338/commits/0f4c118f8db903525041beca0acef88b0113f536)
- [fix for nice list border bug](https://github.com/fkie-cad/FACT_core/pull/1338/commits/4da0997c27702d4860cb56d85e9f2f8250cd3705)

### DIFF
--- a/src/web_interface/components/statistic_routes.py
+++ b/src/web_interface/components/statistic_routes.py
@@ -33,6 +33,13 @@ class StatisticRoutes(ComponentBase):
         plugin_dict.pop('unpacker')
         return render_template('system_health.html', analysis_plugin_info=plugin_dict)
 
+    @roles_accepted(*PRIVILEGES['status'])
+    @AppRoute('/plugins', GET)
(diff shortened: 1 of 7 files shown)


euwint (changes requested):
Separating the plugin cards into an own page definitely makes sense.
The new Analysis Plugin Status table does look good. 
However, I have two change requests:
1. The positional relationship between the numbers and the corresponding symbols (e.g. "currently processing") should be adjusted. 
2. The textual representation of the progress should be correctly centered within the progress bars. The respective progress bars are shown when clicking on "show details of the currently analysed firmware".

euwint (approved):
Now it looks good to me

PR #1340, opened 2025-02-04
fix: even more dark mode fixes

jstucke:
* fixed altert-primare element that is displayed when starting a binary search
* multiple small fixes for the text file diff


euwint (approved):
Looks good to me. The text file diff clearly looks better

Issue #1302, opened 2024-11-19
FACT comments 2.0


dorpvom:
### Feature description

The comment feature in FACT needs to be thoroughly upgraded to offer a real value in collaboration and note keeping. Some issues that should be addressed are:

- Aggregate comments on firmware level somehow
  - Making sure no comments are lost
- Highlight on top of analysis page (tag?) if comments exist
- Attach related plugin to comment
  - If navigated to comment, open plugin result (akin to pattern search, summary)
- Add comment should be modal
- Comments need not be collapsed
- Comment meta (user, timestamp) can be hidden and shown if wanted (hover, collapsible, ..)

Proposed plan of action:

1. Identify additional features, code dependecies
2. Adapt database scheme to fit analysis plugin
3. **First PR**: Add analysis plugin relation to comments
4. **Second PR**: Comment highlighting and aggregation
5. **Third PR**: Modal, new look comment rendering

PR #1341, opened 2025-02-04
add plugin to comments

jstucke:
* [refactor: split show_analysis.html template into included parts](https://github.com/fkie-cad/FACT_core/commit/19c81ec84cb3ce71e01eddc62fefee1b81915967)
* [fix: replace deprecated elements in comments template](https://github.com/fkie-cad/FACT_core/commit/3a1a80089c7ac57b1ad364d06d643b33b1bdaf3e)
* [feat: add plugin to comment](https://github.com/fkie-cad/FACT_core/commit/e236af47ebf174380d7d8bcf16092eff46dfb76e)
* [feat: always display comments (i.e. remove collapse)](https://github.com/fkie-cad/FACT_core/commit/6bfa33fc0e3c83d29877b99916872428fb4dfd28)

part of #1302


euwint (changes requested):
The changes definitely make sense.  
Splitting up the show_analysis.html also provides a clearer overview.  
However, I still have some change requests:

- Could the plugins in the comment dropdown also be sorted alphabetically?
-  The information about the author of the comment is currently lost. I think it would be interesting to have information about the author and the creation time of the comment. Therefore, the information should also appear in the table
-   Perhaps part of another pull request, but a comment function on the firmware comparison page also seems useful to me
-  Also perhaps part of another pull request, but in the comment info on the start page, it might be useful to include information about the firmware for which the comment was left.

jstucke:
> The information about the author of the comment is currently lost. I think it would be interesting to have information about the author and the creation time of the comment. Therefore, the information should also appear in the table

The information is actually still there: You can see it when you hover over the comment with your mouse

euwint (approved):
Looks good now :)

PR #1344, opened 2025-02-06
feat: revised IP&URI template

jstucke:
* fixed order of keys ('interesting' URIs -> URIs -> IPv4 -> IPv6)
* sorted results
* don't display 'interesting' URIs twice (also in URIs)


euwint (approved):
Looks good to me!

PR #1346, opened 2025-02-10
feat: converted init_systems plugin to v1

jstucke:


src/plugins/analysis/init_systems/code/init_system.py
```diff
@@ -1,155 +1,139 @@
...
+                    version=Version(1, 0, 0),
+                    Schema=self.Schema,
+                )
+            )
+        )
 
-    def additional_setup(self):
-        self.content = None
+    SYSTEMD_EXECSTART_REGEX = re.compile(r'ExecStart=(.*)', re.MULTILINE)
+    SYSTEMD_DESCRIPTION_REGEX = re.compile(r'Description=(.*)', re.MULTILINE)
```
maringuu, line 41: Unneeded `re.MULTILINE`

src/plugins/analysis/init_systems/code/init_system.py
```diff
@@ -1,155 +1,139 @@
...
-    VERSION = '0.4.2'
-    FILE = __file__
+class AnalysisPlugin(AnalysisPluginV0):
+    class Schema(BaseModel):
+        init_type: Optional[str] = Field(None, description='The type of init system that was identified for this file')
+        data: Optional[dict] = Field(
+            None,
+            description='Optional meta information and init data contained in this init script',
+        )
+        is_init: bool = False
```
maringuu, line 25: I would prefer sth. like this:
```python
class Schema(BaseModel):
    class SystemD:
        #: A list of all ExecStart lines
        exec_start: list[str]
        #: The Description of the unit
        description: str

    class Inittab:
        class Entry:
            id: str
            state: str
            action: str
            process: str
        entries: list[Entry]

    class Upstart:
        ...

    class SysVInit:
        ...

    systemd: Optional[SystemD]
    inittab: Optional[Inittab]
    sysvinit: Optional[SysVInit]
    upstart: Optional[Upstart]
```

I find it hard to argue for/against anything since I haven't had a use case for this plugin.
The only thing I can imagine is searching for units that execute a specific executable.
However, I think that the usefulness of a plugin increases with a more structured result.

maringuu: Also, isn't `init_type != None` equivalent to `is_init`?

maringuu (dismissed):
Contrary to my comment above, I'm not sure if we should simplify this plugin to do the following:
1. Detect init-systems by filename (As is done currently)
2. Display the whole file contents in the frontend

Currently, the schema does not seem too useful to me.
What do you think?

src/plugins/analysis/init_systems/code/init_system.py
jstucke: > I would prefer sth. like this:
> 
> ```python
> class Schema(BaseModel):
>     class SystemD:
>         #: A list of all ExecStart lines
>         exec_start: list[str]
>         #: The Description of the unit
>         description: str
> 
>     class Inittab:
>         class Entry:
>             id: str
>             state: str
>             action: str
>             process: str
>         entries: list[Entry]
> 
>     class Upstart:
>         ...
> 
>     class SysVInit:
>         ...
> 
>     systemd: Optional[SystemD]
>     inittab: Optional[Inittab]
>     sysvinit: Optional[SysVInit]
>     upstart: Optional[Upstart]
> ```
> 
> I find it hard to argue for/against anything since I haven't had a use case for this plugin. The only thing I can imagine is searching for units that execute a specific executable. However, I think that the usefulness of a plugin increases with a more structured result.

Hmm I don't think it makes a lot sense to have a result with the keys systemd, inittab, sysvinit and upstart and every value except one (at most) is `None`. Having the result as a list also doesn't make a lot of sense. How about inheritance? Can we have a "InitData" base class with more concrete inheritors with more fields? Does JSON schema support that?

> Also, isn't init_type != None equivalent to is_init?

Yes, absolutely. This could only bring better readability in code/queries but is totally redundant . It's probably better to throw it out

src/plugins/analysis/init_systems/code/init_system.py
jstucke: fixed

src/plugins/analysis/init_systems/code/init_system.py
jstucke: I changed it so that is uses multiple data classes with `Union[]`. This is equivalent to a "anyOf" relation in JSON schema. There seems to be no way to connect it directly to the type of init system, though.

euwint (approved):
Looks good and works fine :)

PR #1351, opened 2025-02-17
Fix ci test timeout

jstucke:


rhelmke (approved):
ok. Lets see if that fixes the timeout issues.

PR #1355, opened 2025-03-19
fix: strings plugin hex offset fix

jstucke:
fixes a bug where the hex offsets in the strings plugin view are displayed as decimals


euwint:
This requires discussion, as I do not see any changes.

euwint (approved):
Problem is solved :)

PR #1363, opened 2025-04-11
merged BinaryService and FSOrganizer

jstucke:
The BinaryService needs a DB interface to retrieve the file name. This...
- leads to unnecessary DB accesses, because most of the time the file name is actually not used
- this can lead to errors if there are multiple requests at once (because it is accessed from a threaded "intercom backend responder" and the DB session there is not thread safe)
- is kinda unnecessary anyway because in all cases where the file name is actually needed, we have access to a DB interface anyway and can simple use that to look it up

This PR removes DB access from BinaryService. There is now no real reason to have a separate "BinaryService" and "FSOrganizer" because they are closely related (the FSO stores files and the BS reads them), so we merge both and called the result "FileService".


src/intercom/back_end_binding.py
```diff
@@ -195,10 +200,11 @@ class InterComBackEndPeekBinaryTask(InterComListenerAndResponder):
 
     def __init__(self, *args):
         super().__init__(*args)
-        self.binary_service = BinaryService()
+        self.binary_service = FileService()
```
dorpvom: I'd argue towards being consequent in the renaming. At most other places you renamed the invocation of `FileService` as `file_service`. Why then keep the naming here as is? Note, that I'd be find with keeping the old names everywhere as well. Just aiming for consistency..

src/storage/file_service.py
```diff
@@ -0,0 +1,68 @@
...
+        local_file_path = self.generate_path_from_uid(uid)
+        delete_file(local_file_path)
+
+    def generate_path(self, file_object: FileObject) -> str:
+        return self.generate_path_from_uid(file_object.uid)
+
+    def generate_path_from_uid(self, uid: str) -> str:
+        return str(self.data_storage_path / uid[0:2] / uid)
+
+    def get_file(self, file_object: FileObject) -> bytes | None:
```
dorpvom: (Also applies to other functions in this class) The naming here might be misleading. When you call a function named `get_file`, I'd think of both a FileIO as well as a FileObject first, maybe even a name or path, but not a binary. This would be my 4th or so guess, so it's not totally misleading, but I'd say we can do better.
My proposal: Let's look at the uses and decide what fits best in most contexts.

src/test/common_helper.py
```diff
@@ -257,7 +261,7 @@ def objects_exist(compare_id):
 
     @staticmethod
     def get_hid_dict(uid_set, root_uid):  # noqa: ARG004
-        return {uid: 'hid' for uid in uid_set}
+        return dict.fromkeys(uid_set, 'hid')
```
dorpvom, line 264: Out of curiosity: Why change this? Is this now considered bad style? dict-comprehensions were , to my knowledge, considered fine.

src/web_interface/components/io_routes.py
```diff
@@ -91,7 +95,7 @@ def _get_file_download_mime(self, binary: bytes, uid: str) -> str:
 
     @roles_accepted(*PRIVILEGES['download'])
     @AppRoute('/ida-download/<compare_id>', GET)
-    def download_ida_file(self, compare_id):
+    def download_ida_file(self, compare_id: str) -> Response | str:
```
dorpvom, line 98: Note to self: We should remove this legacy code

dorpvom (changes requested):
All in all pretty sound. No code quality issues, just some naming convention stuff

src/test/common_helper.py
jstucke: not intentional; ruff auto-fix (https://docs.astral.sh/ruff/rules/unnecessary-dict-comprehension-for-iterable/#unnecessary-dict-comprehension-for-iterable-c420)
I restored the previous version and added it to the ignore list

src/storage/file_service.py
jstucke: renamed to `get_file_content`

src/intercom/back_end_binding.py
jstucke: not intentional, fixed

jstucke:
also fixed the "Call to requests without timeout" from codacy

src/storage/file_service.py
dorpvom: Hm. Good but why not go all-the-way and rename `get_partial_file` and `get_repacked_file` too?

dorpvom (approved):

PR #1364, opened 2025-04-24
feat: show pointer cursor for clickable tags

jstucke:
- show pointer cursor for clickable tags for analysis tags on the analysis page

Thanks to @stauchert for the suggestion!


euwint (approved):
Useful addition. Looks good

PR #1365, opened 2025-04-29
fix: fix postgres installation for older distros WIP

jstucke:


jstucke:
fixed by changing the apt cache config

PR #1366, opened 2025-05-05
feat: added switch for hiding symlinks in file tree

jstucke:


euwint (approved):
Looks good to me! Nice little refinement

PR #1367, opened 2025-05-14
feat: extended pw cracking wordlist of users&pw plugin

jstucke:
- added more pw cracking word lists
- updated unix passwd regexes to...
  - reduce FPs (especially IPv6 addresses)
  - add support for yescrypt


euwint (approved):
Looks good to me :)

PR #1375, opened 2025-06-02
chore: deprecate EoL distros Debian 11 and Ubuntu 20.04

jstucke:


src/helperFunctions/install.py
```diff
@@ -227,14 +227,15 @@ def check_distribution(allow_unsupported=False):
 
     :return: The codename of the distribution
     """
-    debian_code_names = ['buster', 'stretch', 'bullseye', 'bookworm', 'kali-rolling']
+    debian_code_names = ['buster', 'stretch', 'bullseye', 'bookworm', 'trixie', 'kali-rolling']
```
dorpvom, line 230: Shouldn't these shrink? Instead of being added to?

src/helperFunctions/install.py
```diff
@@ -250,7 +251,7 @@ def check_distribution(allow_unsupported=False):
         return 'fedora'
     msg = (
         f'Your Distribution ({distro.id()} {distro.version()}) is not supported. '
-        'FACT Installer requires Ubuntu 20.04/22.04/24.04, Debian 11/12 or compatible!'
+        'FACT Installer requires Ubuntu 22.04/24.04, Debian 12 or compatible!'
```
dorpvom, line 254: Thinking on this, the `allow_unsupported` parameter should be reachable from the `install.py` via command line argument. Maybe that's an additional PR though.

dorpvom (changes requested):

dorpvom (approved):

PR #1378, opened 2025-06-03
fix: kernel_config plugin: don't try to unpack compressed kernel images

jstucke:


rhelmke (approved):
The tests complete successfully, there is less code, and we finally get rid of some functionality that is covered by the unpackers either way.

win win!

PR #1384, opened 2025-06-18
feat: converted kernel config plugin to new base class

jstucke:


euwint (approved):
Looks good and works fine! :)

PR #1385, opened 2025-07-08
feat: added magika tool to file type plugin

jstucke:
TODO:

- [x] evaluate if this actually adds valuable information
- [x] evaluate possible negative effects on the plugin's performance
- [x] write test(s)


src/plugins/analysis/file_type/code/file_type.py
jstucke: The version should be raised to 1.1.0

src/plugins/analysis/file_type/code/file_type.py
```diff
@@ -1,29 +1,46 @@
...
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     import io
 
 
+class MagikaResult(BaseModel):
+    label: str
+    mime: str
+    group: str
```
jstucke: Could you add some description to the fields "label" and "group"? Based on the name alone, I'm not sure what they do exactly.

jstucke (commented):
Code looks good apart from the missing version bump.  `test_analysis_fail` fails in the CI, though.

src/plugins/analysis/file_type/code/file_type.py
da-lauer: in commit.

src/plugins/analysis/file_type/code/file_type.py
da-lauer: changed in commit

jstucke:
For whatever reason, the test `test_analysis_fail[normal content-expected_result0]` fails for me locally:

```python
E             Full diff:
E               {
E                   'dependant_analysis': {
E                       'full': 'ASCII text, with no line terminators',
E                       'magika': {
E             -             'confidence': 0.7856149077415466,
E             ?                                     -- ^^ --
E             +             'confidence': 0.7856144309043884,
E             ?                                   +++   ^^^
E                           'description': 'Generic text document',
E                           'group': 'text',
E                           'is_text': True,
E                           'label': 'txt',
E                           'mime': 'text/plain',
E                       },
E                       'mime': 'text/plain',
E                   },
E                   'first_byte': '6e',
E                   'name': 'b55f174c36fd4f56ecc04931099300014d4014377c73af1fa433e258a9b38604_14',
E                   'number': 42,
E                   'virtual_file_path': {},
E               }

src/test/integration/scheduler/test_analysis_fail.py:57: AssertionError
```

It seems there are some slight differences regarding the confidence float value. I suggest rounding the confidence value (more than 3 or 4 decimals are probably not that interesting anyway):

```python
confidence=round(magika_result.score, 3)
```

src/plugins/analysis/file_type/code/file_type.py
jstucke: I actually meant using `Field(description='...')` like in the Schema class :sweat_smile:

jstucke:
The test fails on the CI may be related to a side-effect of #1507 which should be fixed in #1540

PR #1392, opened 2025-08-01
fix: fix for skip/fail in dependencies in v1 plugin

jstucke:
during scheduling


euwint (approved):
Looks good and does fix the errors! :)

PR #1393, opened 2025-08-01
feat: converted known_vulnerabilities plugin to new base class

jstucke:
- [x] needs #1392 to work properly


euwint (approved):
Looks good & works flawlessly. However, we should bear in mind that we should update the HTML code soon to force line breaks. On narrow windows, the text currently extends into the next column within the table.

PR #1395, opened 2025-08-15
Cpu architecture v1

jstucke:
- renamed plugin directory to match internal plugin name
- fixed an empty device_tree plugin result resulting in an error when collecting dependency results for the cpu_architecture plugin
- converted the cpu_architecture to the new base class
- added a new test to improve coverage (analyze/summarize was not yet tested)
- fixed a bug in the ELF component that produced empty summary strings


euwint (approved):
Looks good! :)

PR #1396, opened 2025-08-15
feat: converted hashlookup plugin to new base class

jstucke:


euwint (approved):
Looks good, works flawlessly :)

PR #1397, opened 2025-08-15
feat: converted cwe_checker plugin to new base class

jstucke:


euwint (approved):
Looks very good and works fine! :)

PR #1411, opened 2025-09-19
fix: fix lief section.type and tag parsing errors

jstucke:


euwint (approved):
Looks good and does fix the errors :)

PR #1428, opened 2025-11-28
Fix: Resolve JSONB cast error in get_latest_comments method

willardy101-ank:
Pull Request description
## Problem
When clicking the Home page in FACT 4.2, users encounter a database error:
CAST(anon_1.jsonb_array_elements_1 AS JSONB)['time'] DESC

This occurs because SQLAlchemy 2.0 automatically adds CAST operations when using dictionary-style access on JSONB columns, resulting in invalid PostgreSQL syntax.

## Root Cause
The original code used:
```python
query = select(subquery).order_by(subquery.c.jsonb_array_elements['time'].desc())
SQLAlchemy translates this into a CAST operation on already-unpacked JSONB array elements, which PostgreSQL rejects.

Solution
Add explicit column label to the JSONB array elements
Use SQLAlchemy's text() function with PostgreSQL's native -> operator
This bypasses SQLAlchemy's automatic type conversion

Changes
Modified get_latest_comments() method in src/storage/db_interface_frontend.py
Added text import from sqlalchemy
Used .label('comment_data') for explicit column naming
Used text("comment_data -> 'time' DESC") for ordering

Testing
Verified that Home page loads without database errors
Confirmed that latest comments are displayed correctly
Tested with SQLAlchemy 2.0.30 and PostgreSQL

Additional Notes
This issue affects FACT 4.2 installations. Users experiencing this issue need to:
Apply this fix
Clear Python cache: find . -name "__pycache__" -exec rm -rf {} +
Restart uwsgi processes: pkill -9 -f uwsgi

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -5,7 +5,7 @@ import re
 from pathlib import Path
 from typing import Any, NamedTuple, Optional
 
-from sqlalchemy import Column, Integer, cast, func, or_, select
+from sqlalchemy import Column, func, or_, select, text
 
 from helperFunctions.data_conversion import get_value_of_first_key
 from helperFunctions.tag import TagColor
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jstucke** [comment]
Hi and thank you for your contribution. Weirdly enough, I cannot seem to reproduce the error. Which version of PostgreSQL are you using? Does the this test fail for you (without your fix)?

```sh
pytest src/test/integration/storage/test_db_interface_frontend.py::test_get_latest_comments
```

PR #1456, opened 2026-03-12
AnalysisPluginV0 typing improvements and fixes 

jstucke:
AnalysisPluginV0 typing improvements 
- improve plugin type hinting
- convert `TagColor` class to Enum to make sure custom plugins don't use arbitrary strings for `Tag.color`


dorpvom (changes requested):
Why is this called "convert TagColor" and then the "change of the TagColor" is only like 5 lines of the 50?

PR #1467, opened 2026-03-27
fix: fixed two bugs in the ipc ghidra script

jstucke:
* fixed AttributeError in get_vars_from_varnode
* fixed IndexError in filter_relevant_indices

### DIFF
--- a/src/plugins/analysis/ipc/test/test_ipc_ghidra.py
+++ b/src/plugins/analysis/ipc/test/test_ipc_ghidra.py
@@ -1,0 +1,165 @@
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def mock_ghidra_modules(monkeypatch):
+    ghidra_mock = MagicMock()
+    pcode_op_mock = MagicMock()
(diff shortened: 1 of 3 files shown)


src/plugins/analysis/ipc/test/test_ipc_ghidra.py
```diff
@@ -0,0 +1,165 @@
...
+
+    yield
+
+    for key in list(sys.modules):
+        if 'ipc_analysis' in key or 'format_strings' in key:
+            monkeypatch.delitem(sys.modules, key, raising=False)
+
+
+@pytest.fixture
+def format_strings():
```
dorpvom, line 60: I guess these "imprort-replacing-fixtures" are necessary to enforce, that the `mock_ghidra_modules` is applied before loading these imports?

dorpvom (approved):

src/plugins/analysis/ipc/test/test_ipc_ghidra.py
jstucke: Yes it is to lazy import these modules, so that the mock is available during import

PR #1473, opened 2026-04-27
Byte filter fix

jstucke:
* update to bitmath (dependency of common_helper_files) causes tests to fail
  * fixed filter function
* fixed ruff fails and typos

### DIFF
--- a/src/test/unit/web_interface/test_filter.py
+++ b/src/test/unit/web_interface/test_filter.py
@@ -67,7 +67,16 @@ def test_handle_uids():
     assert result == expected_result, 'output not correct'
 
 
-def check_nice_list_output(input_data):
+@pytest.mark.parametrize(
+    'input_data',
+    [
+        ['a', 'b'],
+        set('ab'),
+        b'ab',
+        'ab',
+    ],
+)
+def test_nice_list(input_data):
     expected = (
         '<ul class="list-group list-group-flush">\n'
         '\t<li class="list-group-item">a</li>\n'
@@ -76,14 +85,6 @@ def check_nice_list_output(input_data):
     )
     result = flt.list_group(input_data)
     assert result == expected, 'output not correct'
-
-
-def test_nice_list_set():
-    check_nice_list_output(set('ab'))
-
-
-def test_nice_list_list():
-    check_nice_list_output(['a', 'b'])
 
 
 def test_list_to_line_break_string():
@@ -189,7 +190,7 @@ def test_byte_number_filter(input_data, verbose, expected):
 
 @pytest.mark.parametrize(
     ('input_data', 'expected'),
-    [(b'abc', b'abc'), (123, '123'), (1234, '1,234'), (1234.1234, '1,234.12'), (None, 'not available')],
+    [(b'abc', "b'abc'"), (123, '123'), (1234, '1,234'), (1234.1234, '1,234.12'), (None, 'not available')],
 )
 def test_nice_number(input_data, expected):
     assert flt.nice_number_filter(input_data) == expected
--- a/src/web_interface/filter.py
+++ b/src/web_interface/filter.py
@@ -11,12 +11,13 @@ import stat
 import zlib
 from base64 import b64decode, standard_b64encode
 from collections import defaultdict
+from collections.abc import Container, Hashable, Iterable, Sized
 from datetime import timedelta
 from operator import itemgetter
 from re import Match
 from string import ascii_letters
 from time import localtime, strftime, struct_time, time
-from typing import TYPE_CHECKING, Any, Iterable, Union
+from typing import TYPE_CHECKING, Any
 
 import bitmath
 import packaging.version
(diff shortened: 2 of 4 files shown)


src/web_interface/filter.py
```diff
@@ -37,171 +38,177 @@
...
     input unix_time_stamp
     output string 'YYYY-MM-DD HH:MM:SS'
     """
-    if isinstance(unix_time_stamp, (float, int)):
-        tmp = localtime(unix_time_stamp)
-        return strftime('%Y-%m-%d %H:%M:%S', tmp)
-    return unix_time_stamp
+    if not isinstance(unix_time_stamp, (float, int)):
+        return str(unix_time_stamp)
+    tmp = localtime(unix_time_stamp)
```
dorpvom, line 189: Why?

src/web_interface/filter.py
```diff
@@ -341,25 +355,25 @@ def vulnerability_class(score):
...
     for key, value_list in privileges.items():
         for value in value_list:
             inverted_privileges.setdefault(value, []).append(key)
     return sorted(roles, key=lambda role: len(inverted_privileges[role]))
 
 
-def filter_format_string_list_with_offset(offset_tuples):
-    max_offset_len = len(str(max(list(zip(*offset_tuples))[0]))) if offset_tuples else 0
+def filter_format_string_list_with_offset(offset_tuples: list[tuple[int, str]]) -> str:
+    max_offset_len = len(str(max(list(zip(*offset_tuples, strict=True))[0]))) if offset_tuples else 0
```
dorpvom, line 376: Why?

src/web_interface/filter.py
```diff
@@ -37,171 +38,177 @@
...
-    if isinstance(i, (float, int)):
-        return nice_number_filter(i)
-    if isinstance(i, str):
-        return replace_underscore_filter(i)
-    if isinstance(i, bytes):
-        return bytes_to_str_filter(i)
-    return i
+def generic_nice_representation(i: Any) -> str:  # noqa: ANN401
+    match i:
+        case struct_time():
```
dorpvom, line 43: Since this is tested I guess this works. But why python would to this I can't fathom. Probably for exact this kind of construction.

src/web_interface/filter.py
```diff
@@ -37,171 +38,177 @@
...
-    return i
+    return str(i)
 
 
-def byte_number_filter(i, verbose=False):
+def byte_number_filter(i: int, verbose: bool = False) -> str:
     if not isinstance(i, (float, int)):
         return 'not available'
-    output = human_readable_file_size(i)
+    output = bitmath.Byte(bytes=i).best_prefix().format('{value:.2f} {unit}')
```
dorpvom, line 73: Why did you decide to move the logic here? Is the common_helper_files module viewed as obsolete?

src/test/unit/web_interface/test_filter.py
```diff
@@ -189,7 +190,7 @@ def test_byte_number_filter(input_data, verbose, expected):
 
 @pytest.mark.parametrize(
     ('input_data', 'expected'),
-    [(b'abc', b'abc'), (123, '123'), (1234, '1,234'), (1234.1234, '1,234.12'), (None, 'not available')],
+    [(b'abc', "b'abc'"), (123, '123'), (1234, '1,234'), (1234.1234, '1,234.12'), (None, 'not available')],
```
dorpvom, line 193: Is this a change from the library that we encourage? Or is this a change due to the other 100-ish changed lines in the filter module?

src/web_interface/filter.py
jstucke: https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

src/web_interface/filter.py
jstucke: There is actually a *very* lengthy discussion about this (see https://peps.python.org/pep-0634/). You can do indecipherable black magic stuff with the match statement but you can also use it for more simple matching like this. IMHO it can improve readabiltiy if used "correctly".

src/web_interface/filter.py
jstucke: 1. The problem was that bitmath received an update and it was not pinned, because it is a dependency of a dependency (common_helper_files). We probably need a lock file (e.g. uv.lock) to fix problems like this
2. I would like to reduce dependencies on these common libraries in the long term, because they are mostly unmaintained with really questionable code (in this case, the function was only one line long -- no real reason to even use a function here)

src/test/unit/web_interface/test_filter.py
jstucke: IMHO the filter functions should return strings (for direct insertion into the template). This one also usually does this -- except for this special case where you pass a bytes object to the function (which in itself does not make a lot of sense). The result in the template will probably not change either way (because it would also have been rendered as a string before) but I would like to make the return type of the function explicit

src/web_interface/filter.py
jstucke: I just found it better to handle the irregular case first (as in most other filter functions) and have the actual function logic outside the if without indentation. But I can change it back if you like

src/test/unit/web_interface/test_filter.py
jstucke: Is just returning the input value even a good strategy (i.e. silently failing)? Wouldn't it make more sense to raise a `ValueError` if you pass a byte string into the number filter?

dorpvom: I guess this solution is sound. Since no byte-strings should be stored in the database anyway the case should never trigger IRL.

dorpvom (approved):

PR #1475, opened 2026-04-29
fully replace self-compiled YARA with yara-python

jstucke:
Goal: fully replace CLI YARA calls with the `yara-python` module
Changes:
- introduces a new `helperFunctions.yara` module with YARA-related common functions
- `software_components`: replaces the `no_text_file` yara signature with a meta entry and uses the `file_type` plugin as a dependency
- changed various modules to use `yara-python`/`helperFunctions.yara`:
  - `yara_binary_search`
  - `extract_os_names`
  - `yara_signature_testing`
  - `compile_yara_signatures`


src/helperFunctions/yara.py
```diff
@@ -0,0 +1,209 @@
...
+
+    rule_source = _concat_yara_files_from_dir(source_path) if source_path.is_dir() else source_path.read_text()
+    rules = yara.compile(source=rule_source, externals=external_variables or {})
+    rules.save(str(target_path))
+
+
+def _concat_yara_files_from_dir(source_path: Path) -> str:
+    all_rules = []
+    for signature_file in sorted(source_path.iterdir()):
+        if signature_file.suffix in {'.yara', '.yar'}:
```
dorpvom: Is there something like `yara.test_rule_syntax(rule_file: Path) -> bool` that we could use to collect rule files based on syntax instead of by suffix? (thinking of 7z parameter to test container soundness). This would be more robust and we could even produce high-fidelity warnings.

src/helperFunctions/yara.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def _load_rules(rule_file: Path) -> yara.Rules:
+    if _rules_are_compiled(rule_file):
+        return yara.load(str(rule_file))
+    return yara.compile(str(rule_file))
+
+
+def _rules_are_compiled(rule_file: Path) -> bool:
+    with rule_file.open('rb') as fp:
```
dorpvom: Similar to above comment about robustness. We could probably just `yara.load` and catch an exception to fall back on non-compiled rules.

src/plugins/analysis/software_components/code/software_components.py
```diff
@@ -56,7 +59,9 @@ def __init__(self):
         self._yara = addons.Yara(plugin=self)
 
     def analyze(self, file_handle: FileIO, virtual_file_path: dict, analyses: dict[str, BaseModel]) -> Schema:
-        del virtual_file_path, analyses
+        del virtual_file_path
+        file_type_analysis: FileTypePlugin.Schema = analyses['file_type']
+        is_text_file = file_type_analysis.mime == 'text/plain'
```
dorpvom, line 66: Making this two lines looks to be for debugging purposes

src/install/requirements_pre_install.txt
```diff
@@ -11,6 +11,11 @@ werkzeug~=3.1.5
 toml==0.10.2
 # needed during installation of cve_lookup plugin
 ijson==3.4.0
+# FixMe: compile signatures during backend startup instead of installation
```
dorpvom, line 14: Yes indeed. Having these in pre-install seems out of place and this shouldn't even be needed in the frontend, no?

src/install/backend.py
```diff
@@ -77,31 +64,31 @@ def main(skip_docker):
...
-def _install_docker_images():
+def _install_docker_images() -> None:
     # pull extraction docker container
     logging.info('Pulling fact extraction container')
-
-    docker_process = subprocess.run(
-        'docker pull fkiecad/fact_extractor', shell=True, stdout=PIPE, stderr=STDOUT, text=True, check=False
+    # FixMe: install docker images using Docker Python API: client.images.pull("...")
+    docker_process = subprocess.run(  # noqa: S602
+        '/usr/bin/docker pull fkiecad/fact_extractor', shell=True, stdout=PIPE, stderr=STDOUT, text=True, check=False
```
dorpvom, line 69: I always dislike this / absolute paths (for obvious reasons). Why do we need it here?

src/helperFunctions/yara.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    matched_strings = []
+    for match in matches:
+        for string in match.strings:
+            for instance in string.instances:
+                matched_strings.append(instance.matched_data.decode('utf-8', errors='replace'))
+    return matched_strings
+
+
+def compile_plugin_yara_signatures(
```
dorpvom: I'm not fond of this naming. `compile_yara_rules_from_directory` or something akin would be more precise. This is used in two places (compile script in `src/` and backend installation) and in one of them (the latter) the name is misleading.

src/plugins/analysis/software_components/signatures/network.yara
dorpvom, line 295: No change needed here? Does this not need a match-text-file flag then?

src/plugins/analysis/software_components/internal/extract_os_names.py
```diff
@@ -1,39 +1,28 @@
...
-SIGNATURE_PATH = '../signatures/os.yara'
-TARGET_PATH = '../bin/__init__.py'
+SIGNATURE_PATH = Path(__file__).parent.parent / 'signatures/os.yara'
+TARGET_PATH = Path(__file__).parent.parent / 'bin/__init__.py'
 
 
-def get_software_names(yara_file_path):
-    scanned_software = []
+def get_software_names(yara_file_path: Path) -> list[str]:
+    rules = yara.compile(str(yara_file_path))
```
dorpvom: Might be overkill, but why not use the new helper functions and structs here?

src/plugins/analysis/software_components/internal/extract_os_names.py
```diff
@@ -1,39 +1,28 @@
...
 
-SIGNATURE_PATH = '../signatures/os.yara'
-TARGET_PATH = '../bin/__init__.py'
+SIGNATURE_PATH = Path(__file__).parent.parent / 'signatures/os.yara'
+TARGET_PATH = Path(__file__).parent.parent / 'bin/__init__.py'
 
 
-def get_software_names(yara_file_path):
-    scanned_software = []
+def get_software_names(yara_file_path: Path) -> list[str]:
```
dorpvom: Legacy issue, but this seems to be a "private" function and thus deserves a `_` prefix.

src/plugins/analysis/software_components/internal/extract_os_names.py
```diff
@@ -1,39 +1,28 @@
...
-            scanned_software.append(software_name)
-
-    return scanned_software
-
-
-def extract_names(yara_file_path=SIGNATURE_PATH, target_path=TARGET_PATH):
-    stashed_directory = os.getcwd()  # noqa: PTH109
-    os.chdir(os.path.dirname(__file__))  # noqa: PTH120
 
+def extract_names(yara_file_path: Path = SIGNATURE_PATH, target_path: Path = TARGET_PATH) -> None:
```
dorpvom: I'm feeling a bit silly being so picky, but this also is a legacy thing: The function should be named to reflect it's functionality. It doesn't extract so much as it writes to a file. This is too big a "side effect" to not be mentioned in the name. Maybe `store_software_names_in_file`. TBH I would refactor this wholly:
Call this in the install as ~

`store_list_of_names_in_file(names=get_software_names(path-to-sig-file), file_name=path-to-os-list-file)` 

That makes it far less side-effecty and moves the information burden to the plugin's `install.py`

src/helperFunctions/yara_binary_search.py
```diff
@@ -18,86 +16,46 @@ class YaraBinarySearchScanner:
     This class provides functionality to scan files in the database for yara patterns. The public method allows to
     either match a given set of patterns on all files in the database or focus only on files included in a single
     firmware.
-
-    :param config: The FACT configuration.
+    FixMe: class is only used in back_end_binding -> move there
```
dorpvom, line 19: Ok. But maybe as a standalone file in the intercom module? It's 70-ish lines after all.

src/test/unit/helperFunctions/test_yara.py
```diff
@@ -0,0 +1,147 @@
...
+    return output_path
+
+
+def test_rules_are_compiled_compiled_file(temp_compiled_rules):
+    assert _rules_are_compiled(TEST_DIR / 'test.yara') is False
+    assert _rules_are_compiled(temp_compiled_rules) is True
+
+
+def test_compile_rules(tmp_path):
+    output_path = tmp_path / 'compiled_dir.yc'
```
dorpvom: You could also use the fixture from above for the first two lines, which are identical and then you also have a test for the fixture

dorpvom (changes requested):

src/plugins/analysis/software_components/code/software_components.py
jstucke: It is to help the type checker and IDE. Would you like me to convert it to one line?

src/install/requirements_pre_install.txt
jstucke: It is needed during installation, that's why we need to install it during "pre-install", right?
Compiling during backend startup is future work for another PR

src/install/backend.py
jstucke: I don't like it either but it is a valid safety concern: https://docs.astral.sh/ruff/rules/start-process-with-partial-path/

src/helperFunctions/yara.py
jstucke: Well ... it is just a wrapper around `compile_rules` which already does just that, but adjusted for a plugin folder. Maybe changing the docstring and renaming parameters would help?

src/plugins/analysis/software_components/signatures/network.yara
jstucke: I came to the conclusion that this was indeed a bug

src/plugins/analysis/software_components/internal/extract_os_names.py
jstucke: Let's move this into the init of the plugin (now that it uses yara-python it should be easy)

src/plugins/analysis/software_components/internal/extract_os_names.py
jstucke: Good question... actually the case where you load a file, compile it and don't save it, but rather return the compiled rules is currently not implemented by any of the functions. And since it is just one line, IMHO there is not really a reason to create another function for this

src/plugins/analysis/software_components/internal/extract_os_names.py
jstucke: I renamed the function to `_get_os_names`, moved it to software_components.py where it is called in the `__init__` function and removed the rest of the module

src/test/unit/helperFunctions/test_yara.py
jstucke: changed

src/helperFunctions/yara.py
jstucke: changed

src/helperFunctions/yara.py
jstucke: I added a function `is_valid_yara_source_file` to do just that

src/test/unit/helperFunctions/test_yara_binary_search.py
```diff
@@ -23,93 +16,43 @@ def get_all_files_in_fw(uid):
...
         assert isinstance(result, str)
         assert 'There seems to be an error in the rule file' in result
 
-    @patch('helperFunctions.yara_binary_search.subprocess.run', side_effect=mock_check_output)
-    def test_get_binary_search_yara_error(self, _):  # noqa: PT019
-        result = self.yara_binary_scanner.get_binary_search_result((self.yara_rule, None))
-        assert isinstance(result, str)
-        assert 'Error when calling YARA' in result
-
-    def test_parse_raw_result(self):
```
dorpvom, line 70: While the `parse_raw_result` function was removed, the functionality lives on (in spirit at least) in the `_convert_matches_to_result` function. Did you not test this explicitly, since you plan to remove that function in the near future?

src/plugins/analysis/software_components/code/software_components.py
dorpvom: Nah

dorpvom (approved):

PR #1476, opened 2026-05-04
resolute support

jstucke:


rhelmke (approved):
I tested the installation and some firmware images with various plugin presets. Seems to work!

PR #1487, opened 2026-06-12
chore: switch to hadolint-docker

jstucke:
* hadolint-docker does not need to be manually installed (unlike the version without docker)
* this fixes the pre-commit CI action


rhelmke (approved):
I mean it's sad that we now pull in docker dependencies for pre-commit hooks as well, but i figure that any fact-development requires docker either ways.

I just checked. it also works on Apple Silicon.

PR #1490, opened 2026-06-18
Fix CVE lookup setup

jstucke:
Since yesterday the NVD CVE feeds include SSVC data in addition to the CVSS data which caused a `KeyError` in the installation of the `cve_lookup` plugin


rhelmke (approved):
Works for me. Thanks.

PR #1496, opened 2026-07-01
feat: optional inverted summary

dorpvom:
* Added checkbox to analysis view summary block to allow inverting summary
  * Summary is displayed `1-n` where `1` is item and `n` is matched files.
  * Inversion shows `1` as matched file and `n` as matched items.
* Added parameter to database function to invert summary dict after query of database
* Created hand-over chain for checkbox to make "inverted" parameter accessible where needed
* Added helper function to fix "boolean string problem" to make future code duplication unnecessary
* Added tests for new functionalities


jstucke (approved):
LGTM!

PR #1498, opened 2026-07-06
feat: Software components  summary OS suffix

jstucke:


dorpvom (approved):
LGTM. But I think this will have either a merge conflict with #1475 or git is smart enough to detect that your changes are identical

PR #1499, opened 2026-07-07
Binary search cache miss fix

jstucke:


src/test/unit/web_interface/test_app_binary_search.py
```diff
@@ -79,6 +79,13 @@ def test_app_binary_search_post_single_firmware(self, test_client):
         )
         assert 'test_uid' in response
 
+    def test_binary_search_missing_cache(self, test_client):
+        response = test_client.get(
+            f'/database/browse?query={QUERY_CACHE_UID}1234',
```
dorpvom: Why not `f'/database/browse?query=invalid_uid'` to be explicit?

src/web_interface/components/database_routes.py
```diff
@@ -65,11 +70,14 @@ def _add_date_to_query(query, date):
...
     @AppRoute('/database/browse', GET)
-    def browse_database(self, query: str = '{}'):
+    def browse_database(self, query: str = '{}') -> str | Response:
         page, per_page = extract_pagination_from_request(request)[0:2]
         offset, limit = per_page * (page - 1), per_page
         total = None
-        parameters = self._get_search_parameters(query)
+        try:
+            parameters = self._get_search_parameters(query)
+        except RuntimeError as error:
```
dorpvom: I'm surprised, pre-commit / ruff allowed this. `RuntimeError` usually is frowned upon. Can we not scope this narrower? `ValueError` at least?

src/web_interface/components/database_routes.py
```diff
@@ -409,6 +421,7 @@ def proxy_graphql(self):
             data=request.get_data(),
             cookies=request.cookies,
             allow_redirects=False,
+            timeout=60,
```
dorpvom, line 424: #SideEffect

dorpvom (changes requested):

src/web_interface/components/database_routes.py
jstucke: Actually, this was also a ruff fix: https://docs.astral.sh/ruff/rules/request-without-timeout/

jstucke: The side effect is that it won't hang indefinitely (which is probably a good thing?)

src/web_interface/components/database_routes.py
jstucke: changed

src/test/unit/web_interface/test_app_binary_search.py
jstucke: that would be nice, but it must be a valid UID because the format is checked internally with `is_uid`. I introduced a new variable `uid_not_in_db` instead

dorpvom (approved):

PR #1501, opened 2026-07-14
Memory Optimizations

jstucke:
various memory optimizations to drastically reduce the required RAM to run FACT

* removed FileObject.binary and created, constructor methods (because the file can not longer be initialized "from binary")
  * redesigned `FileObject` and `Firmware` as a dataclass which assures that certain fields are always set
  * added method to store files to the intercom (because the file content is no longer transferred as part of the FileObject)
* added move method to FileService so that unpacked files don't need to be copied
* added more efficient UID calculation based on file paths without reading whole files into memory
* merged create_re_analyze_task and create_analyze_task
* optimized the file header comparison plugin, so that it does not read in whole files


src/unpacker/unpack.py
```diff
@@ -91,22 +92,29 @@ def generate_objects_and_store_files(
...
+            current_file = FileObject.from_uid(uid, file_name=path.name)
             current_virtual_path = get_relative_object_path(path, extraction_dir)
             current_file.temporary_data['parent_fo_type'] = magic.from_file(parent.file_path, mime=True)
 
             if current_file.uid not in extracted_files:
                 # the same file can be contained multiple times in one archive -> only the VFP needs an update
-                self.unpacking_locks.set_unpacking_lock(current_file.uid)
-                self.file_service.store_file(current_file)
+                self.unpacking_locks.set_unpacking_lock(uid)
+                if not path.is_symlink():
```
dorpvom, line 107: Ok, now talk this through with me: Moving instead storing ✔️  Storing symlinks (to ensuring correct relative path?) ✔️ But why do we store the symlink anyway, while we only move non-symlinks if they already exist. This does not seem consistent with the legacy behavior. Before we would store symlinks only once, just like any other file. Right?

src/test/unit/scheduler/test_unpack.py
```diff
@@ -79,7 +78,7 @@ def test_throttle(self, unpacking_scheduler, monkeypatch):
 def test_cancel_unpacking(unpacking_scheduler, caplog):
     unpacking_scheduler.sync_lock = Lock()
     unpacking_scheduler.currently_extracted = {}
-    test_fo = FileObject(binary=b'foo')
+    test_fo = create_test_file_object()
```
dorpvom: You could use `uid='foo'` to shorten this by a line, I guess

src/test/acceptance/test_file_download.py
```diff
@@ -19,25 +19,26 @@ def _show_analysis_page(self, test_client, fw):
...
+        assert 'testfile1' in file_names, 'test files could not be found in tar download'
 
     def test_firmware_download(self, backend_db, file_service, test_client):
         test_fw = create_test_firmware()
+        contents = test_fw.file_path.read_bytes()
         test_fw.processed_analysis.pop('dummy')
         test_fw.uid = test_fw.uid
         backend_db.add_object(test_fw)
-        file_service.store_file(test_fw)
+        file_service.store_file(contents)
```
dorpvom: Again minor, but use `uid=` to save 3,4 lines of executed code.

dorpvom (changes requested):

src/unpacker/unpack.py
jstucke: The problem why symlinks can't use `move_file_to_storage`:
- we don't save the actual symlink, instead we save a file with the content `b'symbolic link -> XYZ'` (FACT does not store actual symlinks)
  -  this is consistent with legacy behavior (see also `get_binary_from_file` from `common_helper_files` which was used by `store_file` before this change)
  - we don't really want to call `store_file` for identical symlinks, but it has `overwrite=False`, so it does not really matter (but we could also add another `if current_file.uid not in extracted_files:` here)
- calling `move_file_to_storage` on the symlink would move the file behind the link

src/test/unit/scheduler/test_unpack.py
jstucke: changed

src/test/acceptance/test_file_download.py
jstucke: changed. 
Also removed the utterly useless `test_fw.uid = test_fw.uid` 2 lines above :thinking:

src/unpacker/unpack.py
dorpvom: So in effect we traverse a number of in-expensive lines of code to reach the `overwrite=False` clause. In the context of filesystem interaction this seems trivial. So no need to add the `if` for performance. Now, I'm thinking about readability. I guess 

```suggestion
                if not path.is_symlink():
                    self.file_service.move_file_to_storage(path, uid)
                else:
                    content = f'symbolic link -> {path.readlink()}'.encode()
                    self.file_service.store_file(content, uid)
```

could improve this by reducing risk of getting lost in reasoning. Aside from one line of duplicate code, this would work, no?

dorpvom (approved):

dorpvom (approved):

PR #1505, opened 2026-07-16
feat: collapsed sw. comp. matching strings

jstucke:
* hides the matching strings of the software components results in a collapsed element
* put collapsible card in a macro, so that it can be reused

### DIFF
--- a/src/plugins/analysis/software_components/view/software_components.html
+++ b/src/plugins/analysis/software_components/view/software_components.html
@@ -1,5 +1,7 @@
 {% extends "analysis_plugins/general_information.html" %}
 
+{% import 'macros.html' as macros %}
+
 {% block analysis_result_details %}
 
     {% for software_data in analysis_result.software_components %}
@@ -54,7 +56,8 @@
                     {% endif %}
                     <tr>
                         <td>Matched Sequence</td>
-                        <td class="p-0">
+                        <td>
+                            {% call macros.expandable_element((software_data.matching_strings | length | string) + " Matching Strings", software_data.rule + "-matches") %}
                             <table class="table table-bordered table-sm m-0">
                                 <tr>
                                     <th>identifier</th>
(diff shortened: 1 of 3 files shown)


src/web_interface/templates/macros.html
```diff
@@ -60,3 +60,20 @@ <h5 class="card-title mb-3"><i class="fas fa-{{ icon }}"></i> {{ panel_title }}<
         {{ caller() }}
     </li>
 {% endmacro %}
+
+{% macro expandable_table(title, label, style="primary") %}
+    <div class="card border-{{ style }}">
+        <div class="card-header bg-{{ style }} text-white p-0" data-toggle="collapse"
+             data-target="#{{ label }}-matches" aria-expanded="false"
```
dorpvom: The `data-target` extends currently to e.g. `#OpenSSL-matches-matches`. Seems like a `matches` to many ;)

src/plugins/analysis/software_components/view/software_components.html
```diff
@@ -54,7 +56,8 @@
                     {% endif %}
                     <tr>
                         <td>Matched Sequence</td>
-                        <td class="p-0">
+                        <td>
+                            {% call macros.expandable_table((software_data.matching_strings | length | string) + " Matching Strings", software_data.rule + "-matches") %}
```
dorpvom: See other comment.

src/web_interface/templates/macros.html
```diff
@@ -60,3 +60,20 @@ <h5 class="card-title mb-3"><i class="fas fa-{{ icon }}"></i> {{ panel_title }}<
         {{ caller() }}
     </li>
 {% endmacro %}
+
+{% macro expandable_table(title, label, style="primary") %}
```
dorpvom: From my understanding, the "caller" must not be a table, but could be any html-element. While in most cases it will probably be used for tables, e.g, a figure / image would be possible as well.

dorpvom (changes requested):

src/web_interface/templates/macros.html
jstucke: You are totally right -> renamed to `expandable_element`

src/web_interface/templates/macros.html
jstucke: Oops... fixed

dorpvom (approved):

PR #1507, opened 2026-07-17
Unpacking sched connection recycling

jstucke:
* uses `close()` instead of `invalidate()` for DB sessions (so that sessions can be correctly recycled)
* shares connection between unpacking threads
  * sessions are not thread safe, but the connection pool/engine is, so we can share it


dorpvom (approved):
Code-wise, nothing to critique.

PR #1508, opened 2026-07-21
feat: add size to FO list meta data

jstucke:
* show file size in fw/fo list (home page and db search/browse)

### DIFF
--- a/src/storage/db_interface_frontend.py
+++ b/src/storage/db_interface_frontend.py
@@ -39,6 +39,7 @@ class MetaEntry(NamedTuple):
     hid: str
     tags: dict
     submission_date: int
+    size: int
 
 
 class CachedQuery(NamedTuple):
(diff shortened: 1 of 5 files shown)


src/storage/db_interface_frontend.py
```diff
@@ -39,6 +39,7 @@ class MetaEntry(NamedTuple):
     hid: str
     tags: dict
     submission_date: int
+    size: int = 0
```
dorpvom: From a quick inspection I find three invocations of `MetaEntry` in the tests where I'm curious why the tests seem to still work. While this might be a fallback behavior from the interpreter I'd propose to add the size field to these invocations. Be it just for clarity.

dorpvom (changes requested):

src/storage/db_interface_frontend.py
jstucke: removed the default value and fixed all tests; also added an assert in a test

dorpvom (approved):

PR #1519, opened 2026-07-29
fix: Fixed truncation to produce valid device tree.

dorpvom:
- Replaced `()` with `"<>"` to ensure result of truncation is valid device tree
- Fixed tests accordingly and added test in cpu_architecture to check for fallback behavior (regression)


src/plugins/analysis/cpu_architecture/internal/dt.py
```diff
@@ -17,20 +17,20 @@ def _get_compatible_entry(dts: str) -> str | None:
...
+            dtc_process = subprocess.run(  # noqa: S603
+                shlex.split(f'dtc -I dts -O yaml {tmp.name}'),
+                input=dts,
+                stdout=PIPE,
+                stderr=DEVNULL,
+                text=True,
+                check=True,
+            )
+    except subprocess.CalledProcessError:
+        logging.warning('Error in calling device tree compiler.')
```
jstucke: This is better than before, but it also completely hides what is going wrong. Maybe we could at least retrieve the (high level) error message from the output and log that? Or maybe log stderr with debug? dtc seems to produce error messages like "FATAL ERROR: Unable to parse input tree" or "1.1-2 syntax error" which are more informative than the current error message.

jstucke (approved):

PR #1526, opened 2026-08-05
feat: show better button with pointer for removing files from comparison

jstucke:
* replaces "x" button for removing selected files from comparison with trash bin
* shows pointer cursor when hovering over the button


dorpvom (approved):
Code LGTM

PR #1535, opened 2026-08-30
feat: detect Clash-compatible proxy configuration

bitmeta69:
Adds a software-component YARA rule for Clash-compatible proxy configurations.

The rule requires a top-level `external-controller` endpoint together with `proxy-groups`, `proxies`, or `rule-providers`. This limits false positives from comments, nested examples, inactive values, and unrelated YAML documents.

Includes direct positive/negative regression coverage and an end-to-end software-components plugin test.

Validation:

- Pre-commit checks passed
- YARA bundle compiled successfully
- All 24 relevant tests passed

This identifies compatible configuration structure; it does not attribute the firmware to a particular distribution or indicate malicious use.

### DIFF
--- a/src/plugins/analysis/software_components/signatures/network.yara
+++ b/src/plugins/analysis/software_components/signatures/network.yara
@@ -64,6 +64,36 @@ rule curl
         $a
 }
 
+rule ClashConfiguration
+{
+	meta:
+		software_name = "Clash-compatible proxy"
+		open_source = true
+		website = "https://wiki.metacubex.one/"
+		description = "Clash-compatible proxy configuration identified by its external controller and proxy settings"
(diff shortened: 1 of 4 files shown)


jstucke (commented):
Hi and thank you for your contribution.
Normally, the signatures match the executables. Why this focus on the configuration file and especially on the field "external-controller"? As far as I understand, this field is also optional, so we would not match the software if it is not included. Is this a potential security flaw? If so, the `known_vulnerabilities` plugin might be a better place for this rule.

Also, what devices is this targeted at? I tested the signature on the [LFwC](https://github.com/fkie-cad/linux-firmware-corpus) and got not a single match.

PR #1536, opened 2026-09-02
feat: add test for unused jinja filters and delete unused ones

da-lauer:
- Create python test that fails if filter is unused
- delete filters out of jinja_filters.py
- also delete the corresponding functions in filter.py and their tests in test_filter.py that are unused elsewhere

### DIFF
--- a/src/test/unit/web_interface/test_junused_jinja_filter.py
+++ b/src/test/unit/web_interface/test_junused_jinja_filter.py
@@ -1,0 +1,52 @@
+# ruff: noqa: S112
+import ast
+import re
+from pathlib import Path
+
+
+def test_unused_finja_filter() -> None:
+    filters_file_path = Path('src/web_interface/components/jinja_filter.py')
+    templates_dir = Path('src/')
+
+    assert filters_file_path.exists(), f'{filters_file_path} not found'
+    assert templates_dir.is_dir(), f'Verzeichnis {templates_dir} not found'
+
+    # extract filters
+    tree = ast.parse(filters_file_path.read_text(encoding='utf-8'))
+    setup_filters = next(
+        (node for node in ast.walk(tree) if isinstance(node, ast.FunctionDef) and node.name == '_setup_filters'), None
+    )
+    assert setup_filters, '_setup_filters not found'
+
+    filter_names = set()
+    for node in ast.walk(setup_filters):
+        if (
+            isinstance(node, ast.Call)
+            and isinstance(node.func, ast.Attribute)
+            and node.func.attr == 'update'
+            and node.args
+            and isinstance(node.args[0], ast.Dict)
+        ):
+            for key in node.args[0].keys:
+                if isinstance(key, ast.Constant) and isinstance(key.value, str):
+                    filter_names.add(key.value)
+
+    assert filter_names, 'no filters found in _setup_filters'
+
+    # read relevant filetypes
+    contents = []
+    template_suffixes = {'.html', '.j2', '.tmpl'}
+    for file_path in templates_dir.rglob('*'):
+        if file_path.is_file() and file_path.suffix.lower() in template_suffixes:
+            try:
+                contents.append(file_path.read_text(encoding='utf-8', errors='ignore'))
+            except Exception:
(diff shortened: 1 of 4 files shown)


src/test/unit/web_interface/test_junused_jinja_filter.py
```diff
@@ -0,0 +1,52 @@
...
+        if (
+            isinstance(node, ast.Call)
+            and isinstance(node.func, ast.Attribute)
+            and node.func.attr == 'update'
+            and node.args
+            and isinstance(node.args[0], ast.Dict)
+        ):
+            for key in node.args[0].keys:
+                if isinstance(key, ast.Constant) and isinstance(key.value, str):
+                    filter_names.add(key.value)
```
jstucke, line 32: IMHO using `ast` for this is a bit overkill. `_setup_filters` already tries to store the complete dict of filters somewhere, so we could simply give it some object to do just that and read the filters from there, e.g.

```python
from web_interface.components.jinja_filter import FilterClass


class MockApp:
    class JinjaEnv:
        def __init__(self):
            self.filters = {}

        def add_extension(self, _):
            pass

    def __init__(self):
        self.jinja_env = self.JinjaEnv()


def _get_filters() -> set[str]:
    app = MockApp()
    FilterClass(app, '', None)
    return set(app.jinja_env.filters)
```

src/test/unit/web_interface/test_junused_jinja_filter.py
```diff
@@ -0,0 +1,52 @@
...
+import re
+from pathlib import Path
+
+
+def test_unused_finja_filter() -> None:
+    filters_file_path = Path('src/web_interface/components/jinja_filter.py')
+    templates_dir = Path('src/')
+
+    assert filters_file_path.exists(), f'{filters_file_path} not found'
+    assert templates_dir.is_dir(), f'Verzeichnis {templates_dir} not found'
```
jstucke, line 12:
```suggestion
    assert templates_dir.is_dir(), f'{templates_dir} not found'
```

src/test/unit/web_interface/test_junused_jinja_filter.py
```diff
@@ -0,0 +1,52 @@
+# ruff: noqa: S112
+import ast
+import re
+from pathlib import Path
+
+
+def test_unused_finja_filter() -> None:
+    filters_file_path = Path('src/web_interface/components/jinja_filter.py')
```
jstucke, line 8: breaks when not running the test from the root directory (the suggestion below would fix this anyway, though)

src/test/unit/web_interface/test_junused_jinja_filter.py
```diff
@@ -0,0 +1,52 @@
...
+            except Exception:
+                continue
+
+    unused_filters = []
+    for filter_name in filter_names:
+        pattern = re.compile(rf'\|\s*{re.escape(filter_name)}\b')
+        if not any(pattern.search(content) for content in contents):
+            unused_filters.append(filter_name)
+
+    assert not unused_filters, f'unused Jinja-Filter: {", ".join(unused_filters)}'
```
jstucke, line 52:
```suggestion
    assert not unused_filters, f'unused Jinja filters: {", ".join(unused_filters)}'
```

src/web_interface/components/jinja_filter.py
```diff
@@ -163,34 +162,25 @@ def _setup_filters(self) -> None:
                 'all_items_equal': lambda data: len({str(value) for value in data.values()}) == 1,
                 'as_ascii_table': flt.as_ascii_table,
                 'auth_enabled': self.check_auth,
-                'base64_encode': flt.encode_base64_filter,
-                'bytes_to_str': flt.bytes_to_str_filter,
```
jstucke, line 167: seems to actually be used in `ajax_routes.py` in `ajax_get_binary()`. I would suggest just using `bytes.decode()` there instead of keeping the filter, though.

jstucke (changes requested):
looks good apart from some smallish things. The test seems to work nicely.

